PR #6, opened 2023-08-20
1.16 nether, new blocks and new structures

BlusteryS:
This pull adds vanilla nether and some structures.


lt-name:
In nk projects, it is common to use 4 spaces instead of a tab for indentation
I think it's better to change to 4 spaces, keep the same as before

BlusteryS:
In many files (not only mine) there are tabs, replace them with spaces?

BlusteryS:
image

lt-name:
Try to ensure that the new content uses four spaces

lt-name:
The default world generator doesn't seem to work, only the void

BlusteryS:
> The default world generator doesn't seem to work, only the void

image
image

src/main/java/cn/nukkit/level/Level.java
```diff
@@ -39,29 +51,80 @@
...
+import cn.nukkit.math.SimpleAxisAlignedBB;
+import cn.nukkit.math.Vector2;
+import cn.nukkit.math.Vector3;
+import cn.nukkit.math.Vector3f;
 import cn.nukkit.metadata.BlockMetadataStore;
 import cn.nukkit.metadata.MetadataValue;
 import cn.nukkit.metadata.Metadatable;
 import cn.nukkit.nbt.NBTIO;
-import cn.nukkit.nbt.tag.*;
-import cn.nukkit.network.protocol.*;
```
lt-name, line 56: Like this, I think you can continue to use * instead of citing item by item

lt-name:
> > The default world generator doesn't seem to work, only the void
> 
> image image

I tried again, it's the plugin's influence, sorry

src/main/java/cn/nukkit/level/Level.java
```diff
@@ -2969,36 +2970,31 @@ public void generateChunkCallback(int x, int z, BaseFullChunk chunk) {
...
-                    loader.onChunkPopulated(chunk);
-                }
-            }
-        } else if (this.chunkGenerationQueue.containsKey(index) || this.chunkPopulationLock.containsKey(index)) {
-            this.chunkGenerationQueue.remove(index);
-            this.chunkPopulationLock.remove(index);
+            setChunk(x, z, chunk, false);
+            chunk = getChunk(x, z, false);
+            if (chunk != null && (oldChunk == null || !isPopulated) && chunk.isPopulated() && chunk.getProvider() != null) {
+                new ChunkPopulationTask(this, chunk, getDimension() == Level.DIMENSION_NETHER ? Generator.netherPopulator : Generator.overworldPopulators).start();
```
lt-name, line 2988: Something doesn't seem right here.

BlusteryS: resolved

lt-name (approved):

PR #20, opened 2023-08-24
Fixed placement of trapdoors and added new trapdoors

BlusteryS:


src/main/java/cn/nukkit/block/Block.java
```diff
@@ -351,6 +351,12 @@ public static void init() {
             list[BLOCK_KELP] = BlockKelp.class; //393
             list[DRIED_KELP_BLOCK] = BlockDriedKelpBlock.class; //394
 
+            list[ACACIA_TRAPDOOR] = BlockTrapdoorAcacia.class; //400
+            list[BIRCH_TRAPDOOR] = BlockTrapdoorBirch.class; //400
```
lt-name, line 355: wrong id

lt-name (changes requested):

lt-name (approved):

PR #25, opened 2023-08-26
Fix Furnace Crafting and Add Experience Drops

glorydark:
1. Now furnace can select the right output from the input. (A Small Tweak)
TC20@LBH)L)$BJDX3FTZXDE

2. Now player can gain experience from using furnace to make items. (From PNX)


glorydark:
I think pnx has done a good job in this change, which enables us to get the source player easily. 
```java
@Since("1.19.50-r3")
@Override
public boolean setItemByPlayer(Player player, int index, Item item, boolean send) {
	if (index == 2 && (item.getId() == 0 || item.getCount() == 0)) {
		var holder = getHolder();
		var xp = holder.calculateXpDrop();
		if (xp > 0) {
			holder.setStoredXP(0);
			holder.level.dropExpOrb(player, xp);
		}
	}
	return setItem(index, item, send);
}
```

lt-name (approved):

PR #27, opened 2023-08-27
Added handling of MISSED_SWING action

BlusteryS:
Since 1.20.10 we need to send this sound to the players


glorydark (approved):

glorydark:
Great! 

But for crawling support mentioned in AuthAction, more effort seems to be needed, especially in recalculating player's bounding box.

Like: [Minecraft Wiki - Crawling](https://minecraft.fandom.com/wiki/Crawling).

PR #30, opened 2023-08-27
Basic Sneak Enhancement

glorydark:
``` java
Add hunger change for jump while sneaking
Modify sneak speed // 75% normal speed
Modify Eyeheight & height // Correct values are from WaterdogPE
Change bounding box while sneaking
Add Basic Methods & Events for Crawling
// Crawling should be implemented by modifying several things like BlockTrapDoor.
// But here we are only talking about fixing sneaking
```
目前如上的功能均经过测试，可以通过潜行，缩小自身体积，进入1.5格的空间。

Minecraft 2023_8_28 3_46_16
Minecraft 2023_8_28 3_46_25

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3245,7 +3245,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                     }
                 }
 
-                if(protocol >= ProtocolInfo.v1_20_10){
+                if (protocol >= ProtocolInfo.v1_20_10) {
                     if (authPacket.getInputData().contains(AuthInputAction.START_CRAWLING)) {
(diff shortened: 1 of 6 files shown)


src/main/java/cn/nukkit/Player.java
```diff
@@ -3242,6 +3245,28 @@ public void onCompletion(Server server) {
                     }
                 }
 
+                if (protocol >= ProtocolInfo.v1_20_10) {
```
lt-name, line 3248: 应该改成v1_20_10_21

lt-name (changes requested):

lt-name:
Crawling works fine in 1.20.10 with experimental mode turned on
M0{_)OTV )B25`SY62U{IRL

glorydark (commented):
Nothing wrong in review and test in local server.

PR #33, opened 2023-08-27
Initial code for the server-side enchanting

BlusteryS:
This code will not work due to the following:
1. ServerAuthInventories is false (StartGamePacket - 298 line):
image
2. We do not handle ItemStackRequest, so if we enable ServerAuthInventories the enchantment table won't work (PlayerAuthInputPacket - 73 line):
image


BlusteryS:
If `ServerAuthInventories `is false, then enchantments are generated on the client side. So we have to enable `ServerAuthInventories `and do some processing for item stack request packets.

CoolLoong:
very cool!

PR #66, opened 2023-10-17
[Multiversion] Support for 1.20.40

lt-name:


lt-name:
The block palette needs to wait for pmmp data

PR #68, opened 2023-10-18
use vanilla time numbers, command parameters for some commands

BlusteryS:


lt-name:
When using the whitelist command, the prompt content is not translated correctly.

PR #70, opened 2023-10-20
scoreboard api (pnx)

BlusteryS:
- [x] ScoreBoard api
- [ ] /scoreboard command


src/main/java/cn/nukkit/scoreboard/data/DisplaySlot.java
```diff
@@ -1,16 +1,11 @@
-package cn.nukkit.network.protocol.types;
```
lt-name, line 1: I think we can leave this part in network.protocol package

BlusteryS: But in pnx it is in the package cn.nukkit.scoreboard

lt-name: It doesn’t need to be exactly the same as pnx

lt-name (approved):

Issue #103, opened 2023-11-27
Crafting no-pattern rockets


MaXFeeD:
# 🐞 Bug Feedback | BUG反馈

## 📝 Bug Description | BUG描述

When crafting rockets without a pattern, they are divided by 3 into a new slot, although the stackability remains. Rockets without a pattern release a creeper pattern, I don't know what this is related to.

## 🔍 Steps to Reproduce | 复现方法

1. Take 1x paper and 1x powder
2. Craft rocket, 3x will appear in new slot

## ✅ Expected Behavior | 预期行为

Stack crafted rockets, find out what can be with the pattern and whether it is related.

## 📸 Screenshots / Video | 截图或视频

![Screenshot_20231122-235312](https://github.com/MemoriesOfTime/Nukkit-MOT/assets/46795182/2f949d87-228c-466b-b239-de5eddace322)

## 💻 OS and versions | 操作系统和核心版本

* Nukkit-MOT master/8cac79a
* Version: 1.16.200.02 (protocol 422)
* Java: OracleJDK
* Java version: 17
* Platform: Android (Horizon)

glorydark:
After testing, it proves to be truuee in mot. Please wait for the following fix.

PR #119, opened 2023-11-30
Fix the issue that firework default possesses the creeper explosion type

glorydark:
Related to #103 

Just made some deletions umm...


glorydark:
Some useful methods may be added to help developers, such as setting explosion type and so on.

lt-name (approved):

PR #140, opened 2024-01-14
Fixed a vulnerability in TextPacket

rubinovpw:
If a proxy client or bot sends too many Jukebox popups, the server may crash.


src/main/java/cn/nukkit/network/protocol/TextPacket.java
```diff
@@ -64,6 +64,9 @@ public void decode() {
             case TYPE_JUKEBOX_POPUP:
                 this.message = this.getString();
                 int count = (int) this.getUnsignedVarInt();
+                if (count > 10) {
+                    break;
+                }
                 this.parameters = new String[Math.min(count, 128)];
```
lt-name, line 70: This is already limited to 128

PR #146, opened 2024-01-20
Set meta whilst getting vanilla namespaceID items

glorydark:
# Brief Desciption 修复内容
修复获取原版字符串id物品特殊值不设置的bug
Set value whilst getting vanilla namespaceID items

# Bug Description 介绍
It's quite weird that executing "/give BizarreDark minecraft:potion:24 1" returns a mundane potion. Instead, using numeric id to get the same item will be fine.

### DIFF
--- a/src/main/java/cn/nukkit/item/Item.java
+++ b/src/main/java/cn/nukkit/item/Item.java
@@ -1019,7 +1019,11 @@ public class Item implements Cloneable, BlockID, ItemID, ProtocolInfo {
             Supplier<Item> constructor = NAMESPACED_ID_ITEM.get(namespacedId);
             if (constructor != null) {
                 try {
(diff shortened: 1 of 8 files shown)


src/main/java/cn/nukkit/item/Item.java
```diff
@@ -1019,7 +1019,11 @@ public static Item fromString(String str) {
             Supplier<Item> constructor = NAMESPACED_ID_ITEM.get(namespacedId);
             if (constructor != null) {
                 try {
-                    return constructor.get();
+                    Item item = constructor.get();
+                    if (metaGroup != null) {
```
lt-name, line 1023: 可以直接使用meta变量，不需要重复解析一次

PR #173, opened 2024-02-10
Fix chest boat drop

labarjni:


lt-name:
The java8 branch regularly merges modifications to the main branch and does not require a separate PR update.

PR #204, opened 2024-03-04
fix(player): re-send BlockEntityDataPacket after fail block break

aabstractt:
This is a visual block bug for the player who was trying break a block like Double Chest, it dont was update correctly

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4554,6 +4554,11 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             handItem = this.level.useBreakOn(blockPos.asVector3(), face, handItem, this, true);
             if (handItem == null) {
                 this.level.sendBlocks(new Player[]{this}, new Vector3[]{blockPos.asVector3()}, UpdateBlockPacket.FLAG_ALL_PRIORITY);
+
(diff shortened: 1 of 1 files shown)


lt-name (approved):

src/main/java/cn/nukkit/Player.java
```diff
@@ -4554,6 +4554,11 @@ private void onBlockBreakComplete(BlockVector3 blockPos, BlockFace face) {
             handItem = this.level.useBreakOn(blockPos.asVector3(), face, handItem, this, true);
             if (handItem == null) {
                 this.level.sendBlocks(new Player[]{this}, new Vector3[]{blockPos.asVector3()}, UpdateBlockPacket.FLAG_ALL_PRIORITY);
+
+                BlockEntity blockEntity = this.level.getBlockEntity(blockPos.asVector3());
+                if (blockEntity instanceof BlockEntitySpawnable) {
```
lt-name, line 4559: I think it can be done directly like this
```
if (blockEntity instanceof BlockEntitySpawnable blockEntitySpawnable) {
    blockEntitySpawnable.spawnTo(this);
}
```

aabstractt: I don't see the need to create an extra variable if it will only be used once.

PR #211, opened 2024-03-14
Fix ResourcePacksInfoPacke packet (1.20.70)

labarjni:


BlusteryS:
`infoPacket.hasAddonPacks = false;`
it's already false

PR #214, opened 2024-03-15
Bad level tick

labarjni:
[main] ERROR - Could not tick level "survival": java.lang.NullPointerException: Cannot invoke "cn.nukkit.inventory.PlayerInventory.getItemInHandFast()" because the return value of "cn.nukkit.Player.getInventory()" is null
	at cn.nukkit.entity.passive.EntityRabbit.targetOption(EntityRabbit.java:61)
	at cn.nukkit.entity.EntityJumping.checkTarget(EntityJumping.java:42)
	at cn.nukkit.entity.EntityJumping.updateMove(EntityJumping.java:143)
	at cn.nukkit.entity.passive.EntityJumpingAnimal.onUpdate(EntityJumpingAnimal.java:28)
	at cn.nukkit.level.Level.doTick(Level.java:1012)
	at cn.nukkit.Server.checkTickUpdates(Server.java:1343)
	at cn.nukkit.Server.tick(Server.java:1419)
	at cn.nukkit.Server.tickProcessor(Server.java:1139)
	at cn.nukkit.Server.start(Server.java:1107)
	at cn.nukkit.Server.<init>(Server.java:821)
	at cn.nukkit.Nukkit.main(Nukkit.java:87)

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4997,7 +4997,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             this.perm = null;
         }
 
-        this.inventory = null;
+        this.inventory = new PlayerInventory(null);
         this.chunk = null;
 
         this.server.removePlayer(this);


lt-name:
In fact, there has been a similar problem before, and a simpler and more crude method was used to deal with it:
`Objects.requireNonNullElse(player.getInventory(), EMPTY_INVENTORY)`

lt-name:
I think this is a multithreading-related issue, where even though we check that it's not null, a NullPointerException might still occur in subsequent calls
When the inventory is empty, it typically means that the entity has been closed
So I think we can directly pass an empty inventory

labarjni:
If change this in Player#close(), and set empty inventory instead of null, will it work?
`this.inventory = new PlayerInventory(null);`

PR #218, opened 2024-03-18
CraftingGrid clear on cancel PlayerDropItemEvent

labarjni:
If the inventory is full, they drop when the event is canceled, they remain in the inventory in this field, when switching to another mode, they also remain there


lt-name:
This will cause the item to disappear directly

labarjni:
yes, it still needs to be finalized, but the fact is that if items are saved for each mode separately on someone's server, this is the dupe method

lt-name:
I think clearing items is best done by a plugin
Most developers cancel PlayerDropItemEvent because they don't want players to drop items, rather than making items disappear.
But thank you for your pr

PR #219, opened 2024-03-19
Mineral drops

labarjni:


lt-name:
I think we can add options to the configuration file like PM1E to make this function optional.

PR #220, opened 2024-03-21
Minor fixes in the commands

labarjni:


BlusteryS:
you shouldn't kick players in "whitelist off"

src/main/java/cn/nukkit/command/defaults/WhitelistCommand.java
```diff
@@ -90,11 +98,11 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
 
         switch (args[0].toLowerCase()) {
             case "add" -> {
-                sender.getServer().getOfflinePlayer(args[1]).setWhitelisted(true);
+                sender.getServer().getOfflinePlayer(args[1].toLowerCase()).setWhitelisted(true);
```
lt-name, line 101: The Server#lookupName() method will convert it to lowercase, so there is no need to repeat the operation here.

labarjni: Apparently, this bug has already been resolved earlier

PR #222, opened 2024-03-22
Implement PlayerDuplicateLoginEvent

labarjni:


lt-name (approved):

src/main/java/cn/nukkit/Player.java
```diff
@@ -2495,15 +2495,6 @@ protected void processLogin() {
             return;
         }
 
-        for (Player p : new ArrayList<>(this.server.playerList.values())) {
```
lt-name, line 2498: I think it's best not to modify the checking position of players with the same name

lt-name (commented):
I think it's best not to modify the checking position of players with the same name

lt-name:
I think this function can be done using PlayerPreLoginEvent
For now, this change may cause more problems

PR #232, opened 2024-03-28
fix: JsonSyntaxException

labarjni:
This could have caused an error and shutdown the RakNetInterface:
`com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was NUMBER at line 1 column 2 path $`


lt-name:
Exception caught in RakNetPlayerSession#serverTick()

PR #276, opened 2024-05-31
Rewrite speed calculation

lt-name:


glorydark:
Good work. Using multiplier can definitely make the operation easier and visualized.

PR #311, opened 2024-08-30
Fix the inability to override commands (Backport of https://github.com/CloudburstMC/Nukkit/pull/2196)

Yexeed:
This PR backports from upstream pull https://github.com/CloudburstMC/Nukkit/pull/2196

tl;dr fixes command overriding which is not possible without this changes
sample code:
```java
public static boolean unregisterCommand(Command command){
    command.setLabel(command.getName() + "__unregistered");
    return command.unregister(Server.getInstance().getCommandMap());
}
```

### DIFF
--- a/src/main/java/cn/nukkit/command/Command.java
+++ b/src/main/java/cn/nukkit/command/Command.java
@@ -2,12 +2,7 @@ package cn.nukkit.command;
 
 import cn.nukkit.Player;
 import cn.nukkit.Server;
-import cn.nukkit.command.data.CommandData;
-import cn.nukkit.command.data.CommandDataVersions;
-import cn.nukkit.command.data.CommandEnum;
-import cn.nukkit.command.data.CommandOverload;
-import cn.nukkit.command.data.CommandParamType;
-import cn.nukkit.command.data.CommandParameter;
+import cn.nukkit.command.data.*;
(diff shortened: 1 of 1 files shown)


lt-name:
I think different judgments need to be made based on usage.
It should not be null when registering a command

Yexeed:
> I think different judgments need to be made based on usage.
> It should not be null when registering a command

What? Just try to override a command without this PR, it wont work because this if-block was broken 8 years ago, mismatching "this.commandMap" with "commandMap". With this PR it's once again possible to override a command. However, why should we make some judjements about something THAT simple? Command SHOULD allow changes from foreign command map if:
- the command itself doesnt have a valid command map attached to it (`this.commandMap == null`)
or
- the command has command map equal to passed command map (`this.commandMap == commandMap`)
Its that simple

Yexeed:
Also, registering a command (not overriding it) works fine

lt-name:
```
public boolean register(CommandMap commandMap) {
    if (this.allowChangesFrom(commandMap)) {
        this.commandMap = commandMap;
        return true;
    }
    return false;
}
```
I mean register(null) will return true
After this pr merge, although unregister is fixed, register will be broken

Yexeed:
Null safety check should fix the problem?

PR #334, opened 2024-10-24
fix: small fixes & feat: end platform

KoshakMineDEV:
- Fixed Pink Petals breaking
- Fixed spawnpoint in end
- Added End Platform


src/main/java/cn/nukkit/Player.java
```diff
@@ -1830,7 +1830,7 @@ protected void checkBlockCollision() {
                 } else {
                     Level end = this.getServer().getLevelByName("the_end");
                     if (end != null) {
-                        this.teleport(end.getSafeSpawn(), TeleportCause.END_PORTAL);
+                        this.teleport(new Position(100.5, 49, 0.5, end), TeleportCause.END_PORTAL);
```
lt-name, line 1833: I think this will invalidate the spawn point setting of the end world
I think we should change the default configuration of end world generation

KoshakMineDEV:
I will change it tomorrow

lt-name:
It doesn't seem to be very good yet, but it can be merged as a temporary solution

PR #339, opened 2024-11-13
Fix Slab duplicate when placed on the top of a block

angga7togk:
## Before:
https://github.com/user-attachments/assets/6c96f8ce-ccfe-4912-ac76-8b4e98a839b7

## After:
https://github.com/user-attachments/assets/e5351d8f-7bfe-4aec-9199-45a860f72c89

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockSlab.java
+++ b/src/main/java/cn/nukkit/block/BlockSlab.java
@@ -14,6 +14,9 @@ import cn.nukkit.math.SimpleAxisAlignedBB;
  */
 public abstract class BlockSlab extends BlockTransparentMeta {
 
+    public static final int SLAB_BLOCK_TYPE_BIT = 0x07;
+    public static final int SLAB_TOP_BIT = 0x08;
+
     protected final int doubleSlab;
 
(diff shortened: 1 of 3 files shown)


src/main/java/cn/nukkit/block/BlockSlab.java
```diff
@@ -14,9 +14,6 @@
  */
 public abstract class BlockSlab extends BlockTransparentMeta {
 
-    public static final int SLAB_BLOCK_TYPE_BIT = 0x07;
```
lt-name, line 17: Why delete this?

angga7togk: i thinks it's no longer working, i have change in the setDamage and in Level.java i have fixed :)

is it still needed?

lt-name: Obviously, there are a lot of places below that use 0x07 and 0x08. Using this variable instead of writing 0x07 directly is to better show the meaning of the parameter.

angga7togk: oooh iam sorry :), wait repull :)

PR #340, opened 2024-11-13
Improve leveldb memory usage

lt-name:


KoshakMineDEV:
8048 chunks with old one
image

8048 chunks with improved leveldb
image

PR #341, opened 2024-11-15
feat: more items

KoshakMineDEV:


KoshakMineDEV:
If I haven't missed something we have all items from all mc versions now

PR #346, opened 2024-11-24
feat: mob improvements

KoshakMineDEV:
1. Needs more tests
2. EntityThreadTicker location and name is not final and can be changed

Current task list:
- [ ] Make entity ticking multithreaded, not just in one separate thread from main
- [X] Improve target finding with switching from `for in all server entities` to Server#getOnlinePlayers(). (Currently not pushed on Github)
- [X] Make some entities searching their target not every tick(reduce amount of target finding tasks) (Currently not pushed on Github)


KoshakMineDEV:
Closing this PR at least for now cuz at the moment I'm going to try to make similar to PNX mob behavior system to easily port PNX mobs and make AI much more extensible.

PR #364, opened 2024-12-30
Gen deepslate

MaXoNeRYT:
Deepslate Generator


MaXoNeRYT:
Deepslate Generator

KoshakMineDEV:
I will check it today, but I think that PR will be not accepted. I personally think that we shouldn't change Normal generator anymore at least for good support of ZoteCoreLoader project which uses MOT as it's base and works on 1.16.201.

It will be much better to add deepslate generation into a NewNormal generator and also make its generation lower than 0 coordinates

lt-name:
I think you can add a setting in the configuration file to set whether to deprecate this function

KoshakMineDEV:
> I think you can add a setting in the configuration file to set whether to deprecate this function

I wanna make it another way. In Normal generator we will just chances of ore generation and nothing more.
Also there will be NewNormal with generation of world with extended height, deepslate and deepslate ores. Maybe a few more biomes.

KoshakMineDEV:
Moved to #365

PR #386, opened 2025-02-05
feat: Allow forced safety enchantments to be turned off in configuration files

labarjni:


src/main/java/cn/nukkit/Server.java
```diff
@@ -3265,6 +3270,7 @@ private static class ServerProperties extends ConfigSection {
             put("op-in-game", true);
             put("xp-bottles-on-creative", true);
             put("spawn-eggs", true);
+            put("forced-safety-enchant", false);
```
lt-name, line 3273: I think it should default to true

labarjni: It doesn't really affect anything, but okay

labarjni: When get enchantments of item, all enchantments that exceed the max level are set to the max value, so I did this

lt-name (approved):

PR #404, opened 2025-02-17
change knockback to match vanilla

BlusteryS:


KoshakMineDEV:
is this done?

BlusteryS:
I think so, this is as close as possible to vanilla knockback, but it might be worth adding a toggle so that servers wanting to keep the old pvp aren't affected.

tachibana-kanade-angel:
## Codex Code Review

### Blocking

- EntityLiving.java:225 will not compile: the oldKnockback branch still references x/z after the parameters were renamed to deltaX/deltaZ.
- EntityDamageByEntityEvent.java:53 swaps damager and entity when delegating the 6-arg constructor, so existing callers/plugins using that overload will create events with attacker/victim reversed.

### Physics mismatch

- EntityLiving.java:253 returns when horizontal delta is near zero. Vanilla handles overlapping entities by adding a small random horizontal vector; this PR would apply no knockback when attacker and target occupy nearly the same X/Z.

### Likely vanilla mismatch

- Player.java:7204 hardcodes netherite resistance as 0.15 per armor piece and ignores the r parameter plus existing Attribute.KNOCKBACK_RESISTANCE. This makes resistance player-only and may produce incorrect values for mobs/custom attributes/plugins.

### Check formula

- EntityLiving.java:263 uses base * (modifier + 1), making Knockback I double and Knockback II triple the base impulse. Verify this against the intended vanilla Bedrock formula; it is a large behavior change from the previous +0.1 per level.

---
*Reviewed by Hermes Codex (PR #404)*

PR #407, opened 2025-02-20
refactor(CraftingManager): change packet fields to private and optimize packet caching

lt-name:


Mcayear:
The `rebuildPacket` function no longer builds packets for all versions, but only for the latest version (`ProtocolInfo.CURRENT_PROTOCOL`).

**Positive:**

This will optimize initial startup speed and memory usage.

**Negative:**

I hypothesize that a player *not* using `ProtocolInfo.CURRENT_PROTOCOL` might experience lag-induced disconnections upon their first login due to this change.
___
Overall, this is a viable performance optimization.

Mcayear:
Start Speed：
| Test Group | Raw Data (seconds)                                                                  | Data after Removing Outliers (seconds)                        | Average (seconds) |
| ------ | --------------------------------------------------------------------------------------- | -------------------------------------------------------- | ---------- |
| 1      | 8.474, 8.378, 8.697, 9.465, 8.867, 9.417, 8.773, 8.63                                | 8.474, 8.697, 8.867, 9.417, 8.773, 8.63               | 8.810      |
| 2      | 9.702, 9.683, 9.996, 9.254, 7.884                                                        | 9.702, 9.683, 9.254                                   | 9.546      |

I have used "Data after Removing Outliers" which is a common and clear way to describe the data after the highest and lowest values have been removed. I kept the numerical results the same, just translating the column headers. I have also centered the number in the cells.

RAM Usage：
| Test Group | Memory Data (MB)        | Average Memory Usage (MB) |
| :-------- | :---------------------- | :------------------------ |
| 1         | 308.58, 306.15         | 307.365                   |
| 2         | 361.45, 378.81         | 370.13                    |

___

Overall, the speed optimization is not significant, while the memory savings are approximately 60 MB.

Mcayear:
the `null` assignments, like `packet313 = null`, serve no purpose here.

https://github.com/MemoriesOfTime/Nukkit-MOT/pull/407/files#diff-4880b821743163bc060ccdcb2c07f185d0b117625ac2f427d7b09a43b0c57b5fR739

glorydark (approved):

PR #419, opened 2025-02-22
Fix Missing Recipes

glorydark:
I tested it locally, but it requires more testing.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1126,7 +1126,9 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             return;
         }
 
-        this.server.sendRecipeList(this);
+        if (this.protocol >= ProtocolInfo.v1_21_60) {
+            this.server.sendRecipeList(this);
+        }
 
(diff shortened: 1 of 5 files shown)


src/main/java/cn/nukkit/Player.java
```diff
@@ -1126,6 +1126,8 @@ protected void doFirstSpawn() {
             return;
         }
 
+        this.server.sendRecipeList(this);
```
lt-name, line 1129: For this modification, may need to test the compatibility of lower versions and TrimDataPacket

KoshakMineDEV: maybe just check protocol before sending it?

glorydark: > For this modification, may need to test the compatibility of lower versions and TrimDataPacket

Through local testing, I found that TrimDataPacket can stay here without moving. And I've added a protocal check for lower 1.21.60 and upper 1.21.60 while using Player#sendRecipeList().

However, when using smithing table to trim any armor, the transaction fails. Maybe it's another bug I suppose.

lt-name (approved):

PR #420, opened 2025-02-22
feat: more blocks

KoshakMineDEV:


glorydark (approved):
Minecraft 2025_2_23 7_47_00

It works well.

src/main/java/cn/nukkit/block/BlockWood.java
```diff
@@ -19,12 +19,12 @@ public class BlockWood extends BlockSolidMeta {
     public static final int JUNGLE = 3;
 
     public static final short[] faces = new short[]{
```
lt-name, line 21: This modification is wrong, we need the first two digits to distinguish the type of wood

KoshakMineDEV: Without this changes Bamboo Block rotation is just not working

lt-name: You should change runtime_block_states instead of it
Modifying this will break compatibility with all blocks with id 17

KoshakMineDEV: about them.. They are not working both on my changed version of BlockWood & on the old one. Try to rotate them on master(any except oak wood) and u will see what I'm speaking about

lt-name: master:
image

glorydark: BlockStem seems similar with Bamboo Block, whose data value is both determined by pillar_axis. 

Maybe BlockBambooBlock should extend BlockStem.

KoshakMineDEV: > master: image

this is logs, not wood. Logs are fine, while Stripped Wood and Wood not

lt-name: Wood and Stripped Wood do not use this face

glorydark (approved):

PR #426, opened 2025-02-24
Wrap Distance Check In Player Movement Check

glorydark:
Sometimes this part will cause the player to pull back. I suggest that this should be placed under movement check.


lt-name:
我认为应当想办法修复这个问题，而不是把这个功能关闭来解决问题

PR #432, opened 2025-03-03
fix: `execute` command slash_command type, soft enum correct.

Mcayear:
- Fixed the issue causing the `/execute` command to crash.

> Temporary fix: removed `ENUM_AS_CHAINED_COMMAND`

- Excluded duplicate values in CommandEnum

> Solved by overriding the hashCode and equals methods of the CommandEnum class

- Fixed command completion for `/execute run <command>`

> Since protocol 575, replaced `CommandParam.COMMAND` with `CommandParam.SLASH_COMMAND`
> The hook method still needs improvement.

- AvailableCommandsPacket will correctly send `softEnums`

> Source: [PowerNukkitX#AvailableCommandsPacket](https://github.com/PowerNukkitX/PowerNukkitX/blob/master/src/main/java/cn/nukkit/network/protocol/AvailableCommandsPacket.java#L83)

### DIFF
--- a/src/main/java/cn/nukkit/command/Command.java
+++ b/src/main/java/cn/nukkit/command/Command.java
@@ -3,6 +3,7 @@ package cn.nukkit.command;
 import cn.nukkit.Player;
 import cn.nukkit.Server;
 import cn.nukkit.command.data.*;
+import cn.nukkit.command.defaults.ExecuteCommand;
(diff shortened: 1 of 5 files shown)


Mcayear:
TODO：
`putString`, `putShort`, and other NBT action methods need to add the `this.` prefix.

lt-name (approved):

PR #434, opened 2025-03-07
fixed: register custom block

Reider745:


KoshakMineDEV (commented):
LGTM
But needs more tests. Works for 1.21.50, 1.21.60 will be checked today

lt-name (approved):

lt-name:
There seems to be a problem with the multi-state blocks. I'm not sure if it's my usage problem.

KoshakMineDEV:
> There seems to be a problem with the multi-state blocks. I'm not sure if it's my usage problem.

There's still many problems with blocks after this update. Now they are not breaking whole palette, but u can use not more than 2 properties for blockstate, blocks are not breakable, u can't set it with string id in setblock and many other problems

Also API is still really bad. I will merge this PR to another branch, completely refactor custom blocks and only after that merge it into master

PR #435, opened 2025-03-07
feat: enable update notification

KoshakMineDEV:


BlusteryS:
I think this should be added to the default server.properties

KoshakMineDEV:
Oh I just forgot about that XD

lt-name:
There is already 'update-notifications' for managing update information

PR #436, opened 2025-03-10
Update MobEquipmentProcessor.java

Jullius1392:
fix: kick "Too many failed packets"

### DIFF
--- a/src/main/java/cn/nukkit/network/process/processor/common/MobEquipmentProcessor.java
+++ b/src/main/java/cn/nukkit/network/process/processor/common/MobEquipmentProcessor.java
@@ -44,16 +44,16 @@ public class MobEquipmentProcessor extends DataPacketProcessor<MobEquipmentPacke
 
         Item item = inv.getItem(pk.hotbarSlot);
 
-        if (!item.equals(pk.item)) {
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/network/process/processor/common/MobEquipmentProcessor.java
```diff
@@ -44,16 +44,16 @@ public void handle(@NotNull PlayerHandle playerHandle, @NotNull MobEquipmentPack
...
         Item item = inv.getItem(pk.hotbarSlot);
 
-        if (!item.equals(pk.item)) {
-            player.getServer().getLogger().debug(player.getName() + " tried to equip " + pk.item + " but have " + item + " in target slot");
-            playerHandle.setFailedMobEquipmentPacket(playerHandle.getFailedMobEquipmentPacket() + 1);
-            if (playerHandle.getFailedMobEquipmentPacket() > MAX_FAILED) {
-                player.close("", "Too many failed packets");
+        if (!item.equals(pk.item, false, true)) {
+            Item fixItem = Item.get(item.getId(), item.getDamage(), item.getCount(), item.getCompoundTag());
+            if (fixItem.equals(pk.item, false, true)) {
```
lt-name, line 49: I think it should be true.

In addition, under normal circumstances, special values ​​should be completely consistent, and you may need to check other places

PR #437, opened 2025-03-17
fix: update stems on grow

labarjni:
- [ ] Fixed update states for stems
- [ ] Attached Melon Stem
- [ ] Attached Pumpkin Stem


lt-name:
Is this pr already done?

labarjni:
> Is this pr already done?

https://github.com/user-attachments/assets/73393523-c832-4ec6-9bc9-3c6560c13c9a


No, there's something else I need to do, I haven't fixed the problem yet when the pumpkin has already grown

labarjni:
image

It remains to finalize the facing for this block and add states to the rest of the protocols
I think I added some states incorrectly, I'll come back to that later

PR #444, opened 2025-03-28
feat: entity climate variants

KoshakMineDEV:
Contains some code from PowerNukkitX


KoshakMineDEV:
If I make EntityClimateVariation as interface for some reason Nukkit fails to register entity(some problems with reflection)
If @lt-name u will be able to fix this that would be perfect.

lt-name (approved):

PR #445, opened 2025-03-28
fix: attached stems pumpkin and melon

labarjni:
- [x] Fixed an updating the stem for the attached
- [x] Block states for attached stems
- [x] Return to the previous grow after the fruits are harvested


labarjni:
Review and merge after #447

labarjni:
This PR is done

src/main/java/cn/nukkit/block/BlockStemMelon.java
```diff
@@ -7,12 +7,21 @@
...
+
 /**
  * Created by Pub4Game on 15.01.2016.
  */
-public class BlockStemMelon extends BlockCrops {
+public class BlockStemMelon extends BlockCrops implements Faceable, BlockPropertiesHelper {
+
+    protected static final IntBlockProperty GROWTH = new IntBlockProperty("growth", false, 13, 0);
+
+    protected static final BlockProperties PROPERTIES = new BlockProperties(GROWTH);
```
lt-name, line 24: I think direction should also be added to PROPERTIES instead of being handled separately

src/main/java/cn/nukkit/block/BlockStemPumpkin.java
```diff
@@ -7,12 +7,21 @@
...
+
 /**
  * Created by Pub4Game on 15.01.2016.
  */
-public class BlockStemPumpkin extends BlockCrops {
+public class BlockStemPumpkin extends BlockCrops implements Faceable, BlockPropertiesHelper {
+
+    protected static final IntBlockProperty GROWTH = new IntBlockProperty("growth", false, 13, 0);
+
+    protected static final BlockProperties PROPERTIES = new BlockProperties(GROWTH);
```
lt-name, line 24: Same here

In addition, the value of growth should be 0-7, not 0-13

labarjni: PROPERTIES does not work correctly with this, it just sets 0 damage if it is higher than GROWTH

labarjni: or need to fix the PROPERTIES so that it allows you to set the damage higher

lt-name: Because you didn't add the direction property
Incorrect use of the GROWTH property to implement all states

src/main/java/cn/nukkit/block/BlockMelon.java
```diff
@@ -58,10 +58,10 @@ public boolean onBreak(Item item) {
         for (BlockFace face : BlockFace.Plane.HORIZONTAL) {
             Block block = this.getSide(face);
             if (block instanceof BlockStemMelon stemMelon) {
-               if (stemMelon.getBlockFace() == face.getOpposite()) {
-                   stemMelon.setPropertyValue(BlockCrops.GROWTH, 7);
-                   this.getLevel().setBlock(stemMelon, stemMelon, true, true);
-               }
+                if (stemMelon.getBlockFace() == face.getOpposite()) {
+                    stemMelon.setDamage(7);
```
lt-name, line 62: I do not recommend directly operating damage after using property.

src/main/java/cn/nukkit/block/BlockCrops.java
```diff
@@ -40,12 +53,14 @@ public boolean place(Item item, Block block, Block target, BlockFace face, doubl
...
         // Bone meal
         if (item.getId() == Item.DYE && item.getDamage() == 0x0f) {
-            if (this.getDamage() < 7) {
+            if (this.getPropertyValue(GROWTH) < 7) {
                 BlockCrops block = (BlockCrops) this.clone();
-                block.setDamage(block.getDamage() + Utils.random.nextInt(3) + 2);
-                if (block.getDamage() > 7) {
-                    block.setDamage(7);
+
+                block.setPropertyValue(GROWTH, block.getDamage() + Utils.random.nextInt(3) + 2);
```
lt-name, line 59: getPropertyValue should be used here instead of getdamage, and we need to check whether the limit is exceeded before setting

labarjni: ready

src/main/java/cn/nukkit/block/BlockPumpkin.java
```diff
@@ -79,6 +79,21 @@ public boolean place(Item item, Block block, Block target, BlockFace face, doubl
         return true;
     }
 
+    @Override
+    public boolean onBreak(Item item) {
+        for (BlockFace face : BlockFace.Plane.HORIZONTAL) {
+            Block block = this.getSide(face);
+            if (block instanceof BlockStemPumpkin stemPumpkin) {
+                if (stemPumpkin.getBlockFace() == face.getOpposite()) {
+                    stemPumpkin.setDamage(7);
```
lt-name, line 88: I think it is possible to use setPropertyValue(FACING_DIRECTION)

labarjni: Do I need to set it to DOWN?or how to use it

labarjni: Is everything right now?

lt-name (approved):

PR #449, opened 2025-04-06
feat: some settings are useful for servers

labarjni:
Adding useful settings for servers

`min_engine_version` in manifest.json
> default: loading all resource packs

Allows you to configure supported resource packs on specific protocols, so that there are no conflicts, you can configure a resource pack for each protocol

**Official parameter on wiki bedrock**

image

`space-name-mode`
> default: ignore (allow spaces completely)

The space check mode in nicknames will help to avoid errors with space checks if the `replacing` mode is set

`portal-ticks`
> default: 80

If the value of `vanilla-portals` is false, it allows you to set custom ticks before activating the portal

### DIFF
--- a/src/main/java/cn/nukkit/resourcepacks/ResourcePack.java
+++ b/src/main/java/cn/nukkit/resourcepacks/ResourcePack.java
@@ -10,7 +10,7 @@ public interface ResourcePack {
 
     UUID getPackId();
 
-    Integer getPackProtocol();
+    int getPackProtocol();
 
     String getPackVersion();
 
--- a/src/main/java/cn/nukkit/resourcepacks/ResourcePack.java
+++ b/src/main/java/cn/nukkit/resourcepacks/ResourcePack.java
@@ -10,6 +10,8 @@ public interface ResourcePack {
 
     UUID getPackId();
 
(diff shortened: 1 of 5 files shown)


lt-name:
Regarding the resource pack version, can we add a file in the resource pack to identify it?
I think this is more convenient than changing server.properties
Especially some plugins package the resource package into jar

labarjni:
Yes, I've thought about it, need to try to implement it

labarjni:
> Regarding the resource pack version, can we add a file in the resource pack to identify it? I think this is more convenient than changing server.properties Especially some plugins package the resource package into jar

Do you mean adding a property to the manifest file of the resource pack?

labarjni:
like that
image

src/main/java/cn/nukkit/resourcepacks/ResourcePack.java
```diff
@@ -10,6 +10,8 @@ public interface ResourcePack {
 
     UUID getPackId();
 
+    int getPackProtocol();
```
lt-name, line 13: A default return value should be added to ensure compatibility

lt-name (approved):

PR #457, opened 2025-06-03
Fix Multi Recipe & Wind Charge

glorydark:
1. Multi Recipe: Now, it only checks three multi recipes (Book Cloning, Item Repairing, Map Cloning) while others are unchecked and controlled by clients. 
2. Wind Charge: Add initial implementation of wind charge


glorydark:
Related to #137

lt-name (approved):

PR #477, opened 2025-07-17
fix kick on rapidly breaking

labarjni:


lt-name:
I think it is still necessary to record the number of errors, and if the misjudgment is large, it can be appropriately increased.

PR #489, opened 2025-08-03
fix: flights should be canceled when they are prohibited

CoolMagic233:


src/main/java/cn/nukkit/Player.java
```diff
@@ -559,6 +559,9 @@ public void resetInAirTicks() {
 
     public void setAllowFlight(boolean value) {
         this.adventureSettings.set(Type.ALLOW_FLIGHT, value);
+        if (!value && (this.gamemode == 0 || this.gamemode == 2)) {
```
lt-name, line 562: 应当使用player类中的变量而不是直接数字

lt-name (approved):

KoshakMineDEV:
WHY???

It was absolutely normal before. Why restrict allow flight?

PR #555, opened 2025-12-13
fix: optimization Ai mobs and redstone performance

labarjni:
**Optimization**
- [x] Optimization for `checkTarget` method
- [x] Replaced bad `Level#getEntites` to `Level#getNearbyEntities` [in follow range]
- [x] Collision optimize [cached - `Entity#getBlocksInBoundingBox` & `Entity#getBlocksAround` & `Entity#getCollisionBlocks`]
- [x] `BlockEntityHopper#onUpdate()` optimization, while for updating blocks instead of `.ifremove`
- [x] The infinite iteration of `BlockRedstoneWire#calculateCurrentChanges` is limited to 16

**Fixes & new features**
- [x] The isAiMob flag in `Level#getNearbyEntities` was implemented in such a way that entities could only recalculate nearby entities when the player or the entity itself moved
- [x] The EntityRanges class has been implemented for vanilla mob follow ranges
- [x] `BlockTypes` has been added to check block types by known IDs groups
- [x] Zombie & pillager breaking door in hardcore
- [x] Fix skeleton and stray aiming

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -285,9 +285,9 @@ public class Level implements ChunkManager, Metadatable {
     private final Object2ObjectMap<GameVersion, ConcurrentMap<Long, Int2ObjectMap<Player>>> chunkSendQueues = new Object2ObjectOpenHashMap<>();
     private final Object2ObjectMap<GameVersion, LongSet> chunkSendTasks = new Object2ObjectOpenHashMap<>();
 
-    private final Cache<Object, Object> entityNearbyCacheDirty = Caffeine.newBuilder()
-            .maximumSize(4096)
-            .expireAfterWrite(5, TimeUnit.SECONDS)
+    private final Cache<Long, Boolean> entityNearbyCacheDirty = Caffeine.newBuilder()
+            .maximumSize(512)
+            .expireAfterWrite(1, TimeUnit.SECONDS)
             .build();
 
     private final Cache<Long, Entity[]> nearbyEntitiesCache = Caffeine.newBuilder()
@@ -3061,7 +3061,7 @@ public class Level implements ChunkManager, Metadatable {
     public void setDirtyNearby(Entity entity) {
         if (entity == null || entity.getLevel() != this) return;
         long chunkKey = chunkHash(((int) entity.x) >> 4, ((int) entity.z) >> 4);
-        entityNearbyCacheDirty.put(chunkKey, Boolean.TRUE);
+        entityNearbyCacheDirty.put(chunkKey, true);
     }
 
     @NonComputationAtomic


src/main/java/cn/nukkit/blockentity/BlockEntityHopper.java
glorydark: There is an issue where two stacked pistons sometimes both fail to work after rejoining the server, but this works correctly in the latest master build.

labarjni: I'll check it out now and commit

src/main/java/cn/nukkit/Player.java
```diff
@@ -7941,4 +7940,5 @@ public float recalculateMovementSpeedToSend() {
         }
         return Math.max(newMovementSpeed, 0.00f);
     }
-}
\ No newline at end of file
+
```
glorydark: You can just sync your branch with master.

src/main/java/cn/nukkit/blockentity/BlockEntityHopper.java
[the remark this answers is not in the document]
glorydark: I'm not knowing why isChunkInUse returns false after I rejoin the server. It's a bit of weird.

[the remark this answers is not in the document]
labarjni: Yes, I removed this isChunkInUse, but there is still an issue where previously placed hopper (before this change) no longer accept more than 1 slot

It's not ready for merging yet, there are a couple of things that need to be fixed, but I'm almost done

src/main/java/cn/nukkit/Player.java
```diff
@@ -1975,7 +1975,7 @@ protected void checkBlockCollision() {
             if (this.inPortalTicks == this.server.portalTicks || (this.server.vanillaPortals && this.inPortalTicks == 25 && this.gamemode == CREATIVE)) {
                 EntityPortalEnterEvent ev = new EntityPortalEnterEvent(this, EntityPortalEnterEvent.PortalType.NETHER);
 
-                if (this.portalPos == null) {
+                if (this.portalPos == null && this.server.portalTicks >= 80) {
```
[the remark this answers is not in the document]
labarjni: There's no difference, but this.inPortalTicks would be more correct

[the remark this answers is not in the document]
glorydark: Using this.inPortalTicks would be better. Could we go with that?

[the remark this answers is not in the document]
labarjni: Ok, thanks

labarjni:
> There is still a flaw in the behavior: when a skeleton is spawned from an egg, it first wanders aimlessly for a while before targeting the player. However, after your recent fixes, the skeleton’s movement is noticeably smoother than before.
> 
> All other changes look good. **But I’ve also left a comment in the changes of Player.java for you to check.**

It looks like mobs target find is the same as before, but it still doesn't work instantly on negative Y, and idk why

labarjni:
It's done, the mobs are now finding their target, i think PR ready for merging

labarjni:
Its done, the PR is ready

glorydark (commented):
I found a weird pathfinding bugs during my local test: Mobs also can't walk over slabs/stairs at all now (and they might be stucked somehow). I remember they could work relatively well in master.

Minecraft for Windows 2026_1_2 22_54_04

Minecraft for Windows 2026_1_2 22_50_38

src/main/java/cn/nukkit/block/BlockDoor.java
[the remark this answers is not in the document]
labarjni: Damage can be dealt to it by more than 1 entity, but I don't think that's right. I originally did that
We need to see how it is in vanilla

labarjni:
> This is a good change. Mobs can now climb slabs. However, there are some issues left:

Are you sure that this worked in master? The pathfinding logic is now identical to master, and it will take much more time to fix movement and pathfinding. I don't have that time right now.

 It's possible in future PRs, but the main thing is that the entities are now more optimized.

glorydark:
> There are some collision issues, possibly in Entity#move or in updateMove

My comments are just nitpicks and soory for overwhelming you with a flood of comments. I agree that something further should be realized in the future.

If your PR is ready, I'll do a final review and then request a final review from lt-name today. Just free to tell me.

labarjni:
> > There are some collision issues, possibly in Entity#move or in updateMove
> 
> 
> 
> My comments are just nitpicks and soory for overwhelming you with a flood of comments. I agree that something further should be realized in the future.
> 
> 
> 
> If your PR is ready, I'll do a final review and then request a final review from lt-name today. Just free to tell me.

Yes, I think I have nothing more to add here

src/main/java/cn/nukkit/entity/mob/EntityStray.java
glorydark: Is "import org.apache.commons.math3.util.FastMath;" unused?

src/main/java/cn/nukkit/entity/mob/EntitySkeleton.java
glorydark: Is "import org.apache.commons.math3.util.FastMath;" unused?

glorydark (approved):
Except some unused imports, I think this pr is good to me.

src/main/java/cn/nukkit/Player.java
```diff
@@ -3443,6 +3443,13 @@ public void onCompletion(Server server) {
                     break;
                 }
 
+                // When the player moves, set dirty nearby entities to check the target
+                if (!this.isSpectator() && !this.isCreative()) {
```
lt-name: Different versions may send movement data to the server using different data packets. I think this should be migrated to handleMovement.

src/main/java/cn/nukkit/level/Level.java
```diff
@@ -284,6 +285,12 @@ public class Level implements ChunkManager, Metadatable {
     private final Object2ObjectMap<GameVersion, ConcurrentMap<Long, Int2ObjectMap<Player>>> chunkSendQueues = new Object2ObjectOpenHashMap<>();
     private final Object2ObjectMap<GameVersion, LongSet> chunkSendTasks = new Object2ObjectOpenHashMap<>();
 
+    private final LongSet entityNearbyCacheDirty = new LongOpenHashSet();
```
lt-name: I believe that entityNearbyCacheDirty still carries the risk of memory leaks.
It can also be implemented using Caffeine, which only requires storing boolean values ​​or a static value.

PR #582, opened 2026-01-11
fix: block duplication during multiple explosions

labarjni:


src/main/java/cn/nukkit/level/util/ExplosionSource.java
```diff
@@ -0,0 +1,9 @@
+package cn.nukkit.level.util;
+
+import cn.nukkit.block.Block;
+import cn.nukkit.entity.Entity;
+
+public sealed interface ExplosionSource permits ExplosionSource.EntitySource, ExplosionSource.BlockSource {
```
lt-name: If not used elsewhere, it can be written into Explosion as a subclass

PR #601, opened 2026-01-30
fix: entity collision leaks

labarjni:
## Problem
When an entity is created, an object is created for the entity's collision cache, but it is never closed or cleared when the entity is closed. This also applies to other caches within the class.

## Memory leak                                                                                                                                                                                                        
<img width="1280" height="923" alt="image" src="https://github.com/user-attachments/assets/5fcf0ef1-9c9c-489c-9055-60dc5f5963a4  " />

## Cause
Multiple instances of the EntityCollision class
`recentBlockChanges` is never cleared and tracks all block changes globally

## Fix
-  Added `EntityCollision#clearCaches()` method and call it in `Entity#close()` to invalidate internal caches
- Only the min cache for chunks is left through Caffeine `cacheChunks`
- All collision handling methods have been simplified and improved
- Added null and closed-state checks in `getCollisionBlocks()`
- Minimized block calls through `Level#getBlock` and `Chunk#getBlockId` methods
- Fixed liquid collision through blocks and X-axis errors
                                                         
## Testing
- Tested with a large number of entities, and performance is maintained
- RAM began to be unloaded as entities were closed
- The player and other entities no longer take damage when they are not in/above the block

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -227,7 +227,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     protected int lastTeleportTick = -1;
 
-    protected boolean connected = true;
+    protected volatile boolean connected = true;
     protected final InetSocketAddress rawSocketAddress;
     protected InetSocketAddress socketAddress;
     protected boolean removeFormat = true;
@@ -3175,7 +3175,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     }
 
     public void handleDataPacket(DataPacket packet) {
-        if (!connected) {
+        if (!this.connected) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


labarjni:
I will need more tests within 1-2 days

tachibana-kanade-angel:
### 🤖 AI Review: ✅ LGTM

**Changes:** Memory leak fix for entity collision caches (+149/-290 lines)

**Key improvements:**
- Added `EntityCollision#clearCaches()` called in `Entity#close()`
- Replaced LinkedHashMap with Caffeine cache (simpler, more efficient)
- Removed static `recentBlockChanges` set (was never cleared - main leak source)
- Added null/closed-state checks in `getCollisionBlocks()` and `getBlocksAround()`
- Fixed liquid bounding box calculations

**Code quality:**
- Good defensive programming with null checks
- Proper resource cleanup on entity close
- Simplified collision logic

---
<sub>🤖 AI Generated</sub>

PR #602, opened 2026-01-30
fix: allow activatable items (projectiles/armor) to be used when right-clicking blocks

lt-name:
## Summary                                                                                                                                                                                                                        
  - Fix projectile items (snowball, egg, ender pearl, etc.) not being throwable when right-clicking blocks                                                                                                                          
  - Fix armor items not being equippable when right-clicking blocks                                                                                                                                                                 
  - Fix item activation return logic in Level.useItemOn                                                                                                                                                                             
                                                                                                                                                                                                                                    
  ## Problem                                                                                                                                                                                                                        
  When right-clicking a block with projectile items or armor, the client predicts item consumption and sends `itemInHand` as air. The server's strict `Item.equals()` check fails because server still has the item, causing the    
  action to be rejected.                                                                                                                                                                                                            
                                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                                        
  - **ProjectileItem**: Add `onActivate()` to allow throwing when clicking blocks                                                                                                                                                   
  - **ItemArmor**: Add `canBeActivated()` and `onActivate()` to allow equipping when clicking blocks; refactor to extract `equipArmor()` and `playEquipSound()` methods                                                             
  - **Level.java**: Fix `useItemOn` return logic - return item after successful `onActivate()` regardless of remaining count                                                                                                        
  - **Player.java**: Add special case handling for client-predicted complete consumption (activatable item with count=1, client reports air)                                                                                        
  - **MobEquipmentProcessor**: Fix debug log format string

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -2746,8 +2746,8 @@ public class Level implements ChunkManager, Metadatable {
                 if (item.canBeActivated() && item.onActivate(this, player, block, target, face, fx, fy, fz)) {
                     if (item.getCount() <= 0) {
                         item = new ItemBlock(Block.get(BlockID.AIR), 0, 0);
+                        return item;
                     }
-                    return item;
                 }
             } else {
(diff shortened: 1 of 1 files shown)


lt-name:
@codex review

PR #607, opened 2026-02-09
feat: new collision api

labarjni:
Collisions methods were migrated to a separate class independently of `Level` and `Entity`

`CollisionHelper`

Backward compatibility was preserved for old methods

**CollisionHelper Usage Examples**

From `Entity` Class

**Old Way (Deprecated)**

```java
// Returns List<Block>
List<Block> blocks = entity.getCollisionBlocks();

// Returns Block[]
Block[] blocksArray = entity.getBlocksAround();
```

**New Way** 

```java
// Get CollisionHelper instance
CollisionHelper helper = entity.getCollisionHelper();

// Get colliding blocks as array
Block[] blocksArray = helper.getCollisionBlocks();

// Get blocks within entity's bounding box
Block[] aroundArray = helper.getBlocksAround();

// Get blocks in custom bounding box
Block[] inBoxArray = helper.getBlocksInBoundingBox(customBoundingBox);

// Check if inside specific block type
boolean inFire = helper.isInsideBlock(entity.boundingBox, Block.FIRE);
```

From `Level/Other` Class

**Old Way (Deprecated)**

```java
// Get colliding blocks
Block[] blocks = level.getCollisionBlocks(boundingBox, true);

// Check for collision blocks
boolean hasCollision = level.hasCollisionBlocks(entity, boundingBox);

// Get collision cubes (AABBs)
AxisAlignedBB[] cubes = level.getCollisionCubes(entity, boundingBox, true);

// Get colliding entities
Entity[] entities = level.getCollidingEntities(boundingBox, entity);
```

**New Way**
```java
// Get colliding blocks (with various options)
List<Block> blocks = CollisionHelper.getCollisionBlocks(
    level, 
    boundingBox, 
    entity,           // optional entity reference
    true,             // targetFirst - return first found collision
    false             // ignoreCollidesCheck
);

// Check for collision blocks
boolean hasCollision = CollisionHelper.hasCollisionBlocks(
    level,
    entity,           // optional entity reference  
    boundingBox,
    true              // checkCanPassThrough
);

// Get collision cubes (AABBs)
List<AxisAlignedBB> cubes = CollisionHelper.getCollisionCubes(
    level,
    entity,           // optional entity reference
    boundingBox,
    true,             // include entities
    false             // only solid entities
);

// Get colliding entities
List<Entity> entities = CollisionHelper.getCollidingEntities(
    level,
    boundingBox,
    entity            // entity to exclude (can be null)
);
```

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2720,7 +2720,7 @@ public abstract class Entity extends Location implements Metadatable {
      */
     @Deprecated
     public List<Block> getCollisionBlocks() {
-        if (this.collisionBlocks == null) {
+        if (this.collisionBlocks == null || this.positionChanged) {
             this.collisionBlocks = Arrays.asList(getCollisionHelper().getCollisionBlocks());
         }
         return this.collisionBlocks;


tachibana-kanade-angel (commented):
### 🤖 AI Review: ⚠️ Changes Requested

Thank you for this refactoring PR! Moving collision logic to a dedicated `CollisionHelper` class is a good architectural improvement that makes the codebase more modular and maintainable.

However, there are several important concerns that should be addressed before merging:

## 🔴 Critical Issues

### 1. **Performance Optimizations Removed** (Severe)
The original `EntityCollision` class contained several important performance optimizations that are now missing:

- **Adaptive check interval** - reduced collision checks when entity is stationary
- **Chunk cache** - avoided repeated chunk lookups
- **Collision cache** - cached collision results to avoid redundant calculations
- **Block change tracking** - detected when cache needs invalidation
- **Sampling for large areas** - avoided checking all blocks in large bounding boxes

These optimizations were crucial for performance, especially on servers with many entities. The new implementation performs full collision checks every time, which may cause significant performance degradation.

**Suggestion**: Consider preserving these optimizations in the new `CollisionHelper` class, or benchmark to ensure performance impact is acceptable.

### 2. **Thread Safety Concern**
```java
public CollisionHelper getCollisionHelper() {
    if (this.collisionHelper == null) {
        this.collisionHelper = new CollisionHelper(this);
    }
    return this.collisionHelper;
}
```

This method lacks thread-safety. If accessed from multiple threads (e.g., async chunk loading), it could create multiple instances or have visibility issues.

**Suggestion**: Make `collisionHelper` volatile or use synchronized initialization.

### 3. **Potential NPE After Entity.close()**
```java
public void close() {
    if (!this.closed) {
        this.closed = true;
        this.collisionHelper = null;  // Set to null
        // ...
    }
}
```

After `close()`, calling `getCollisionHelper()` will create a new CollisionHelper instance for a closed entity, which could cause issues.

**Suggestion**: Add a check in `getCollisionHelper()` to prevent creating new instances after close.

## 🟡 Important Issues

### 4. **Missing Performance Caching**
The old `EntityCollision.isInsideSpecialBlock()` cached the last checked bounding box and results (fire, lava, water, etc.) to avoid redundant checks. The new `CollisionHelper.isInsideBlock()` doesn't have this optimization.

For frequently-called methods like `isInsideOfFire()` and `isInsideOfLava()`, this could impact performance.

### 5. **Barrier Block Handling Missing**
```java
// Old code in Level.getCollisionCubes():
if (block.getId() == BlockID.BARRIER && entity.canPassThroughBarrier()) {
    continue;
}
```

This logic for handling barrier blocks is missing in the new `CollisionHelper.getCollisionCubes()`. This could change gameplay behavior.

### 6. **BlockNetherPortal Behavior Change**
```java
double growK = Server.getInstance().portalTicks > 20 ? 0.25 : 0.35;
```

This hardcoded logic for portal bounding box calculation should be documented or made configurable. Why 20 ticks? What's the gameplay impact?

## 🟢 Minor Issues

### 7. **Deprecation Messages**
The `@Deprecated` methods should include a clear migration guide in the Javadoc:
```java
/**
 * @deprecated Use {@link CollisionHelper#getCollisionBlocks(Level, AxisAlignedBB)}
 *             This method will be removed in version X.X.X
 */
```

### 8. **Missing Documentation**
- Record class `CollisionHelper` should document its purpose and usage examples
- Static utility methods vs instance methods - when to use which?

## 📊 Testing Concerns

- **No unit tests** provided for the new `CollisionHelper` class
- **No performance benchmarks** comparing old vs new implementation
- **No integration tests** verifying gameplay behavior is unchanged

**Suggestion**: Add tests covering:
- Edge cases (empty bounding boxes, world boundaries)
- Thread safety
- Performance regression tests
- Gameplay scenarios (portals, collisions, etc.)

## 💡 Positive Aspects

- ✅ Clean separation of concerns with `CollisionHelper`
- ✅ Good use of record class (Java 16+)
- ✅ Backward compatibility maintained with `@Deprecated` methods
- ✅ Modern Java APIs (List instead of arrays)
- ✅ Well-structured method overloads
- ✅ Good usage examples in PR description

## 🎯 Recommendation

**Cannot merge in current state** due to:
1. Removed performance optimizations may cause significant performance regression
2. Missing barrier block handling changes gameplay behavior
3. Thread safety concerns

**Suggested Path Forward**:
1. Benchmark old vs new implementation on a test server with many entities
2. Consider merging performance optimizations back into `CollisionHelper`
3. Add unit tests and integration tests
4. Fix barrier block handling
5. Address thread safety
6. Consider marking as `@Experimental` if you want to merge for wider testing

---

<sub>🤖 AI Generated</sub>

lt-name:
It may be necessary to test the EntityCollision compatibility issue in detail, or we can keep a compatibility class and mark it as deprecated before removing it entirely.

labarjni:
> It may be necessary to test the EntityCollision compatibility issue in detail, or we can keep a compatibility class and mark it as deprecated before removing it entirely.

There's no need for that, EntityCollision ([Entity#collisionCache](https://github.com/MemoriesOfTime/Nukkit-MOT/pull/607/changes#diff-c51a9eb6e8d61e1be2e17c71ccf0e4028bce6d9ce81eadb2c287af80bc44ee45L676)) was private and was only called from Entity, where I already marked the methods as deprecated

lt-name:
I feel that CollisionHelper doesn't really need to be a separate class. Many of its methods are already defined elsewhere, and it's just that CollisionHelper is being called additionally at the moment.

lt-name:
## Code Review: `feat/collisionapi` Branch

### Overview

This PR refactors collision detection logic from `EntityCollision` (a `ChunkLoader` implementation with caching) into `CollisionHelper` (a lightweight `record` in `cn.nukkit.utils` with static utility methods). `Level`'s collision methods are marked `@Deprecated` and now delegate to `CollisionHelper`.

**Scope:** 11 files, +874 / -651 lines.

---

### Critical Issues

#### 1. Collision offset calculation order is broken in `Entity.move()`

**Files:** `Entity.java:2633-2639`, `Entity.java:2665-2669`

The original code performs axis resolution **sequentially**, with an intermediate `boundingBox.offset()` between each axis:

```java
// Original (correct):
for (AxisAlignedBB bb : list) {
    dy = bb.calculateYOffset(this.boundingBox, dy);
}
this.boundingBox.offset(0, dy, 0);       // ★ BB moves in Y first

for (AxisAlignedBB bb : list) {
    dx = bb.calculateXOffset(this.boundingBox, dx);  // uses Y-adjusted BB
}
this.boundingBox.offset(dx, 0, 0);       // ★ BB moves in X second

for (AxisAlignedBB bb : list) {
    dz = bb.calculateZOffset(this.boundingBox, dz);  // uses XY-adjusted BB
}
this.boundingBox.offset(0, 0, dz);       // ★ BB moves in Z last
```

The new code calculates all three axes **independently against the same unmoved bounding box**:

```java
// New (incorrect):
double finalDy = collisions.stream().reduce(dy, (motion, bb) -> bb.calculateYOffset(this.boundingBox, motion), Double::min);
double finalDx = collisions.stream().reduce(dx, (motion, bb) -> bb.calculateXOffset(this.boundingBox, motion), Double::min);
double finalDz = collisions.stream().reduce(dz, (motion, bb) -> bb.calculateZOffset(this.boundingBox, motion), Double::min);

this.boundingBox.offset(finalDx, finalDy, finalDz);  // single offset at the end
```

**Why the order matters:** Each `calculateOffset` method performs **cross-axis overlap checks** before determining if a block can obstruct movement. For example, `calculateXOffset` checks whether the entity BB and block BB overlap on Y and Z axes — if they don't overlap, the block can't obstruct X movement. By not performing the intermediate `offset()` calls, the X and Z calculations use stale Y positions (and Z uses stale X positions), leading to:

- **Wall clipping:** Blocks that should obstruct movement on one axis may not register as overlapping on perpendicular axes because the BB hasn't been adjusted yet.
- **False obstruction:** The entity may be incorrectly blocked because overlap is evaluated against the pre-move position.
- **Broken step-height climbing:** The step-height logic (lines 2665-2669) is a 4-step sequential process (move up → resolve X → resolve Z → settle down). Flattening this into parallel calculation makes stair climbing fundamentally incorrect.

**This affects all entities** — players, mobs, projectiles, vehicles — since `Entity.move()` is the base movement method.

This is a well-known pattern from Minecraft's physics engine (both Java Edition and Bedrock). The axis order (Y→X→Z) is intentional and cannot be parallelized.

#### 2. Same issue in `BaseEntity.move()`

**File:** `BaseEntity.java:424-428`

```java
dx = collisions.stream().reduce(dx, (motion, bb) -> bb.calculateXOffset(this.boundingBox, motion), Double::min);
dz = collisions.stream().reduce(dz, (motion, bb) -> bb.calculateZOffset(this.boundingBox, motion), Double::min);
dy = collisions.stream().reduce(dy, (motion, bb) -> bb.calculateYOffset(this.boundingBox, motion), Double::min);

this.boundingBox.offset(dx, dy, dz);
```

The original axis order here is X→Z→Y (different from Entity's Y→X→Z), but the same principle applies — intermediate `offset()` calls between axes are required. All AI mob movement is affected.

#### 3. `Stream.reduce` combiner `Double::min` has incorrect semantics

**Files:** `Entity.java:2633-2635`, `BaseEntity.java:424-426`

The three-argument `reduce(identity, accumulator, combiner)` requires the combiner to be **associative** and **compatible with the accumulator**. Collision offset calculation is a **sequential fold** — each block constrains the motion based on the result from previous blocks. Using `Double::min` as the combiner would produce incorrect results if the stream were ever parallelized, because it would take the minimum of two independently computed partial results instead of sequentially accumulating constraints.

While this doesn't cause issues with sequential streams today, it represents incorrect semantics and a latent bug.

#### 4. `BlockNetherPortal.recalculateBoundingBox()` expands collision box by 0.3 blocks

**File:** `BlockNetherPortal.java:99-105`

```java
@Override
protected AxisAlignedBB recalculateBoundingBox() {
    // Fixed 0.3 expansion prevents collision misses at low ticks
    return new SimpleAxisAlignedBB(
            this.x - 0.3D, this.y, this.z - 0.3D,
            this.x + 1.3D, this.y + 1.0D, this.z + 1.3D
    );
}
```

The original code returns `this` (the block itself, a 1×1×1 BB). The new code expands the collision box by 0.3 in each horizontal direction (effectively 1.6×1×1.6). This means entities will trigger portal collision from 0.3 blocks away, which deviates from vanilla behavior. The comment suggests this compensates for collision detection misses, but expanding the block's own bounding box is a side-effect-heavy approach — it affects **all** code paths that use `collidesWithBB` on nether portals, not just the entity collision check.

If collision misses are a concern, the detection-side logic should be adjusted instead.

#### 5. `getCollisionCubes` NPE when `entity` is null

**File:** `CollisionHelper.java:457`

```java
if (block instanceof BlockBarrier && entity.canPassThroughBarrier()) {
```

The `entity` parameter is not annotated as `@Nullable`, but the legacy `Level.getCollisionCubes()` methods that delegate to this could theoretically pass null. The 2-parameter overload `getCollisionCubes(level, entity, bb)` does not guard against null `entity`. If null is passed, this line throws NPE.

---

### Medium Issues

#### 6. `Entity.close()` clears state before firing `EntityDespawnEvent`

**File:** `Entity.java:3072-3086`

```java
this.closed = true;

this.collisionHelper = null;
this.blocksAround = null;
this.collisionBlocks = null;
this.effects.clear();
this.passengers.clear();
// ... then later:
this.server.getPluginManager().callEvent(new EntityDespawnEvent(this));
```

The original `close()` method did not clear `effects`, `passengers`, or properties before firing the event. Plugins listening to `EntityDespawnEvent` may rely on reading these fields (e.g., logging what effects an entity had, checking passengers for cleanup). This change is outside the scope of the collision API refactor and breaks the plugin API contract.

**Suggestion:** Move the cleanup to after `EntityDespawnEvent` and `despawnFromAll()`, or remove it entirely (it was not in the original code).

#### 7. Unused import in `BlockNetherPortal.java`

**File:** `BlockNetherPortal.java:3`

`import cn.nukkit.Server;` is added but never referenced anywhere in the file.

#### 8. `BlockSeaPickle.onActivate()` refactored to use Stream with side effects

**File:** `BlockSeaPickle.java:171-185`

```java
blocksAround.stream()
    .filter(b -> b.getId() == CORAL_BLOCK)
    .map(Block::up)
    .filter(up -> up instanceof BlockWater)
    .filter(up -> up.getDamage() == 0 || up.getDamage() == 8)
    .filter(up -> random.nextInt(6) == 0)       // random in filter
    .forEach(up -> {
        level.setBlock(up, 1, new BlockWater(), true, false);  // world mutation in forEach
        level.setBlock(up, event.getNewState(), true, true);
    });
```

Two concerns:
- Using `random.nextInt(6) == 0` inside a `.filter()` is semantically misleading — it's not a property of the element but a side-effectful random decision.
- Calling `level.setBlock()` inside `.forEach()` performs world mutation within a stream pipeline, which is generally considered poor practice.

This refactoring is unrelated to the collision API change and unnecessarily changes working code.

---

### Minor Issues

#### 9. Mixed instance/static API design in `CollisionHelper`

The `record` has instance methods (`getCollisionBlocks()`, `getBlocksAround()`, `isInsideBlock()`) that operate on the bound entity, and static methods (`getCollisionBlocks(Level, ...)`, `getCollisionCubes(...)`) that take explicit parameters. The instance and static `getCollisionBlocks` share the same name but have different return types (`Block[]` vs `List<Block>`), which could be confusing for API consumers.

#### 10. Deprecated `getBlocksAround()` / `getCollisionBlocks()` caching may become stale

`Entity.java:2749-2764` now caches the result and only recomputes when the field is `null`:

```java
public List<Block> getBlocksAround() {
    if (this.blocksAround == null) {
        this.blocksAround = Arrays.asList(getCollisionHelper().getBlocksAround());
    }
    return this.blocksAround;
}
```

`entityBaseTick()` sets `this.collisionBlocks = null` but leaves `this.blocksAround` non-null (line 1893 has a comment: "Use only when entity moves for better performance"). The internal callers have been changed to use `getCollisionHelper().getBlocksAround()` directly (which always recomputes), but any plugin or external code calling the deprecated `getBlocksAround()` method may receive stale data.

#### 11. Missing newline at end of file

`CollisionHelper.java` does not end with a newline character.

---

### Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| Critical | 5 | Collision offset order (Entity + BaseEntity), reduce combiner semantics, nether portal BB expansion, NPE |
| Medium | 3 | `close()` early cleanup, unused import, stream side effects |
| Minor | 3 | API design consistency, caching staleness, missing EOF newline |

**The most critical issue is #1/#2** — the collision offset ordering change fundamentally breaks entity physics for all entity types. This is a core Minecraft mechanics pattern (Y→X→Z sequential resolution with intermediate BB updates) that cannot be parallelized. I'd recommend fixing this before merging, as it will cause widespread movement bugs (wall clipping, stuck entities, broken stair climbing).

The `List.of(Block.EMPTY_ARRAY)` issue from the earlier review has been fixed. ✅

lt-name:
**The following has been reviewed by AI and requires verification.**

## Code Review: `feat/collisionapi` Branch

### Overview

This PR refactors collision detection logic from `EntityCollision` (a `ChunkLoader` implementation with caching) into `CollisionHelper` (a lightweight `record` in `cn.nukkit.utils` with static utility methods). `Level`'s collision methods are marked `@Deprecated` and now delegate to `CollisionHelper`.

**Scope:** 11 files, +874 / -651 lines.

---

### Critical Issues

#### 1. Collision offset calculation order is broken in `Entity.move()`

**Files:** `Entity.java:2633-2639`, `Entity.java:2665-2669`

The original code performs axis resolution **sequentially**, with an intermediate `boundingBox.offset()` between each axis:

```java
// Original (correct):
for (AxisAlignedBB bb : list) {
    dy = bb.calculateYOffset(this.boundingBox, dy);
}
this.boundingBox.offset(0, dy, 0);       // ★ BB moves in Y first

for (AxisAlignedBB bb : list) {
    dx = bb.calculateXOffset(this.boundingBox, dx);  // uses Y-adjusted BB
}
this.boundingBox.offset(dx, 0, 0);       // ★ BB moves in X second

for (AxisAlignedBB bb : list) {
    dz = bb.calculateZOffset(this.boundingBox, dz);  // uses XY-adjusted BB
}
this.boundingBox.offset(0, 0, dz);       // ★ BB moves in Z last
```

The new code calculates all three axes **independently against the same unmoved bounding box**:

```java
// New (incorrect):
double finalDy = collisions.stream().reduce(dy, (motion, bb) -> bb.calculateYOffset(this.boundingBox, motion), Double::min);
double finalDx = collisions.stream().reduce(dx, (motion, bb) -> bb.calculateXOffset(this.boundingBox, motion), Double::min);
double finalDz = collisions.stream().reduce(dz, (motion, bb) -> bb.calculateZOffset(this.boundingBox, motion), Double::min);

this.boundingBox.offset(finalDx, finalDy, finalDz);  // single offset at the end
```

**Why the order matters:** Each `calculateOffset` method performs **cross-axis overlap checks** before determining if a block can obstruct movement. For example, `calculateXOffset` checks whether the entity BB and block BB overlap on Y and Z axes — if they don't overlap, the block can't obstruct X movement. By not performing the intermediate `offset()` calls, the X and Z calculations use stale Y positions (and Z uses stale X positions), leading to:

- **Wall clipping:** Blocks that should obstruct movement on one axis may not register as overlapping on perpendicular axes because the BB hasn't been adjusted yet.
- **False obstruction:** The entity may be incorrectly blocked because overlap is evaluated against the pre-move position.
- **Broken step-height climbing:** The step-height logic (lines 2665-2669) is a 4-step sequential process (move up → resolve X → resolve Z → settle down). Flattening this into parallel calculation makes stair climbing fundamentally incorrect.

**This affects all entities** — players, mobs, projectiles, vehicles — since `Entity.move()` is the base movement method.

This is a well-known pattern from Minecraft's physics engine (both Java Edition and Bedrock). The axis order (Y→X→Z) is intentional and cannot be parallelized.

#### 2. Same issue in `BaseEntity.move()`

**File:** `BaseEntity.java:424-428`

```java
dx = collisions.stream().reduce(dx, (motion, bb) -> bb.calculateXOffset(this.boundingBox, motion), Double::min);
dz = collisions.stream().reduce(dz, (motion, bb) -> bb.calculateZOffset(this.boundingBox, motion), Double::min);
dy = collisions.stream().reduce(dy, (motion, bb) -> bb.calculateYOffset(this.boundingBox, motion), Double::min);

this.boundingBox.offset(dx, dy, dz);
```

The original axis order here is X→Z→Y (different from Entity's Y→X→Z), but the same principle applies — intermediate `offset()` calls between axes are required. All AI mob movement is affected.

#### 3. `Stream.reduce` combiner `Double::min` has incorrect semantics

**Files:** `Entity.java:2633-2635`, `BaseEntity.java:424-426`

The three-argument `reduce(identity, accumulator, combiner)` requires the combiner to be **associative** and **compatible with the accumulator**. Collision offset calculation is a **sequential fold** — each block constrains the motion based on the result from previous blocks. Using `Double::min` as the combiner would produce incorrect results if the stream were ever parallelized, because it would take the minimum of two independently computed partial results instead of sequentially accumulating constraints.

While this doesn't cause issues with sequential streams today, it represents incorrect semantics and a latent bug.

#### 4. `BlockNetherPortal.recalculateBoundingBox()` expands collision box by 0.3 blocks

**File:** `BlockNetherPortal.java:99-105`

```java
@Override
protected AxisAlignedBB recalculateBoundingBox() {
    // Fixed 0.3 expansion prevents collision misses at low ticks
    return new SimpleAxisAlignedBB(
            this.x - 0.3D, this.y, this.z - 0.3D,
            this.x + 1.3D, this.y + 1.0D, this.z + 1.3D
    );
}
```

The original code returns `this` (the block itself, a 1×1×1 BB). The new code expands the collision box by 0.3 in each horizontal direction (effectively 1.6×1×1.6). This means entities will trigger portal collision from 0.3 blocks away, which deviates from vanilla behavior. The comment suggests this compensates for collision detection misses, but expanding the block's own bounding box is a side-effect-heavy approach — it affects **all** code paths that use `collidesWithBB` on nether portals, not just the entity collision check.

If collision misses are a concern, the detection-side logic should be adjusted instead.

#### 5. `getCollisionCubes` NPE when `entity` is null

**File:** `CollisionHelper.java:457`

```java
if (block instanceof BlockBarrier && entity.canPassThroughBarrier()) {
```

The `entity` parameter is not annotated as `@Nullable`, but the legacy `Level.getCollisionCubes()` methods that delegate to this could theoretically pass null. The 2-parameter overload `getCollisionCubes(level, entity, bb)` does not guard against null `entity`. If null is passed, this line throws NPE.

---

### Medium Issues

#### 6. `Entity.close()` clears state before firing `EntityDespawnEvent`

**File:** `Entity.java:3072-3086`

```java
this.closed = true;

this.collisionHelper = null;
this.blocksAround = null;
this.collisionBlocks = null;
this.effects.clear();
this.passengers.clear();
// ... then later:
this.server.getPluginManager().callEvent(new EntityDespawnEvent(this));
```

The original `close()` method did not clear `effects`, `passengers`, or properties before firing the event. Plugins listening to `EntityDespawnEvent` may rely on reading these fields (e.g., logging what effects an entity had, checking passengers for cleanup). This change is outside the scope of the collision API refactor and breaks the plugin API contract.

**Suggestion:** Move the cleanup to after `EntityDespawnEvent` and `despawnFromAll()`, or remove it entirely (it was not in the original code).

#### 7. Unused import in `BlockNetherPortal.java`

**File:** `BlockNetherPortal.java:3`

`import cn.nukkit.Server;` is added but never referenced anywhere in the file.

#### 8. `BlockSeaPickle.onActivate()` refactored to use Stream with side effects

**File:** `BlockSeaPickle.java:171-185`

```java
blocksAround.stream()
    .filter(b -> b.getId() == CORAL_BLOCK)
    .map(Block::up)
    .filter(up -> up instanceof BlockWater)
    .filter(up -> up.getDamage() == 0 || up.getDamage() == 8)
    .filter(up -> random.nextInt(6) == 0)       // random in filter
    .forEach(up -> {
        level.setBlock(up, 1, new BlockWater(), true, false);  // world mutation in forEach
        level.setBlock(up, event.getNewState(), true, true);
    });
```

Two concerns:
- Using `random.nextInt(6) == 0` inside a `.filter()` is semantically misleading — it's not a property of the element but a side-effectful random decision.
- Calling `level.setBlock()` inside `.forEach()` performs world mutation within a stream pipeline, which is generally considered poor practice.

This refactoring is unrelated to the collision API change and unnecessarily changes working code.

---

### Minor Issues

#### 9. Mixed instance/static API design in `CollisionHelper`

The `record` has instance methods (`getCollisionBlocks()`, `getBlocksAround()`, `isInsideBlock()`) that operate on the bound entity, and static methods (`getCollisionBlocks(Level, ...)`, `getCollisionCubes(...)`) that take explicit parameters. The instance and static `getCollisionBlocks` share the same name but have different return types (`Block[]` vs `List<Block>`), which could be confusing for API consumers.

#### 10. Deprecated `getBlocksAround()` / `getCollisionBlocks()` caching may become stale

`Entity.java:2749-2764` now caches the result and only recomputes when the field is `null`:

```java
public List<Block> getBlocksAround() {
    if (this.blocksAround == null) {
        this.blocksAround = Arrays.asList(getCollisionHelper().getBlocksAround());
    }
    return this.blocksAround;
}
```

`entityBaseTick()` sets `this.collisionBlocks = null` but leaves `this.blocksAround` non-null (line 1893 has a comment: "Use only when entity moves for better performance"). The internal callers have been changed to use `getCollisionHelper().getBlocksAround()` directly (which always recomputes), but any plugin or external code calling the deprecated `getBlocksAround()` method may receive stale data.

#### 11. Missing newline at end of file

`CollisionHelper.java` does not end with a newline character.

---

### Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| Critical | 5 | Collision offset order (Entity + BaseEntity), reduce combiner semantics, nether portal BB expansion, NPE |
| Medium | 3 | `close()` early cleanup, unused import, stream side effects |
| Minor | 3 | API design consistency, caching staleness, missing EOF newline |

**The most critical issue is #1/#2** — the collision offset ordering change fundamentally breaks entity physics for all entity types. This is a core Minecraft mechanics pattern (Y→X→Z sequential resolution with intermediate BB updates) that cannot be parallelized. I'd recommend fixing this before merging, as it will cause widespread movement bugs (wall clipping, stuck entities, broken stair climbing).

The `List.of(Block.EMPTY_ARRAY)` issue from the earlier review has been fixed. ✅

PR #634, opened 2026-02-28
refactor: optimize custom block definition serializer chain with v843 material conversion and permutation support

lt-name:


KoshakMineDEV:
Stealed from Lumi without even mentioning 😥

lt-name:
> Stealed from Lumi without even mentioning 😥

Written in Claude code, code review not yet conducted.

PR #647, opened 2026-03-07
新增玩家副手栏变动事件

ReiyansCN:
副手栏是玩家仅有的  使用EntityInventoryChangeEvent处理起来时十分麻烦  它应该被归属于玩家事件来处理


tachibana-kanade-angel (changes requested):
🤖 AI-Assisted Review\n\nThanks for the contribution — I re-reviewed this PR with Codex + manual verification.\n\nI found one blocking compatibility issue:\n\n- In  and , the new override calls only  for  holders and no longer follows the original  path that emits .\n- This changes existing API behavior for offhand updates and can break plugins that currently listen to  for generic inventory change interception/cancellation.\n\nPlease keep backward compatibility by preserving  emission (or explicitly fire both events for player offhand changes).\n\nOnce this compatibility regression is fixed, I’m happy to re-review.

PR #652, opened 2026-03-12
新增副手变动事件

ReiyansCN:
[+]新增使用EntityInventoryChangeEvent处理玩家副手物品变动
[/]修复了粗铜烧制产出为金锭的错误

### DIFF
--- a/src/main/resources/recipes649.json
+++ b/src/main/resources/recipes649.json
@@ -43219,7 +43219,7 @@
         "damage": -1
       },
       "output": {
-        "id": "minecraft:copper_ingot"
+        "id": 266
       },
       "block": "furnace"
     },


tachibana-kanade-angel (changes requested):
🤖 **AI-Assisted Review**: This code review was performed with AI assistance.

Found issues that need to be fixed ⚠️

**Must fix:**
1. Offhand rollback is incomplete when the event is cancelled - Location: `src/main/java/cn/nukkit/inventory/PlayerOffhandInventory.java`
   - Reason: In both `setItem()` and `clear()`, the cancellation path only sends a rollback update to `this.getViewers()`. However, offhand sync uses different packets for the holder and other viewers, so the holder's client may remain out of sync when `EntityInventoryChangeEvent` or `PlayerOffhandInventoryChangeEvent` is cancelled.
   - Suggestion: In the cancelled branches, also send the rollback update to the holder so the offhand slot is fully resynced for both the owner and viewers.

Please fix and resubmit for review.

PR #654, opened 2026-03-12
feat: add basic happy ghast functionality

glorydark:
todo:

- [x] Player can stand on happy ghast entity (Now, we have implemented collideable dataflag, but nothing happened)
- [x] Player should see harness if it is equipped (Now we have make EntityArmorEquipmentPacket to send body item, but nothing happened as well)

* For player to control happy ghast, I think we might as well leave it to the future task list. This pr is dedicated to implement basic functionality for developers.
* Currently, we will mostly not implement different seat positions because I'm lazy LOL.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,11 +6,11 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.java]
+[*.{java,xml}]
 indent_style = space
 indent_size = 4
 
-[*.{xml,yml,yaml,json}]
+[*.{yml,yaml,json}]
 indent_style = space
 indent_size = 2
 


tachibana-kanade-angel (changes requested):
🤖 **AI-Assisted Review**: This code review was performed with AI assistance.

Found issues that need to be fixed ⚠️

**Must fix:**
1. The new moveable property is written under the wrong identifier - Location: `src/main/java/cn/nukkit/entity/EntityMoveable.java:10`
   - Reason: `EntityMoveable` uses `minecraft:moveable`, but `entity_properties.nbt` registers Happy Ghast with `minecraft:can_move`. Because of this mismatch, `setMoveable(true)` does not update any synced entity property, so the client never receives the moveable state introduced by this PR.
   - Suggestion: Use the same identifier as the entity property definition (`minecraft:can_move`), or update the property definition and code consistently on both sides.

2. `setHelmet()` writes into the chest slot - Location: `src/main/java/cn/nukkit/inventory/EntityArmorInventory.java:66`
   - Reason: `setHelmet()` currently calls `setItem(SLOT_CHEST, item)` instead of `setItem(SLOT_HEAD, item)`. This breaks the shared armor inventory behavior for every entity using `EntityArmorInventory`, not just Happy Ghast.
   - Suggestion: Change `setHelmet()` to use `SLOT_HEAD`.

Please fix and resubmit for review.

glorydark:
> 🤖 **AI-Assisted Review**: This code review was performed with AI assistance.
> 
> Found issues that need to be fixed ⚠️
> 
> **Must fix:**
> 
> 1. `EntityArmorInventory` still exposes a 5-slot logical inventory to the holder, but the self-sync path only sends 4 items - Location: `src/main/java/cn/nukkit/inventory/EntityArmorInventory.java:118`
>    
>    * Reason: `getSize()` now returns `5` and the inventory stores the harness in `SLOT_BODY`, but when the holder itself is synced, `sendContents(Player player)` only sends `{helmet, chestplate, leggings, boots}` and omits the body slot entirely. The same special-case path exists in `sendSlot()` as well. This means the holder-side inventory view can never stay consistent with the actual inventory state once slot 4 is used.
>    * Suggestion: Either keep this inventory as a normal 5-slot container for the holder and send all 5 slots, or explicitly prevent holder-side window syncing/opening for `EntityArmorInventory` so the 5th slot is never exposed through an inconsistent packet shape.
> 
> The previously reported issues are fixed:
> 
> * `EntityMoveable` now matches `minecraft:can_move`
> * `setHelmet()` now writes to `SLOT_HEAD`
> 
> Please fix and resubmit for review.

Yes, it exactly has 5 slots. But in MobArmorEquipmentPacket, it has previously added a variable "head". We just use "mobArmorEquipmentPacket.body = this.getBody();" to send body item. The sending of the body item is not missing. Rather, it aims at better consistency because in MobArmorEquipmentPacket there are multi-version support previously set and applied.

PR #676, opened 2026-04-03
fix: add block registry checksum field to StartGamePacket

lt-name:


tachibana-kanade-angel:
## Codex Code Review

### Blocking

- Player sets blockRegistryChecksum unconditionally. For clients below v1_16_100, GlobalBlockPalette.getPaletteByProtocol() has no threshold and can throw during login, even though StartGamePacket only writes this field from v1_18_0. Guard it by protocol.
- Hash.xxh64() is not standard XXH64: several constants are wrong and tail processing handles only one 8-byte lane before byte fallback. This checksum will NOT match Bedrock/vanilla expected values. Use a vetted XXH64 implementation or add known test vectors.

### Major regression

- BlockPalette removes loadBlockStatesExtras(), dropping RuntimeBlockStatesExtras mappings for older palettes. This is unrelated to the checksum fix and can break runtime ID mapping.

### Correctness risk

- Checksum is computed once from the initial vanilla palette, but palettes can later be cleared/rebuilt with custom block states. If the checksum is meant to describe the actual registry sent/used by the server, it becomes stale.

### Needs validation

- calculateChecksum() should be compared against known vanilla checksum values for at least one protocol. Without fixtures, the NBT shape/order (name, states, id, val) is hard to trust.

### Unrelated change

- serverEngine changed from "" to the Minecraft version string. That should be separated or justified; it is not part of the checksum fix.

---
*Reviewed by Hermes Codex (PR #676)*

PR #744, opened 2026-06-06
fix: item frame vertical (up/down) facing placement and palette mapping

### LINKED ISSUE #741: Item frames cannot be placed on top/bottom and map insertion causes orientation change
# 🐞 Bug Feedback | BUG反馈
## 📝 Bug Description | BUG描述
物品展示框在放置时存在两个问题：

1. **放置限制问题**：物品展示框只能附着在方块的四周（类似墙壁），而无法放置在方块的顶部（类似天花板）或底部（类似地面）。

2. **地图放置方向问题**：当在已放置的物品展示框中放入地图时，会出现方向异常：
   - 在朝西方向的物品展示框中放入地图，物品展示框会从墙壁位置状态变为地面位置状态
   - 在朝东方向的物品展示框中放入地图，物品展示框会从墙壁位置状态变为天花板位置状态
   
   手动旋转一次后，物品展示框会恢复正常墙壁状态，且后续无论旋转多少次都不会再变回上下状态。

Item frames have two issues when placing:

1. **Placement Restriction**: Item frames can only be placed on the sides of blocks (like walls), but cannot be placed on the top (like ceiling) or bottom (like floor) of blocks.

2. **Map Placement Orientation Issue**: When placing a map into an already placed item frame, abnormal orientation changes occur:
   - Placing a map in a west-facing item frame changes it from wall-mounted state to floor-mounted state
   - Placing a map in an east-facing item frame changes it from wall-mounted state to ceiling-mounted state
   
   After manually rotating once, the item frame returns to the correct wall state, and no matter how many times it's rotated afterwards, it never returns to the up/down states.

## 🔍 Steps to Reproduce | 复现方法

1. 在方块的侧面放置一个空的物品展示框
2. 尝试在方块的顶部或底部表面放置物品展示框 - 观察到无法放置
3. 在朝西方向的物品展示框中放入地图 - 注意到物品展示框变为地面状态
4. 手动旋转物品展示框一次 - 观察到它恢复正常墙壁状态
5. 继续旋转物品展示框多次 - 确认它不会再变回上下状态
6. 在朝东方向的物品展示框中放入地图 - 注意到物品展示框变为天花板状态
7. 手动旋转物品展示框一次 - 观察到它恢复正常墙壁状态
8. 继续旋转物品展示框多次 - 确认它不会再变回上下状态

1. Place an empty item frame on the side of a block
2. Attempt to place an item frame on the top or bottom surface of a block - observe that it cannot be placed
3. Place a map into a west-facing item frame - notice the item frame changes to floor-mounted state
4. Manually rotate the item frame once - observe it returns to wall state
5. Continue rotating multiple times - confirm it never returns to up/down states
6. Place a map into an east-facing item frame - notice the item frame changes to ceiling-mounted state
7. Manually rotate the item frame once - observe it returns to wall state
8. Continue rotating multiple times - confirm it never returns to up/down states

## ✅ Expected Behavior | 预期行为

1. 物品展示框应该能够正常放置在方块的所有表面，包括顶部和底部
2. 当在物品展示框中放入地图时，物品展示框应该保持原有的方向和位置状态，不会发生意外的方向变化

1. Item frames should be placeable on all surfaces of blocks, including the top and bottom
2. When placing a map into an item frame, the item frame should maintain its original orientation and mounting state without unexpected changes

Image
Image

lt-name:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
This PR adds vertical (up/down) facing support for item frames, including palette mapping for synthetic runtime states. The direction is correct — wall-mounted item frames previously couldn't support up/down faces. However, there are blocking concerns with how the synthetic runtime IDs are registered in the palette.

### 🔍 Issues Found

**Blocking:**
- [ ] `[CRITICAL]` Synthetic vertical runtime IDs may be incorrect (`BlockPalette.java:225`)
  - The vertical runtime IDs are derived from `data=7` wall-map state using `runtimeId + 5` / `runtimeId + 10`. However, the synthetic NBT has `facing_direction=0/1` and `item_frame_map_bit=0`. These runtime IDs may not correspond to valid client states in the palette, potentially causing rendering issues or desync.
- [ ] `[CRITICAL]` `runtimeIdToLegacy.put()` can silently overwrite existing mappings (`BlockPalette.java:239`)
  - Unlike the safe `putIfAbsent()` used elsewhere, `put()` here will overwrite if the synthetic runtime ID collides with a real one. The current tests would still pass because they only verify self-consistency of the overwritten mapping.

**Minor:**
- [ ] `[MINOR]` `getFacingMeta()` edge case for future high-bit meta values (`BlockItemFrame.java:251`)
  - Handles `8/9` (UP/DOWN) and `12/13` (UP/DOWN + map) correctly, but any meta with other high bits (e.g. `item_frame_photo_bit`) falls through to `damage & 0x3`, which could produce incorrect wall face mapping.
- [ ] `[MINOR]` Test coverage is self-referential for palette correctness
  - The palette test verifies round-trip consistency but doesn't validate the synthetic state's `facing_direction` value against the intended face. A `place()` test for UP/DOWN would also improve confidence.
- [ ] `[MINOR]` Test suite could not be executed in sandbox (network/cache restrictions)

### 💡 Recommendations
- Consider using `runtimeIdToLegacy.putIfAbsent()` and logging a warning if collision occurs, rather than silently overwriting.
- Add a test that verifies the `facing_direction` field in the synthetic NBT state matches the expected value (0 for DOWN, 1 for UP).
- Verify the runtime ID offsets (+5, +10) against the actual client palette to ensure they're valid states.

### 📝 Notes
The overall approach of synthesizing vertical item frame states from the wall-map state is sound. The meta encoding changes (8 for DOWN, 9 for UP) and the new `getFacingMeta()` method are well-structured. Once the palette runtime ID registration is confirmed correct, this should be ready to merge.

---
*🤖 AI Review*

PR #745, opened 2026-06-07
Update RakNet library and enhance proxy protocol handling

lt-name:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
No blocking issues found. The changes look reasonable overall. This PR upgrades the RakNet library from 1.0.0.CR4 to 1.1.0.CR1 and refactors the Proxy Protocol handling to support IPv6 CIDR whitelists, delegate decoding to the upstream `cloudburstmc.netty.util.ProxyProtocolDecoder`, and properly clean up proxy address mappings on session disconnect/timeout. All 41 tests pass.

### 🔍 Issues Found

**Minor:**
- [ ] `[MINOR]` `isWhitelisted()` implicitly trusts all IPv6 local/link-local/site-local/ULA addresses before checking the configured CIDR whitelist (`ProxyProtocolHandler.java:243`). This is intentional for proxy deployments where the proxy runs on the same host or LAN, but it means an IPv6 whitelist entry of `::1/128` effectively also trusts `fe80::/10` and `fc00::/7`. Document this behavior prominently in the config comment so operators are aware.
- [ ] `[MINOR]` Test `testTcpIpv4ProxyHeaderFromMappedProxyIsDropped` declares `realAddress` but never uses it, and does not create a prior proxy mapping before sending the TCP header — it tests the unmapped whitelisted sender path rather than the mapped-proxy path as the name suggests (`ProxyProtocolHandlerTest.java:415`).

### 💡 Recommendations
- The `clearProxyProtocolMapping()` + `getProxyProtocolHandler()` helper pattern in `RakNetInterface` is clean and well-structured. The fallback to parent pipeline handles the edge case where the handler is on the server channel rather than the child channel.
- `RAK_PROXY_PROTOCOL` set to `false` with the comment explaining double-decode avoidance is a good defensive measure.

### 📝 Notes
- RakNet library upgrade from `1.0.0.CR4` to `1.1.0.CR1` is a coordinated API change: `RakServerRateLimiter.blockAddress()` now takes `InetSocketAddress` instead of `InetAddress`. The `new InetSocketAddress(address, 0)` wrapping is the correct adaptation.
- The `@ChannelHandler.Sharable` removal is correct — each child channel now gets its own `ProxyProtocolHandler` instance instead of sharing one.

---
*🤖 AI Review*

tachibana-kanade-angel (approved):

PR #752, opened 2026-06-17
[Multiversion] Support for 1.26.30

lt-name:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
PR adds Minecraft 1.26.30 (protocol v1001) support including new packet types, updated resource data, PyRpc codec refactoring, command block entities, dispenser behavior overhaul, and extensive regression tests. However, several blocking issues were identified in the v1001 inventory transaction encoding and packet pool versioning that are likely to cause gameplay failures on 1.26.30.

### 🔍 Issues Found

**Blocking:**
- [ ] `[CRITICAL]` v1001 inventory transaction actions drop craft-slot windowId (`NetworkInventoryAction.java:78-101`)
  - The v1001 write path (line 123-134) sends `putBoolean(true), putBoolean(false)` for `SOURCE_CRAFT_SLOT`, which means windowId is **not written** to the wire. The read path (line 78-81) follows the same logic and skips reading windowId for craft-slot actions. This leaves `windowId = 0`, and `updateTransactionPartFlags` (line 174-179) checks against negative windowId constants (`SOURCE_TYPE_CRAFTING_RESULT = -1`, etc.). Result: v1001 crafting, loom, stonecutter, and smithing actions will fail to route as crafting and be rejected as unmatched inventory actions.

- [ ] `[CRITICAL]` v1001 `InventoryTransactionPacket` legacy slot blobs decoded without bounds (`InventoryTransactionPacket.java:293-299`)
  - `readLegacySlots()` reads an unbounded slot count and trusts each client-supplied `bufLen` for `BinaryStream#get(int)`. A malicious v1001 packet could force large raw copies before validation. The older Cloudburst helper caps this field; this path should cap both slot count and byte-array length.

- [ ] `[CRITICAL]` v1001-only packets registered into shared packet pool (`Network.java:785-788`)
  - New v1001 packet IDs 348-350 (`ClientboundUpdateSoundDataPacket`, `SendPartyDestinationCookiePacket`, `PartyDestinationCookieResponsePacket`) are registered into `packetPoolCurrent` which is shared by all protocols >= v1_21_80 (line 434). This makes 1.21.80–1.26.20 clients appear to support v1001 packet IDs. If server code sends one of these by class or an older client sends those IDs, packet compatibility breaks. Needs either version-specific pools or send/decode guards for v1001-only packets.

- [ ] `[HIGH]` v1001 `BossEventPacket` always writes all fields (`BossEventPacket.java:101-108`)
  - The v1001 encode path writes `playerEid`, `type`, `title`, `filteredTitle`, `healthPercent`, `color`, and `overlay` for every event type, while the pre-v1001 path only writes fields required by the action. If v1001 still has action variants (hide, health update, title update), this may send default/empty fields unintentionally. Existing call sites like `DummyBossBar.java:208` may be affected. Recommend adding explicit tests for every boss action at v1001.

**Minor:**
- [ ] `[MINOR]` `BlockDropper.java:169,195` — Chinese comments in an otherwise mostly English code path. Consider matching local style.
- [ ] `[MINOR]` `BlockDropper.java:223` — Uses `0.20000000298023224`; `0.2` is clearer unless this exact float widening is intentional (matches vanilla behavior).

### 💡 Recommendations
- Consider adding bounds checks to `readLegacySlots()` before allocating memory based on client-supplied lengths.
- The v1001 inventory transaction write/read paths for `SOURCE_CRAFT_SLOT` need verification against the actual Bedrock protocol spec — the current asymmetry (not writing windowId) will break crafting on v1001 clients.
- Add version guards or a separate packet pool for v1001-only packets to prevent cross-version confusion.

### 📝 Notes
- `git diff --check master...feat/v1.26.30` passed (no whitespace errors).
- Could not run the full test suite locally (Gradle/Maven dependencies not cached offline). Recommend running CI before merge.
- Regression test coverage is solid — multiple new test files for packet encode/decode, crafting, command blocks, etc.

### 📋 Status
**[Status] AI Confirmed Pending | AI 确认待核**

---
*🤖 AI Review*

PR #757, opened 2026-06-21
fix: make command enum and boolean parameters case-insensitive

lt-name:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary

The PR makes command enum and boolean parameters case-insensitive, aligning with vanilla Bedrock client behavior. Both `BooleanNode` and `EnumNode` are correctly modified.

### 🔍 Issues Found

**Minor:**
- [ ] `[MINOR]` `EnumNode.fill()` now uses O(n) linear scan per `equalsIgnoreCase` call instead of the previous O(1) `Set.contains` lookup. For typical Minecraft command enums (small value sets), this is negligible. But if future enums grow large, consider building a `Map<String, String>` (lowercase → canonical) for O(1) case-insensitive lookup. (`EnumNode.java:25-30`)

### 💡 Recommendations
- The normalization approach (returning canonical enum casing) is solid — downstream code like `GameRule.parseString` and `Server.getGamemodeFromString` will get stable, predictable keys.
- `BooleanNode` correctly delegates to `Boolean.parseBoolean(arg)` with the original arg, which already handles `TRUE`/`True`/`true` uniformly. No issues there.

### 📝 Notes
- `ENUM_BOOLEAN` is defined as `{"true", "false"}` (lowercase), so the `toLowerCase(Locale.ROOT)` approach in `BooleanNode` is correct.
- Only `BooleanNode` and `EnumNode` among `ParamNode` subclasses perform enum-type matching; other subclasses (`IntNode`, `FloatNode`, `StringNode`, etc.) don't need this change.

---
*🤖 AI Review*

tachibana-kanade-angel (approved):

PR #776, opened 2026-06-26
Added support for dynamic custom item IDs (fallback fix)

laww278-gif:
This change improves item handling by allowing dynamic/custom item IDs without requiring manual registration.

- Added bounds check when accessing Item.list to prevent crashes on invalid IDs
- Ensures unknown or out-of-range IDs fall back to a generic Item instead of failing
- Enables plugins to define temporary custom items at runtime without modifying core code

This makes it possible to use custom IDs (e.g. 5000+) safely and supports dynamic item systems.

### DIFF
--- a/src/main/java/cn/nukkit/item/Item.java
+++ b/src/main/java/cn/nukkit/item/Item.java
@@ -1066,7 +1066,11 @@ public class Item implements Cloneable, BlockID, ItemID, ItemNamespaceId, Protoc
                     c = Block.list[blockId];
                 }
             } else {
-                c = list[id];
+                if (id >= 0 && id < list.length) {
+                    c = list[id];
+                } else {
+                    c = null;
+                }
             }
             Item item;
 


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
No blocking issues found. The changes look reasonable overall. This PR adds a bounds check on `Item.list[id]` access to prevent `ArrayIndexOutOfBoundsException` for out-of-range item IDs (e.g. custom IDs 5000+), falling back to a generic `Item` instead of crashing.

### 🔍 Issues Found

**Minor:**
- [ ] `[MINOR]` The bounds check `id >= 0 && id < list.length` is correct for the `else` branch, but the negative-ID branch (`id < 0`) still performs unchecked `Block.list[blockId]` access where `blockId = 255 - id`. This is pre-existing and not introduced by this PR, but worth noting for completeness. (Item.java:1061-1067)
  - Consider adding a similar bounds guard on `Block.list[blockId]` in a follow-up.

### 💡 Recommendations
- Add a unit test for `Item.get()` with out-of-range positive IDs (e.g. `Item.get(99999)`) to verify the fallback behavior.
- Consider documenting that `Item.list` can be null before `Item.init()`, as the new guard checks length but not nullability.

### 📝 Notes
The fix is minimal and targeted — a 6-line defensive bounds check. The null `c` case is properly handled downstream at line 1081 (`item = new Item(id, meta, count)`), so out-of-range IDs will safely produce a generic Item rather than crash.

---
*🤖 AI Review*

tachibana-kanade-angel:
@lt-name ready to merge

tachibana-kanade-angel (approved):

lt-name:
I believe this only improves performance by avoiding the overhead caused by exception handling. It doesn't actually implement the functionality you described.

PR #779, opened 2026-06-29
fix(minecart): ignite TNT minecart when hit by flaming projectile

### LINKED ISSUE #778: Incendiary bow destroys TNT minecart instead of igniting it
# 🐞 Bug Feedback | BUG反馈

The incendiary bow doesn’t set the cart with dynamite on fire, it kills it

13:13:05 [main] [INFO] #########################################
13:13:05 [main] [INFO] Nukkit-MOT
13:13:05 [main] [INFO] Build: master/948e464
13:13:05 [main] [INFO] Multiversion: Up to version 1.26.30
13:13:05 [main] [INFO] #########################################

lt-name:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
The approach is clean and localized — a single `attack()` override on `EntityMinecartTNT` with the speed-squared explosion strength matching the existing API. However, there are **2 blocking issues** that prevent the primary use case (Flame bow arrows) from working, plus a `GameRule` guard bypass.

### 🔍 Issues Found

**Blocking:**
- [ ] `[CRITICAL]` `getDamager()` misses shooter-owned projectiles (`EntityMinecartTNT.java:104`)
  - Nukkit uses `EntityDamageByChildEntityEvent` when a projectile has a shooter. In that event, `getDamager()` returns the **shooter** (Player), not the projectile (Arrow). A Flame bow arrow appears as `Player → Arrow → TNT minecart`, so this check on `getDamager() instanceof EntityProjectile` falls through to `super.attack()` — the primary advertised case does not work.
  - Suggested fix:
    ```java
    Entity direct = damageByEntityEvent instanceof EntityDamageByChildEntityEvent childEvent
            ? childEvent.getChild()
            : damageByEntityEvent.getDamager();
    ```
    Then check `direct instanceof EntityProjectile projectile && projectile.isOnFire()`.

- [ ] `[CRITICAL]` Bypasses `GameRule.TNT_EXPLODES` guard (`EntityMinecartTNT.java:110`)
  - The existing fuse path checks `TNT_EXPLODES` before exploding (line 80). This new path explodes the cart even when TNT explosions are disabled via GameRule.

**Minor:**
- [ ] `[MINOR]` `EntityProjectile` is broader than vanilla Javas `AbstractArrow` check. May include fireballs or other on-fire projectiles. If Bedrock-style "any flaming projectile" is intended, this is fine — but the Javadoc says it mirrors vanilla `MinecartTNT#hurtServer`, so the type scope should be intentional.
- [ ] `[MINOR]` If `EntityExplosionPrimeEvent` is cancelled, `explode(speedSqr)` returns without closing the minecart, yet `attack()` still returns `true`. The existing fuse path closes the minecart after attempting explosion. Edge-case behavior difference worth aligning.
- [ ] `[MINOR]` The override skips `super.attack(source)` entirely on the special path, so `EntityDamageEvent`/`VehicleDamageEvent`/`VehicleDestroyEvent` listeners do not see this hit. `EntityExplosionPrimeEvent` still fires, but this is a plugin API behavior change.

### 💡 Recommendations
1. Use `EntityDamageByChildEntityEvent` to correctly extract the projectile from the event chain
2. Add `GameRule.TNT_EXPLODES` check before calling `explode()`
3. Consider whether the type scope should be `EntityArrow` (vanilla Java) or `EntityProjectile` (broader Bedrock-style) — document the intent

### 📝 Notes
The `motionX/Y/Z` speed-squared calculation is correct and matches the existing `explode(double)` API.

---
*🤖 AI Review*

lt-name:

tachibana-kanade-angel:
## 📋 PR Review (Re-review)

### ✅ Summary
The PR adds the intended flaming-projectile TNT minecart ignition path, and the 3D `speedSqr` calculation matches vanilla-style projectile velocity handling. However, the new `attack()` override bypasses the normal damage/vehicle event flow, preventing plugins from cancelling or intercepting this damage path.

### 🔍 Issues Found

**Blocking:**
- [ ] `[CRITICAL]` `attack()` explodes before calling `super.attack(source)`, so `EntityDamageEvent`, `VehicleDamageEvent`, and `VehicleDestroyEvent` are never fired for flaming projectiles (`EntityMinecartTNT.java:104`)
  - Plugins cannot cancel/protect this damage path, allowing flame arrows to detonate TNT minecarts even when normal projectile damage would be cancelled by protection plugins.

**Minor:**
- [ ] `[MINOR]` If `TNT_EXPLODES` game rule is `false`, the method returns `true` without exploding, closing, or applying normal damage (`EntityMinecartTNT.java:113`)
  - A flaming projectile is consumed but the minecart is left untouched, which is inconsistent with the existing fuse path that closes the minecart after a failed explosion check.

### 💡 Recommendations
Run the normal damage/event path first (`super.attack(source)`), and only detonate if the attack is accepted (not cancelled). Also make the no-explosion/cancelled-prime behavior consistent with existing TNT minecart fuse handling.

### 📝 Notes
No concurrent modification or null-check issues. The projectile direct-entity selection via `EntityDamageByChildEntityEvent#getChild()` is appropriate.

Previous review blocking issues (getDamager() misses projectiles, GameRule bypass) appear to have been addressed in the updated code.

---
*🤖 AI Review*

PR #780, opened 2026-06-30
Better ways to handle 113 protocol batchpacket

yefengeeeeeeeeeee:


tachibana-kanade-angel:
## 📋 PR Review

### ✅ Summary
This PR adds two features: (1) fire-projectile detonation of TNT minecarts matching vanilla behavior, and (2) v1.1 protocol batch packet detection. The TNT minecart logic is well-structured, but the v1.1 batch detection has several issues that need attention before merging.

### 🔍 Issues Found

**Blocking:**
- [ ] `[CRITICAL]` Buffer boundary issue — no length guard before reading bytes at offsets 1 and 2 (`RakNetPlayerSession.java:116-117`)
  - `buffer.getUnsignedByte(1)` and `buffer.getUnsignedByte(2)` will throw `IndexOutOfBoundsException` on malformed packets shorter than 3 bytes (after the 0xfe prefix byte). Add a `buffer.readableBytes() >= 3` guard before the check.
- [ ] `[CRITICAL]` `processBatch(...)` return value is ignored; failure silently drops packets (`RakNetPlayerSession.java:121`)
  - Normal batch processing (line 162) disconnects on malformed packets via `this.disconnect("Sent malformed packet")`. The v1.1 path silently returns, so a corrupted v1.1 batch is never detected. Either check the return value or let execution fall through to normal error handling.
- [ ] `[CRITICAL]` `EntityMinecartTNT.attack()` returns `true` when `TNT_EXPLODES` is false (`EntityMinecartTNT.java:115-117`)
  - The arrow is consumed but the cart takes no damage and no event is fired. Should either fall through to `super.attack()` or at least call `super.attack()` when the game rule prevents explosion.

**Minor:**
- [ ] `[MINOR]` Hardcoded zlib header `0x78 0xda` is fragile for protocol detection (`RakNetPlayerSession.java:117`)
  - `0x78 0xda` specifically indicates zlib compression level 9, default strategy. Other compression levels produce different second bytes (e.g., `0x9c` for level 6 default). Consider checking only `0x78` as the zlib magic byte, or using a range check for the second byte's CMF/FLG validity.
- [ ] `[MINOR]` Detection runs before decryption stage (`RakNetPlayerSession.java:116`)
  - For encrypted v1.1 sessions (if any exist), this would misidentify encrypted bytes as zlib headers. In practice v1.1 predates encryption, but adding a `this.decryptionCipher == null` guard would be safer.

### 💡 Recommendations
- Consider adding a `buffer.markReaderIndex()` / `resetReaderIndex()` pattern instead of hard-coding byte offsets, to make the code resilient to future packet format changes.
- The projectile child-entity handling in `EntityMinecartTNT` is correct: `EntityDamageByChildEntityEvent#getChild()` properly catches Flame arrows where `getDamager()` returns the shooter, and `projectile.isOnFire()` matches the creation logic.

### 📝 Notes
- Gradle tests could not be run in the review environment (read-only `.gradle` directory). Recommend running the test suite locally before merging.
- The PR description is empty — consider adding context about which vanilla version this behavior is being matched to.

---
*🤖 AI Review*

src/main/java/cn/nukkit/entity/item/EntityMinecartTNT.java
```diff
@@ -92,6 +95,31 @@ public void activate(int x, int y, int z, boolean flag) {
         this.fuse = 80;
     }
 
+    /**
+     * Mirrors vanilla {@code MinecartTNT#hurtServer}: a projectile that is
+     * itself on fire (e.g. a Flame-enchanted bow's arrow) primes and
+     * immediately detonates the TNT minecart instead of merely damaging it.
+     */
+    @Override
+    public boolean attack(EntityDamageEvent source) {
```
lt-name, line 104: Please do not include irrelevant modifications.

src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
```diff
@@ -113,6 +113,17 @@ protected void channelRead0(ChannelHandlerContext channelHandlerContext, RakMess
...
 
             byte[] packetBuffer;
+
+            byte oldPacketId = (byte) (buffer.getUnsignedByte(1) & 0xff);
+            if(oldPacketId == (byte) 0x78 && (byte) (buffer.getUnsignedByte(2) & 0xff) == (byte) 0xda){ //1.1的数据包，0xfe后面必定是 0x78 0xda这两个字节
+                // 1.1 batchpacket 处理
+                packetBuffer = new byte[buffer.readableBytes()];
+                buffer.readBytes(packetBuffer);
+                int protocol = ProtocolInfo.v1_1_0;
+                this.server.getNetwork().processBatch(packetBuffer, this.inbound, compressionIn, protocol, this.player);
```
lt-name, line 123: The version number passed here should be the version of raknetProtocol.

lt-name (changes requested):

lt-name:
I believe this pull request is entirely generated by AI and contains bug fixes, making it unsuitable for merging.

yefengeeeeeeeeeee:
I'm sorry for some careless checks and errors, I will correct them

PR #790, opened 2026-07-12
feat(resourcepack): add CDN configuration and migrate pack storage

lt-name:
- add packs.yml and CDN URL API
- unify resource and behavior pack directories
- migrate legacy keys and harden NetEase migration


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR introduces CDN configuration for resource packs and migrates legacy pack storage to a unified directory structure. The changes touch 11 files and add a new `packs.yml` configuration, `ResourcePackMigration` utility, and corresponding tests. Codex found **5 blocking issues** and 6 non-blocking issues. Several concerns relate to migration safety and data integrity.

### Issues Found

#### Blocking Issues

1. **Migration is not restart-safe** ([ResourcePackMigration.java:97](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/utils/serverconfig/ResourcePackMigration.java#L97), [ResourcePackMigration.java:108](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/utils/serverconfig/ResourcePackMigration.java#L108))
   - The pack is moved before its key is persisted. If the process fails between those operations, the next run treats the remaining key as an orphan but never associates it with the already-moved pack.
   - When the destination already exists, the legacy source remains but the legacy loader has been removed, so that source pack is no longer served.

2. **Manifest discovery can assign a key to the wrong UUID** ([ResourcePackMigration.java:180](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/utils/serverconfig/ResourcePackMigration.java#L180))
   - The fallback accepts any entry whose full name ends in `manifest.json` (e.g. `fake_manifest.json`), unlike the actual pack loader which requires exact basename matching.
   - A preceding unrelated JSON file can cause the key to be written under the wrong UUID. The extracted UUID is not validated before deleting the key.

3. **Invalid `packs.yml` can crash server startup** ([ResourcePackMigration.java:150](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/utils/serverconfig/ResourcePackMigration.java#L150), [ResourcePackManager.java:129](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/resourcepacks/ResourcePackManager.java#L129))
   - Both configuration loads can throw unchecked SnakeYAML/parsing exceptions, but migration catches only `IOException`, while normal config application has no guard at all.
   - Migration should validate/load configuration before moving anything and handle invalid structure explicitly.

4. **Encryption keys are written verbatim to debug logs** ([JarPluginResourcePack.java:87](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/resourcepacks/JarPluginResourcePack.java#L87))
   - Logs the complete content key, exposing protected pack material to log readers and aggregation systems. This line should be removed or redacted.

5. **Removing `JarPluginResourcePack.encryptionKey` is a source/binary compatibility break** ([AbstractResourcePack.java:17](/root/projects/Nukkit-MOT/src/main/java/cn/nukkit/resourcepacks/AbstractResourcePack.java#L17))
   - It was a `protected` field available to subclasses; its replacement is `private`. Existing plugin subclasses that access the protected field will fail compilation or throw an access/linkage error.

#### Non-blocking Issues

1. **Directory-pack keys are not migrated** — The directory branch moves only the directory; key processing exists only in the archive-file branch. Any adjacent `<directory>.key` is left orphaned.

2. **Configuration values are not validated or normalized** — Arbitrary CDN schemes, malformed URLs, whitespace, YAML `null`, and invalid encryption-key lengths are accepted and sent to clients.

3. **Empty configured key does not mean "no encryption"** — `ResourcePackManager` ignores empty values, so it cannot clear a key embedded in a plugin JAR or loaded from a remaining sidecar file.

4. **Configuration silently does nothing for custom `ResourcePack` implementations** — The new setters are default no-ops in `ResourcePack.java`, with no warning when the returned pack cannot retain configured values.

5. **`ResourcePackManager(File path)` uses global config path** — Unexpectedly uses `Nukkit.DATA_PATH/resource_packs/packs.yml` rather than configuration adjacent to the supplied path.

6. **Core CDN behavior is untested** — No tests for applying `cdn`/`key`, packet output, malformed YAML, URL/key validation, clearing overrides, or successful recovery after interrupted migration.

### Recommendations
1. Make migration atomic or restart-safe: validate config before moving files, write keys before moving packs, and handle partial-failure recovery.
2. Tighten manifest discovery to match exact basename `manifest.json` / `pack_manifest.json` only.
3. Add exception handling for invalid YAML in both migration and normal config loading paths.
4. Remove or redact encryption key logging.
5. Restore `protected` field access or document the breaking change with a migration path for plugins.

### Notes
- `git diff --check` passed
- Tests could not be fully executed due to Gradle cache/network restrictions in the review environment
- An unrelated `BlockBush` change (`a816d13c2`) is included in the diff — no defect found in that change

---
[Status] AI Confirmed Pending | AI 确认待核
🤖 AI Review

PR #793, opened 2026-07-13
feat: Plugin Annotation

Kanelucky:
# Plugin Annotation

Add an annotation processor for @PluginMeta that automatically generates plugin.yml during compilation.

## Feature
`@PluginMeta`
```java
@PluginMeta(
        name = "ExamplePlugin",
        version = "1.0.0",
        api = {"1.0.0"},
        authors = {"NotMe"},
        description = "An example plugin using @PluginMeta.",
        website = "https://trustmebro.com",
        prefix = "Example",
        order = PluginLoadOrder.POSTWORLD,
        depend = {@Dependency(name = "SomeOtherPlugin")},
        softDepend = {@Dependency(name = "SomeOtherPlugin")},
        loadBefore = {"SomeOtherPlugin"}
)
public class ExamplePlugin extends PluginBase {
    public void onEnable() {
    }
}
```

### DIFF
--- a/src/main/java/cn/nukkit/plugin/annotation/AnnotationProcessor.java
+++ b/src/main/java/cn/nukkit/plugin/annotation/AnnotationProcessor.java
@@ -3,6 +3,7 @@ package cn.nukkit.plugin.annotation;
 import cn.nukkit.plugin.PluginLoadOrder;
 
 import javax.annotation.processing.*;
+import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
@@ -147,5 +148,10 @@ public class AnnotationProcessor extends AbstractProcessor {
         }
         TypeMirror erasedSuper = types.erasure(superType.asType());
         return types.isAssignable(types.erasure(type.asType()), erasedSuper);
+    }
+
+    @Override
+    public SourceVersion getSupportedSourceVersion() {
+        return SourceVersion.latestSupported();
     }
 }
--- a/src/main/java/cn/nukkit/plugin/annotation/AnnotationProcessor.java
+++ b/src/main/java/cn/nukkit/plugin/annotation/AnnotationProcessor.java
@@ -1,0 +1,151 @@
+package cn.nukkit.plugin.annotation;
+
+import cn.nukkit.plugin.PluginLoadOrder;
+
+import javax.annotation.processing.*;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR introduces an annotation processor for `@PluginMeta` that automatically generates `plugin.yml` during compilation. The concept is useful but **the processor is not correct as shipped** — there is a blocking build issue that will break clean Maven builds.

### Issues Found

#### Blocking Issues
- **Service registration breaks clean Maven builds** — The `META-INF/services/javax.annotation.processing.Processor` file is copied into `target/classes` during `process-resources`, before `compile`. Javac then discovers the service but cannot load `AnnotationProcessor` because that class has not been compiled yet. This was reproduced directly with `javac`. The repository CI builds with Maven, so a clean PR build should fail. Fix: place annotations and processor in separate modules/artifacts, or arrange for the service file to be added only after compiling the processor.

#### Non-blocking Issues
- **Invalid plugin main classes pass validation** — `validateMain()` only checks that the element extends `PluginBase`. It accepts abstract classes, non-public classes, non-static inner classes, and classes without an accessible zero-arg constructor — all of which fail at runtime.
- **`Dependency.version` and `Dependency.optional` are silently ignored** — The annotation exposes these fields, but the processor only maps dependency names. Setting `optional = true` still produces a hard dependency. These fields should be removed or given proper semantics.
- **"Only one `@PluginMeta`" is enforced per round, not per compilation** — The `generated` flag short-circuits later rounds. If another processor generates a second `@PluginMeta` class in a later round, this processor ignores it.
- **Generated descriptor values are insufficiently validated** — Empty `name`, `version`, or `api` values produce malformed YAML. Empty dependency names and invalid plugin-name characters are not rejected.
- **Coexistence with manual `plugin.yml` is undefined** — `createResource()` always creates `plugin.yml`, which may overwrite or duplicate an existing descriptor. The annotation does not model `commands` and `permissions`, making coexistence tempting but undefined.
- **No processor tests** — At minimum, compile-testing coverage is needed for generation, invalid main classes, multiple processing rounds, dependency semantics, escaping, and existing `plugin.yml`.

### Recommendations
1. Separate the annotation processor into its own module/artifact to fix the Maven build issue
2. Strengthen `validateMain()` to require public, non-abstract classes with accessible no-arg constructors
3. Remove or implement `Dependency.version` and `Dependency.optional`
4. Add compile-testing (e.g., Google Compile Testing) for the processor
5. Either reject or merge with existing `plugin.yml` — do not silently overwrite

### Notes
- No direct security vulnerabilities found. Output is constrained to `CLASS_OUTPUT/plugin.yml` with proper escaping.
- Consider renaming `AnnotationProcessor` to `PluginMetaProcessor` for clarity.
- Consider using `RetentionPolicy.SOURCE` for annotations unless runtime reflection is needed.

---
🤖 AI Review

Kanelucky:
i reproduced the issue with a clean maven build. I dont currently know how to solve it within the current project structure. Do you have any suggestions on how this should be handled in nukkit-mot?

PR #797, opened 2026-07-15
feat: shulker box now can change color

Sunch233:
The structure of `shulker_box` type and `shapeless` type is the same in JSON. So can register `shulker_box` as `shapeless` directly.

## Type `shulker_box`
```json
        {
            "id": "bundle_dye_0",
            "type": 5,
            "input": [
                {
                    "type": "default",
                    "count": 1,
                    "itemId": 267,
                    "auxValue": 32767
                },
                {
                    "type": "default",
                    "count": 1,
                    "itemId": 445,
                    "auxValue": 32767
                }
            ],
            "output": [
                {
                    "legacyId": 264,
                    "id": "minecraft:black_bundle"
                }
            ],
            "block": "crafting_table",
            "priority": 0
        },
```

## type `shapeless`
```json
        {
            "id": "minecraft:cobweb_to_string",
            "type": 0,
            "input": [
                {
                    "type": "default",
                    "count": 1,
                    "itemId": 358,
                    "auxValue": 32767
                }
            ],
            "output": [
                {
                    "legacyId": 358,
                    "id": "minecraft:string"
                }
            ],
            "block": "deprecated",
            "priority": 0
        }
```

## before
image

## after

image

### DIFF
--- a/src/main/java/cn/nukkit/inventory/CraftingManager.java
+++ b/src/main/java/cn/nukkit/inventory/CraftingManager.java
@@ -164,8 +164,8 @@ public class CraftingManager {
                         break;
                     case 4: // multi (hardcoded)
                         break;
-                    case 5: // shulker_box
-                        loadShapelessRecipe(itemMapping, recipe);
+                    case 5: // shulker_box (UserDataShapelessRecipe): shapeless 结构，但需保留输入 NBT
+                        loadUserDataShapelessRecipe(itemMapping, recipe);
                         break;
                 }
             } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds a one-line call to `loadShapelessRecipe()` in the `case 5` (shulker_box) branch of `CraftingManager`, enabling 1,084 type-5 recipes (shulker boxes, bundles, harnesses) to be registered as shapeless recipes. The structural similarity between type-5 and shapeless recipes makes this approach tempting, but it introduces critical functional and security gaps.

### Issues Found
#### Blocking Issues
- **Output item is hardcoded empty — container contents and custom names are lost.** `loadShapelessRecipe` creates the output `Item` by string ID alone (e.g. `minecraft:black_shulker_box`). Any NBT data on the container being recolored — items stored inside, custom name, lock tag — is discarded. A shulker box with contents becomes an empty one after recoloring. The same applies to bundles and harnesses that carry NBT. This is data loss in normal gameplay.
- **No server-side output transformation or validation of client-submitted results.** `CraftRecipeActionProcessor` and `CraftRecipeAutoProcessor` accept whatever output the client sends for shapeless recipes without verifying that the NBT matches what the server should produce. A malicious client could fabricate arbitrary shulker box NBT (e.g. injecting items that should not exist). The server must generate and validate the output from the consumed container.

#### Non-blocking Issues
- **Recipe type identity is lost.** All 1,084 type-5 entries are registered as `ShapelessRecipe` rather than a dedicated `ShulkerBoxRecipe` (or a type-tagged variant). This means the wire type sent to the client in `CraftingDataPacket` will be `SHAPELESS` (0) instead of `SHULKER_BOX` (5), which may cause client-side recipe book display issues or mismatches.
- **No tests.** Enabling over a thousand recipes with a single line and no integration or unit test makes regressions hard to catch.

### Recommendations
1. Introduce a `ShulkerBoxRecipe` subclass of `ShapelessRecipe` (or add a type field) that preserves recipe type 5 in the wire format.
2. Add server-side result generation: copy permitted NBT (contents, custom name) from the consumed container input to the recolored output, then validate against what the client submitted.
3. Add tests covering: wire type 5 serialization, NBT preservation on filled/named shulker boxes, bundle contents preservation, and rejection of fabricated output NBT.
4. Consider splitting the PR: empty-container recoloring (simple) vs. NBT-preserving recoloring (complex) can be separate incremental steps.

### Notes
- The structural similarity between type-5 and shapeless JSON is real, but "same parse structure" ≠ "same runtime semantics." The missing NBT handling is the critical gap.
- The Gradle test suite could not be executed in the Codex environment due to a read-only Gradle lock directory; the above findings are based on static code analysis.

---
🤖 AI Review

lt-name:
To prevent the loss of NBT data, I plan to use the PNX solution. I will modify and commit it to this branch when I have time.

tachibana-kanade-angel:

PR #802, opened 2026-07-23
feat(level): async chunk loading, decoding and unload saves

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR implements async chunk loading/decoding with a backpressure mechanism for chunk saves. The architecture is well-designed: off-thread read+decode, main-thread mounting via `doTick`, proper invalidation on unload/replace/reload, and bounded executor shutdown. The pending-write slot system with per-chunk total ordering and newest-wins semantics is sound. Tests are comprehensive (dedup, rejection, queue-full, invalidation, race conditions, backpressure).

### Issues Found

#### Blocking Issues
None.

#### Non-blocking Issues
1. **BaseLevelProvider.putChunkIfAbsent** sets `provider`/`position` on the chunk *before* the `putIfAbsent` check. If the slot already has a chunk, the new chunk is mutated (provider/position set) but never inserted — it's left in an inconsistent state. Move the `setProvider`/`setPosition` calls to after the `putIfAbsent` succeeds (matching LevelDBProvider's correct ordering).

2. **LevelDBProvider.readChunkOffThread** holds the `dbReadCloseLock` read lock during the entire decode, including the `commitPendingWrite` call which performs a `db.write`. Under high load this serializes reads behind writes. Consider committing the pending write before acquiring the read lock, or restructuring so the write doesn't block concurrent reads.

3. **LevelDBProvider.putChunkIfAbsent** throws `IllegalArgumentException` for non-LevelDB chunks. This is fine for the async path (LevelDB-only), but worth noting for future provider implementations that may need a no-op or fallback.

### Recommendations
1. In `BaseLevelProvider.putChunkIfAbsent`, reorder to check `putIfAbsent` first, then set provider/position only on success.
2. Consider whether `readChunkOffThread` needs to commit pending writes at all — the read path shouldn't block on writes; the read lock already protects against close.
3. The `PendingWrite.retries = 0` reset in `stagePendingWrite` is correct (per-batch retries), but add a brief comment to document the intent.

### Notes
- The ticking-time concern for deferred entries during save is not a real bug in practice: async-loaded chunks are not in the provider cache until `mountChunk` puts them, and mounting replays deferred updates first. The save window where deferred entries would be present is effectively zero.
- Executor shutdown is correctly bounded (10s graceful + 5s forced) with interrupt flag restoration.
- The `dbReadCloseLock` read/write pattern correctly guards against use-after-free of the LevelDB handle.
- Config migration (`async-chunk-loading` → `async-chunks`) and `maxPendingChunkWrites` defaults are clean.
- MockServer correctly initializes `maxPendingChunkWrites = 128` to avoid Mockito skipping field initializers.

@lt-name 可以合并

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel (approved):
Async chunk loading, decoding and unload save backpressure implemented correctly. No blocking issues.

PR #803, opened 2026-07-25
fix(skin): prevent NetEase 1.21.124 player-like NPC invisibility and tab-list lingering

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR addresses NPC invisibility and tab-list lingering issues on NetEase V860 (1.21.124) clients. The approach introduces a `sentSkins` tracking set per viewer and a `PlayerEntitySkinSender` utility to manage skin lifecycle. The core idea is sound, but there are two blocking issues that could reintroduce the very bugs this PR aims to fix.

### Issues Found

#### Blocking Issues

1. **Hide/show reintroduces duplicate-ADD invisibility bug**
   `Player.java:678` removes the UUID from `sentSkins`, then only despawns the entity. For real players, `EntityHuman.java:392` does not remove the player-list entry; the base despawn only sends `RemoveEntityPacket`. When `showPlayer()` is called, `spawnTo()` sees the cleared set and sends another `PlayerList(ADD)` without a preceding `REMOVE` — the exact sequence this PR says makes NetEase players invisible.
   - **Fix**: Either retain `sentSkins` across hide/show, or explicitly send `PlayerList(REMOVE)` before clearing it. Add a V860 hide→show regression test.

2. **`sentSkins` can diverge from what the client actually received**
   Several transitions update tracking independently of packet delivery:
   - `Server.java:1396` marks every viewer after broadcast, even if `dataPacket()` returned false
   - `Server.java:1481` ignores failed/cancelled full-list chunks, then marks every current player as sent
   - `Server.java:1448` removes entries without clearing successful removals from the tracker
   - `PlayerEntitySkinSender.java:105` can successfully send `REMOVE`, fail the replacement `ADD`, and leave the entry permanently absent
   
   Because `EntityHuman.java:331` treats this set as authoritative, a cancelled packet or plugin use of the public player-list APIs can suppress a necessary future ADD and make an entity invisible.
   - **Fix**: Centralize per-viewer ADD/REMOVE sending and mutate `sentSkins` only after accepted delivery. Replacement failure needs rollback or retry behavior.

#### Non-blocking Issues

1. **Unbounded per-viewer generation storage**
   `PlayerEntitySkinSender.java:35` retains every NPC UUID encountered by a connected viewer. The outer weak key only helps after the viewer is collected; the inner map never removes completed UUIDs. High NPC UUID churn can steadily grow heap usage.

2. **Tests miss important behavior**
   - `EntityHumanSkinLifecycleTest.java:48` stubs the five-argument overload, but the real implementation delegates to an unstubbed six-argument method — legacy tests observe only `REMOVE`, not the required initial `ADD`.
   - Recording players override `dataPacket()` at `PlayerSkinLifecycleTest.java:162`, so no test encodes packets or verifies the V860 `fullSkinId` workaround in `BinaryStream.java:397`.

3. **Performance and maintainability**
   - Each V860 NPC handshake allocates an empty 8 KiB skin, sends the full representation, then sends the real skin again, and schedules a task per viewer/NPC. The extra `UUID.randomUUID()` during V860 skin encoding uses a cryptographic generator — deserves load testing on NPC-heavy servers.
   - `sentSkins` should be private and exposed through lifecycle methods; its current public mutability makes state invariants difficult to enforce.

### Recommendations
- Address the two blocking issues before merging — both can cause entity invisibility in production
- Add integration tests for hide→show cycles on V860 clients
- Consider making `sentSkins` private with controlled accessors

### Notes
- Security: No direct authentication/authorization/data-exposure vulnerability found
- `git diff --check master...HEAD` passed
- Tests could not run in sandbox environment

---
🤖 AI Review

PR #806, opened 2026-07-25
feat(plugin): auto-resolve plugin.yml libraries with per-plugin ClassLoader isolation

lt-name:


lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. PR adds automatic Maven library resolution from `plugin.yml` declarations with per-plugin ClassLoader isolation (own URL[] before global scan). Coordinate validation, XXE hardening, atomic downloads, and good test coverage for the core isolation semantics are all present.

### Issues Found

#### Non-blocking Issues
1. **Incomplete Maven model resolution** — `parsePom` does not resolve `${...}` properties, `<dependencyManagement>`, BOMs, or parent POM inheritance. Dependencies with unresolvable placeholders are silently skipped (line 316-319). Acceptable for the common case where plugins declare simple direct coordinates, but could be improved with a proper Maven model builder.
2. **Order-dependent transitive conflict resolution** — The resolver uses sequential DFS over direct dependencies. In rare cases where two direct deps transitively pull different versions of the same artifact, the version resolved first wins regardless of depth. A breadth-first approach with depth tracking would be more robust.
3. **Thread-safety of concurrent resolution** — `ensureArtifactDownloaded` has a check-then-act race on the `.tmp` file. Not a practical concern since Nukkit plugin loading is typically sequential during startup.
4. **Interrupted exception handling** — Broad catch at line 233 swallows `InterruptedException` without restoring the interrupt flag.
5. **Public API changes** — `LibraryLoader` constructor is now private and `getBaseFolder()` behavior changed. Should be documented as intentional API break or provide deprecation window.

### Recommendations
- Consider using Maven Resolver/Aether for production-grade dependency resolution in a future iteration.
- Add integration tests for `extraUrls` with actual resolved library JARs.
- Document the API changes (private constructor, `getBaseFolder()` path change) in release notes.

### Notes
- XXE protection in POM parsing is well-implemented with dual-layer defense.
- Test coverage includes isolation semantics, nearest-wins regression, XXE regression, and download directory creation — good foundation.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

PR #807, opened 2026-07-26
feat: improve mob ai

Kanelucky:
Vanilla

https://github.com/user-attachments/assets/8173a4db-d36e-4787-8a17-0504f756a126

Nukkit-MOT (now)

https://github.com/user-attachments/assets/fe6b349e-5d03-482f-a47b-0a0e5da58e25

Nukkit-MOT (after)

https://github.com/user-attachments/assets/fc8cb0ce-1f58-43d7-ae84-bb6405003257

Look carefully at the zombie's head


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR replaces manual yaw calculation with `lookAt(this.followTarget)` to improve mob head rotation toward the entity being followed. The intent is reasonable, but the implementation has critical issues that prevent it from working as intended.

### Issues Found
#### Blocking Issues
- **Yaw overwritten in same tick**: `lookAt(followTarget)` at line 211 sets rotation, but `setBothYaw(...)` is called again at line 250 during normal movement (`stayTime <= 0`, distance > `0.001`). The second call overwrites yaw/head yaw, leaving only pitch from `lookAt()`. This means mobs will NOT reliably face their follow target — only their pitch changes.
- **NaN yaw risk**: `lookAt()` internally divides by horizontal distance (`dx / Math.sqrt(dx*dx + dz*dz)`). When entity and follow target share the same X/Z position (e.g. target directly above), denominator is zero → NaN. The `diff != 0` guard at line 210 checks the waypoint distance, not `followTarget` distance, so it does not protect against this.
- **Semantic mismatch — waypoint vs entity**: The old code used `this.target` (pathfinding waypoint) for yaw calculation. The new code uses `this.followTarget` (actual entity). When these differ, mobs may face sideways relative to their movement direction, and `checkJump()` (line 98) which uses yaw-derived facing for obstacle detection may malfunction.

#### Non-blocking Issues
- **Pitch side-effect**: `lookAt()` changes pitch in addition to yaw/head yaw. The original code only changed yaw/head yaw. For ranged mobs that use `this.pitch` for projectile aiming, this is a functional change, not just cosmetic.
- **No eye-height offset**: `lookAt()` uses base Y coordinates rather than eye height, so differently-sized entities may aim above or below the intended point.

### Recommendations
- Decide explicitly whether body rotation should follow the route waypoint or the follow entity — these serve different purposes.
- If the goal is to make mobs look at their target, use `atan2` with explicit eye/target heights and handle zero horizontal distance.
- Prevent the second `setBothYaw()` call from undoing the `lookAt()` rotation, or integrate the look-at logic into the movement calculation itself.
- Consider separating head rotation (visual) from body rotation (movement direction).

### Notes
- The PR body includes comparison videos showing Vanilla vs Nukkit-MOT behavior — helpful for understanding the intent.
- The change is a single line plus a newline at EOF, so scope is small but the implications are significant for mob AI behavior.

---
🤖 AI Review

tachibana-kanade-angel:

PR #811, opened 2026-07-26
[Multiversion] Support for 1.26.40

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds support for Minecraft Bedrock protocol 1.26.40 (v2168) to the multiversion pipeline: protocol version bumps, new runtime block/item state mappings, v2168 packet codec updates (skin, slots, item instance descriptors, MoveEntityDelta), and extensive regression tests. The multiversion structure is generally well handled — legacy protocol branches remain version-gated and the new resources (item states, block states, NBT) passed integrity checks. However, a remotely reachable memory-exhaustion vulnerability was found in the new v2168 item slot decoder.

### Issues Found
#### Blocking Issues
- **Unbounded `CanPlaceOn`/`CanDestroy` allocation in v2168 slot decoder** — [`BinaryStream.java:982`](https://github.com/MemoriesOfTime/Nukkit-MOT/blob/feat/v1.26.40/src/main/java/cn/nukkit/utils/BinaryStream.java#L982): `getSlotNewV2168` trusts client-controlled `CanPlaceOn`/`CanDestroy` counts read from item user-data and allocates `new String[count]` with no upper bound. A connected client can supply `Integer.MAX_VALUE` in a small payload (reachable via `CraftingEventPacket` at line 40, still registered in `Network.java`), causing `OutOfMemoryError` — an `Error`, not caught by the surrounding `catch (IOException)`. The older decoder enforces a `4096` cap (lines 790-808); the v2168 branch must apply the same `0..4096` validation and release the allocated `ByteBuf` in a `finally` block.

#### Non-blocking Issues
- **`DimensionDefinition` constructor break** — [`DimensionDefinition.java`](https://github.com/MemoriesOfTime/Nukkit-MOT/blob/feat/v1.26.40/src/main/java/cn/nukkit/network/protocol/types/DimensionDefinition.java): adding `dimensionType`/`packId` changes the Lombok-generated constructor from 4 to 6 arguments. Plugins compiled against the 4-arg constructor will fail with `NoSuchMethodError`. Retain a delegating 4-arg constructor with defaults.
- **`FLAG_TELEPORTING` dropped on v2168** — [`MoveEntityDeltaPacket.java`](https://github.com/MemoriesOfTime/Nukkit-MOT/blob/feat/v1.26.40/src/main/java/cn/nukkit/network/protocol/MoveEntityDeltaPacket.java): the v2168 branch decodes per-axis booleans and the `forceMove` field but never restores `FLAG_TELEPORTING` (decode) nor emits it from the flags bitmask (encode only writes `this.forceMove`). Callers using the legacy flags API silently lose teleport state for 1.26.40.
- **`SOUND_UNDEFINED` sentinel bumped globally** — [`LevelSoundEventPacket.java:961`](https://github.com/MemoriesOfTime/Nukkit-MOT/blob/feat/v1.26.40/src/main/java/cn/nukkit/network/protocol/LevelSoundEventPacket.java#L961): `SOUND_UNDEFINED` changed from 611 to 614, but the numeric-format encoder (`getProtocolSound`) sends it unchanged to pre-1.26.30 clients whose sound enum does not include 614. Remap the sentinel to 611 for older protocols.

### Recommendations
- Add malformed v2168 item-user-data coverage for oversized and negative `CanPlaceOn`/`CanDestroy` counts.
- Add regression coverage for `FLAG_TELEPORTING` and mixed-version `SOUND_UNDEFINED` encoding.
- Preserve compatibility constructors when extending public protocol data types.
- Ensure every reference-counted buffer in client-controlled decode paths is released through `finally`.

### Notes
- Static analysis only; no compilation or tests were run. `git diff --check` passed. The new item-state JSON has unique IDs/names; both added compressed NBT/block-state resources passed gzip integrity checks. Principal older-protocol serialization branches remain version-gated.

---
🤖 AI Review

PR #815, opened 2026-08-01
feat(plugin): implement full plugin unload and reload cleanup

lt-name:


lt-name:

lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Re-review after the latest fix commit `5b13d5261e`. The three previously reported blocking issues are now **RESOLVED**: (1) the server-startup NPE is fixed by a null-guard in `resolveOwner()` before dereferencing `getPluginManager()`, (2) separate command-handler ownership is resolved via `resolveSimpleCommandOwner()` with classloader lookup, and (3) displaced-mapping capture/restore is now done at the exact mutation points with `rebaseDisplacedCommands()` handling reverse-order unloads. However, deeper cleanup gaps remain — several paths still pin the plugin `ClassLoader` after unload, so the PR is **not ready to merge** yet.

### Issues Found
#### Blocking Issues

1. **Cross-plugin class-cache pollution: unloading one plugin can evict another plugin's classes, and surviving loaders keep references to the unloaded loader.**
   `PluginClassLoader.findClass(String, boolean)` (PluginClassLoader.java:82) returns from the local `classes` cache regardless of `checkGlobal`, and line 101 caches every globally-resolved class (including classes owned by *other* plugins) into the local cache. `JavaPluginLoader.unloadPlugin` (JavaPluginLoader.java:107) then calls `loader.findClass(entry.getKey(), false)` and removes any global entry whose class compares equal. Consequence:
   - If loader A ever resolved a class owned by plugin B, A's local cache holds B's `Class<?>`. Unloading A returns that cached class even with `checkGlobal=false`, so B's global cache entry is wrongly removed.
   - If surviving loader B resolved a class owned by A, B's local `classes` map keeps A's `Class<?>` after A is unloaded — pinning A's `PluginClassLoader` and all of A's classes forever.
   The new `JavaPluginLoaderUnloadTest` only loads one class per owner directly and never exercises either cross-plugin lookup direction.

2. **Metadata is invalidated but never removed, so stale values remain observable and keep plugin classes reachable.**
   `PluginManager.unloadPlugins` (PluginManager.java:678-680, 725) calls `invalidateAll(plugin)`, but `MetadataStore.invalidateAll` (MetadataStore.java:58-67) only invokes the value's `invalidate()` callback — it never removes the plugin's entry or empty outer keys. `getMetadata()` (lines 30-37) keeps returning the value, and a plugin-defined `MetadataValue` instance (and its defining class) stays strongly referenced regardless of invalidation. `removeMetadata()` already implements the correct remove-with-prune semantics; the unload path should use a bulk removal instead of `invalidateAll`.

3. **The retained public `SimpleCommand` constructor creates commands that unload cannot attribute, leaving the handler object and reflected `Method` pinned.**
   SimpleCommand.java:24-26 keeps the legacy constructor, which delegates with `owningPlugin = null`. The command still strongly stores `object` and `method` (lines 17-18). But `SimpleCommandMap.ownsCommand` (SimpleCommandMap.java:290-296) treats every `PluginIdentifiableCommand` as authoritative: for a null-owner command, `plugin == ((PluginIdentifiableCommand) command).getPlugin()` is always false, so `unregister(plugin)` never removes it — the handler, its reflected `Method`, and the old `ClassLoader` survive unload. Displacement snapshots taken at registration may also be discarded without restoring the displaced vanilla command.

4. **Service registrations can still pin a closed plugin classloader.**
   `NKServiceManager.cancel(Plugin)` (NKServiceManager.java:45-66) removes matching providers but never removes a service key from `handle` once its list becomes empty (line 12, 52-62). If the service interface class is defined by the plugin, the `Class<?>` key remains strongly referenced after unload. Additionally, `JavaPluginLoader.disablePlugin` (JavaPluginLoader.java:171-181) calls `cancel(plugin)` only when `plugin.isEnabled()`, and `register()` (NKServiceManager.java:15-26) does not require an enabled plugin — so a service registered during `onLoad()` on a plugin that is unloaded before being enabled leaves the full provider registration behind (PluginManager.java:648-654 only calls `disablePlugin` when enabled).

5. **Individual plugin unload does not remove runtime-registered permissions.**
   `PluginManager.disablePlugin` (PluginManager.java:517-521) removes only permissions declared in `plugin.yml`. The comment at lines 541-543 explicitly acknowledges that runtime `addPermission()` registrations are cleared only by a full reload (`clearRegistries()`), but `unloadPlugins` never calls it. A single-plugin unload/reload therefore leaves dynamically added permissions registered — the next instance may fail to re-register them, and a plugin-defined `Permission` subclass remains a direct classloader reference.

#### Non-blocking Issues
- `PermissibleBase.clearAttachments(Plugin)` (PermissibleBase.java:124-134) sets `changed = true` only after the removal callback returns; if `PermissionRemovedExecutor.attachmentRemoved` throws, recalculation is skipped and remaining attachments are not processed.
- `SimpleCommandMap` unload uses `new HashSet<>()` plus `owned.contains(entry.getValue())` (SimpleCommandMap.java:233, 244) — equality-based. A plugin command overriding `equals()` broadly could cause other commands' mappings to be removed; identity semantics would match the rest of the unload design.
- The fallback-prefix ownership resolution in `resolveOwnerByClassLoader`-adjacent logic (SimpleCommandMap.java:192) searches by lowercased prefix before checking the command's classloader; two plugin names differing only by case (or a custom fallback prefix colliding with another plugin) can misattribute displacement snapshots. Classloader ownership should take precedence for plain commands loaded by `PluginClassLoader`.

### Recommendations
- Determine class ownership by `entry.getValue().getClassLoader() == loader` (do not call `findClass()` during unload), and purge classes defined by the unloaded loader from every surviving `PluginClassLoader`'s local cache.
- Add a metadata-store bulk-removal operation that deletes the plugin's entries and prunes empty outer keys; use it instead of `invalidateAll()` during unload.
- Track command ownership by command identity at registration, or make null `PluginIdentifiableCommand` ownership fall through to recorded/classloader ownership so legacy-constructor commands are still attributable.
- Remove empty service-map keys in `NKServiceManager.cancel(Plugin)`, and cancel services unconditionally during unload (not only when the plugin is enabled).
- Track runtime permission registrations by owning plugin and drop them on individual unload, not just on full reload.
- Add regression tests for: cross-plugin class lookup in both directions, legacy `SimpleCommand` constructor attribution, metadata removal, runtime permission cleanup, and service-interface key cleanup.

### Notes
- Previous blocking issues 1 (startup NPE), 2(a) (separate handler ownership), and 2(b) (layered alias collisions) are confirmed **RESOLVED** in the current code; the layered alias case is covered by `SimpleCommandMapUnloadTest`.
- Review was read-only; no build or tests were run.

---
🤖 AI Review

PR #819, opened 2026-08-03
fix(level): stop dropping all block updates when a chunk exceeds the per-tick change cache

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. The PR fixes a real bug: the sentinel map (`changeBlocksFullMap`) is never populated, so the old `blocks.size() > MAX_BLOCK_CACHE` check could never detect an overflowing chunk, silently dropping all its block updates for the tick. The new `shouldResendWholeChunk()` helper detects the sentinel by reference identity and correctly restores the whole-chunk resend path.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- The overflow test verifies the predicate directly instead of exercising the actual tick-end flush/delivery path, so it would not catch a regression where `doTick` stops invoking `onChunkChanged` or fails to clear the cache.
- `collectedReferenceResendsWholeChunk` passes `null` directly rather than clearing a `SoftReference` stored in `changedBlocks`; it covers the null branch of the predicate but not the cache-retrieval path.
- No negative case with a distinct-but-equal empty map; a future implementation using `equals()` instead of reference identity would still pass the current suite.
- The exact `MAX_BLOCK_CACHE + 1` boundary for the size-based fallback is untested.

### Recommendations
- Add test cases asserting: a distinct empty map does not request a resend, a cleared `SoftReference` produces the resend decision, exactly `MAX_BLOCK_CACHE + 1` entries trigger the size fallback, and a sentinel entry traverses the full flush path (calls `Player.onChunkChanged` and is removed by cleanup).

### Notes
- Both the producer (`addBlockChange`) and the tick-end flush are synchronized on `changedBlocks`, so no new cache race is introduced.
- No other consumers of `changedBlocks` need changes; the producer already correctly avoids writing to the sentinel.
- Anti-xray and chunk serialization logic are unchanged; this only makes the existing full-resend branch reachable for the sentinel case.
- Runtime overhead is negligible: one identity comparison per changed chunk per tick.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

PR #823, opened 2026-08-07
fix(auth): stabilize offline player identity migration

### LINKED ISSUE #814: Already connected (1.26.30)
# 🐞 Bug Feedback | BUG反馈

Many players disconnect during the game with the reason 'Logged in from another location.' In the logs, this shows up as 'Already connected.' All of this only happens in version 1.26.30. Is this a game bug or a core bug?

The Already connected error appears with the same authentication address that the player was using.

Image

Image

lt-name:


lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The fix commits address several previously-flagged issues: async pre-login ordering is now correct (migration and lookup updates happen only after `PlayerAsyncPreLoginEvent` acceptance), lookup entries record auth provenance, the custom-serializer copy path is parse-validated, and offline identity derivation is case-insensitive to match lookup keys. However, three blocking issues remain — one around legacy 16-byte entries, one around an in-flight async save racing the migration, and one around undeletable corrupt targets silently resetting players.

### Issues Found
#### Blocking Issues
- **[严重] Legacy 16-byte name entries still expose Xbox-authenticated data.** `decodeNameEntry` maps 16-byte (pre-PR) records to `LEGACY_UNKNOWN` (`Server.java:2239`), but `processLogin` only blocks entries explicitly marked `XBOX_AUTHED` (`Player.java:3162`) — `LEGACY_UNKNOWN` falls through to the migration path. Existing Xbox-authenticated entries are precisely 16-byte records, and pre-PR player files carry no XUID marker (the marker is added only on new authenticated saves at `Player.java:3286-3290`), so `isHandoverAllowed` cannot protect them either. An offline login claiming such a name can still migrate the account's old data. The legacy name→UUID conversion is likewise unsafe: `updateName(uuid, name, false)` (`Server.java:2617-2619`) labels old name-keyed data as offline with no evidence of its provenance. Fix: never auto-migrate `LEGACY_UNKNOWN`; require explicit operator confirmation for provenance-less entries; mark legacy conversions as unknown, not offline.
- **[严重] An in-flight async save can run after the migration and recreate the old file.** `onCancel` sets `hasRun = true` *before* acquiring the per-identity lock (`Server.java:2528-2535`), and `flushPendingPlayerDataSaves` only calls `cancel()` (`Server.java:2431`), which returns immediately when `hasRun` is already true. The scheduler submits the task to the pool before cancelling its handler (`ServerScheduler.java:338`), so this interleaving is possible: a pool worker observes `hasRun == false`, sets it true, then pauses before acquiring the data lock; the migration flushes (cancel is a no-op), moves `previous.dat`; the worker resumes and writes `previous.dat` again — orphaning the latest save and leaving the migrated target stale. `volatile` marks "about to save", not "save completed". Fix: perform the save while holding the identity lock, or have the flush wait for actual completion (latch/future per save).
- **[严重] An undeletable corrupt target still becomes a silent player reset.** The new `PlayerDataSerializer.delete` defaults to `false` (`PlayerDataSerializer.java:44`), and `inspectTarget` treats `UNDELETABLE_RESIDUE` the same as valid `HAS_DATA` — both return `SKIPPED` (`PlayerDataMigrator.java:107`). `processLogin` aborts only on `FAILED`; on `SKIPPED` it proceeds to load the corrupt target, whose read error falls back to freshly created data (`Server.java:2354-2357`). Sequence: a custom serializer leaves a partial target and throws → retry finds it, cannot delete it, returns `SKIPPED` → the player is silently reset and the valid source is orphaned, never retried. Additionally, successful custom writes are not verified by rereading the destination (`PlayerDataMigrator.java:135-140`), `inspectTarget` can delete the target before the source is even opened, and the filesystem fallback checks only `previousFile.isFile()` rather than source parseability. Fix: return `FAILED` for `UNDELETABLE_RESIDUE`; verify the destination before returning `MIGRATED`; validate the source before deleting any target.

#### Non-blocking Issues
- `ClientChainData.normalizeOfflineIdentity()` (`ClientChainData.java:292-310`) still derives an offline UUID before name validation, creating a second identity derivation path that contradicts the validated final derivation.
- `playerDataLocks.computeIfAbsent` (`Server.java:2442-2444`) retains one lock per UUID/name forever; a long-running server accumulates unbounded locks.
- The 17-byte lookup format is forward-compatible (length discrimination) but not rollback-compatible: the `master` reader treats any non-16-byte value as invalid and deletes it.

### Recommendations
- Add integration tests for: 16-byte entry pointing at an unmarked Xbox save; legacy named Xbox data conversion; a worker setting `hasRun` before acquiring the data lock; a custom serializer whose `delete()` returns `false`; a serializer whose write closes but does not persist readable NBT; plugin rejection through `PlayerAsyncPreLoginEvent` asserting neither files nor lookup entries change.
- Use an explicitly versioned name-entry envelope if future schema or rollback compatibility matters.
- Remove the preliminary `normalizeOfflineIdentity()` derivation and keep only the validated final-name derivation.

### Notes
- The async pre-login ordering fix is verified correct (`Player.java:8575-8587`): rejection is handled before `processLogin()`, and the `loginDataProcessed` guard keeps the resource-pack completion path single-shot.
- The case-insensitive identity change is verified correct: derivation from the lowercased validated name matches the lowercased lookup key and the existing case-insensitive duplicate-name policy.

---
🤖 AI Review

PR #825, opened 2026-08-08
[Multiversion] some 1.1.0 protocol fix

yefengeeeeeeeeeee:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds 1.1.0 protocol support to several packets (Animate, EntityEvent, MobArmorEquipment, MobEffect, MobEquipment, MovePlayer, PlayerList), switching to unique-ID encoding and unsigned VarInt for pre-1.2.0 protocols. The direction is correct and most packets are wired consistently, but **MovePlayerPacket loses its legacy decode early-return**, which breaks the legacy wire format.

### Issues Found
#### Blocking Issues
- **MovePlayerPacket.java:32-46 — legacy `ridingEid` decode uses the wrong encoding.** The old legacy branch read `ridingEid = getEntityUniqueId()` (zigzag varLong) before returning. After removing the early-return, the shared path reads `ridingEid = this.getEntityRuntimeId()` (unsigned varLong). The legacy encode path still writes `putEntityUniqueId(this.ridingEid)` (line 73). Decode and encode are therefore asymmetric: for protocol < 1.2.0, a riding player's movement packet will decode `ridingEid` to a garbage value. This defeats the very 1.1.0 compatibility the PR is adding. Fix: gate `ridingEid` by protocol — use `getEntityUniqueId()` for protocol < v1_2_0, `getEntityRuntimeId()` otherwise.

- **MovePlayerPacket.java:47-56 — `MODE_TELEPORT` reads fields that the legacy encode never writes.** The legacy encode path returns immediately after `ridingEid` (line 74), but the shared decode path continues: `else if (this.mode == MODE_TELEPORT) { this.teleportCause = this.getLInt(); this.teleportItem = this.getLInt(); }`. For protocol 113, a teleport-mode packet has no such bytes, so the decoder reads 8 bytes past the actual payload → data misalignment / `ArrayIndexOutOfBoundsException` on short buffers, and batch decode aborts. Fix: keep the legacy early-return, or gate the teleport parsing (and the v1_16_100 frame field) with `protocol >= ProtocolInfo.v1_2_0`.

#### Non-blocking Issues
- **PlayerListPacket.java:66,103 — `Entry.uuid` can be null** (`Entry(UUID)` accepts null, line 131). The change now emits `putUUID(entry.uuid)` for protocol 113 too, so a plugin-created entry with a null UUID will NPE during encode. Pre-existing pattern for newer protocols, but the change widens the trigger surface. Suggest validating UUID at construction or null-guarding.
- **MobEquipmentPacket/MobArmorEquipmentPacket decode → `getSlot()`** can throw `NegativeArraySizeException` on negative CanPlaceOn/CanDestroy counts (BinaryStream.java:744). Pre-existing in the touched decode path; a bounded count guard would harden it.

### Recommendations
- Restore the legacy early-return in `MovePlayerPacket.decode()`, or split the pre-1.2.0 path fully (it is already fully split in `encode()` — mirror that structure in `decode()`).
- Add regression fixtures for protocol 113: `MovePlayerPacket` normal + teleport modes, `AnimatePacket` action 128/129, negative `EntityEventPacket` data, `PlayerListPacket` UUID presence at 113.
- Update the existing test at MiscPacketRegressionTest.java:330 — its comment says all pre-v223 protocols omit the UUID, which is no longer true for protocol 113.

### Notes
- No new `Server.mvw()` callers were introduced, so no missing `@Deprecated` annotations.
- The Animate/EntityEvent/MobArmorEquipment/MobEquipment legacy branches are encode/decode symmetric and look correct.

---
🤖 AI Review

yefengeeeeeeeeeee:
> ## 🤖 AI Review
> ### Summary
> This PR adds 1.1.0 protocol support to several packets (Animate, EntityEvent, MobArmorEquipment, MobEffect, MobEquipment, MovePlayer, PlayerList), switching to unique-ID encoding and unsigned VarInt for pre-1.2.0 protocols. The direction is correct and most packets are wired consistently, but **MovePlayerPacket loses its legacy decode early-return**, which breaks the legacy wire format.
> 
> ### Issues Found
> #### Blocking Issues
> * **MovePlayerPacket.java:32-46 — legacy `ridingEid` decode uses the wrong encoding.** The old legacy branch read `ridingEid = getEntityUniqueId()` (zigzag varLong) before returning. After removing the early-return, the shared path reads `ridingEid = this.getEntityRuntimeId()` (unsigned varLong). The legacy encode path still writes `putEntityUniqueId(this.ridingEid)` (line 73). Decode and encode are therefore asymmetric: for protocol < 1.2.0, a riding player's movement packet will decode `ridingEid` to a garbage value. This defeats the very 1.1.0 compatibility the PR is adding. Fix: gate `ridingEid` by protocol — use `getEntityUniqueId()` for protocol < v1_2_0, `getEntityRuntimeId()` otherwise.
> * **MovePlayerPacket.java:47-56 — `MODE_TELEPORT` reads fields that the legacy encode never writes.** The legacy encode path returns immediately after `ridingEid` (line 74), but the shared decode path continues: `else if (this.mode == MODE_TELEPORT) { this.teleportCause = this.getLInt(); this.teleportItem = this.getLInt(); }`. For protocol 113, a teleport-mode packet has no such bytes, so the decoder reads 8 bytes past the actual payload → data misalignment / `ArrayIndexOutOfBoundsException` on short buffers, and batch decode aborts. Fix: keep the legacy early-return, or gate the teleport parsing (and the v1_16_100 frame field) with `protocol >= ProtocolInfo.v1_2_0`.
> 
> #### Non-blocking Issues
> * **PlayerListPacket.java:66,103 — `Entry.uuid` can be null** (`Entry(UUID)` accepts null, line 131). The change now emits `putUUID(entry.uuid)` for protocol 113 too, so a plugin-created entry with a null UUID will NPE during encode. Pre-existing pattern for newer protocols, but the change widens the trigger surface. Suggest validating UUID at construction or null-guarding.
> * **MobEquipmentPacket/MobArmorEquipmentPacket decode → `getSlot()`** can throw `NegativeArraySizeException` on negative CanPlaceOn/CanDestroy counts (BinaryStream.java:744). Pre-existing in the touched decode path; a bounded count guard would harden it.
> 
> ### Recommendations
> * Restore the legacy early-return in `MovePlayerPacket.decode()`, or split the pre-1.2.0 path fully (it is already fully split in `encode()` — mirror that structure in `decode()`).
> * Add regression fixtures for protocol 113: `MovePlayerPacket` normal + teleport modes, `AnimatePacket` action 128/129, negative `EntityEventPacket` data, `PlayerListPacket` UUID presence at 113.
> * Update the existing test at MiscPacketRegressionTest.java:330 — its comment says all pre-v223 protocols omit the UUID, which is no longer true for protocol 113.
> 
> ### Notes
> * No new `Server.mvw()` callers were introduced, so no missing `@Deprecated` annotations.
> * The Animate/EntityEvent/MobArmorEquipment/MobEquipment legacy branches are encode/decode symmetric and look correct.
> 
> 🤖 AI Review 

https://github.com/Nukkit/Nukkit/blob/399ced0a29f306c7378725d10f5a347026fd09c1/src/main/java/cn/nukkit/network/protocol/MovePlayerPacket.java  
MovePlayerPacket also read teleportCause & teleportItem

src/main/java/cn/nukkit/network/protocol/AnimatePacket.java
```diff
@@ -28,13 +28,19 @@ public class AnimatePacket extends DataPacket {
...
+            this.action = Action.fromId((int) this.getUnsignedVarInt());
         } else {
             this.action = Action.fromId(this.getVarInt());
         }
         if (this.action == null) {
             this.action = Action.NO_ACTION;
         }
-        this.eid = getEntityRuntimeId();
+        if (protocol >= ProtocolInfo.v1_2_0) {
+            this.eid = getEntityRuntimeId();
```
lt-name: For versions prior to 1.2.0, the behavior of `getEntityRuntimeId` and `getEntityUniqueId` is identical; this is implemented within the `DataPacket` class, so this modification is unnecessary.

lt-name (changes requested):

src/main/java/cn/nukkit/network/protocol/AnimatePacket.java
yefengeeeeeeeeeee: I will revert the change here

tachibana-kanade-angel:

src/main/java/cn/nukkit/network/protocol/AnimatePacket.java
yefengeeeeeeeeeee: 哥我错了😭

PR #826, opened 2026-08-08
feat(config): add use-hashed-block-network-ids toggle

lt-name:


lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This re-review confirms the two previously flagged blocking issues are resolved: (1) runtime-ID consumers (`PunchBlockParticle`, `BlockSponge`, `TerrainParticle`) now encode per client protocol instead of assuming legacy runtime IDs or broadcasting latest-version IDs to all clients; (2) biome definition caches are rebuilt via `BiomeDefinitionListPacket.rebuildHashSensitiveCaches()` when `closeRegistry()` force-enables hashing. The new `custom-block-settings.use-hashed-block-network-ids` toggle (default `true`) is wired at `Server.init()` and localized in 4 languages.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- `TerrainParticle` no longer extends `GenericParticle` (dropped protected `id`/`data` fields), and `PunchBlockParticle` drops the protected `index` field. No in-tree consumers rely on these, but it is a source/binary compatibility change for plugins that subclass or cast these particles.
- The toggle is read only at startup (`Server.init()`); `/reload` swaps `serverConfig` but does not reapply the flag or rebuild caches. The config comment should state that a restart is required for changes to take effect.
- `GlobalBlockPalette.java:667` javadoc references `use-hashed-block-network-ids` without the full nested path `custom-block-settings.use-hashed-block-network-ids`.

### Recommendations
- Add a release/migration note: the default `true` changes existing 1.19.80+ servers to hashed IDs; setting `false` requires a restart.
- Document the particle API change (`TerrainParticle`/`PunchBlockParticle` protected members) in the changelog, or keep the old members as deprecated shims if plugin compatibility matters.
- Consider CI coverage for both sides of the 1.19.80 protocol cutoff and the custom-block force-enable path.

### Notes
- LevelEventPacket `EVENT_PARTICLE_BREAK_BLOCK_*` constants (3603–3608) already exist on master; this PR only uses them.
- Biome cache rebuild order is correct: `closeRegistry()` sets hashing first, then `compactCaches()`, then `rebuildHashSensitiveCaches()`.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #828, opened 2026-08-09
feat(block): block tags

Kanelucky:
debug-log
This is result from debug log.
btw i think it would be better to create a separate directory for autogenerated file, for example src/autogenerated


Kanelucky:
this is code generator
https://github.com/Kanelucky/LumiCodeGen/blob/master/src/main/java/com/luminiadev/lumi/codegen/generator/BlockTagsGen.java

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR introduces a block tag system: a `BlockTag` interface, a generated `BlockTags` registry, `LazyBlockTag`/`SimpleBlockTag` implementations, a vanilla definition file `tags/block_tags.json`, and integration points in `Block` (`getBlockTags()`/`hasBlockTag()`) plus a demo log block in `Server.start()`. The overall design (registry + lazy-loaded JSON definitions + per-block reverse index) is reasonable, but the tag resolution logic has core correctness problems that leave several built-in tags empty, and the ID-only model cannot represent metadata-distinguished variants.

### Issues Found
#### Blocking Issues
- **BlockID → tag name resolution is broken; 8 built-in tags are guaranteed empty.** `LazyBlockTag.createBlockIdMap()` builds `minecraft:<FIELD_NAME_LOWERCASED>` keys from `BlockID` constants (e.g. `LOG` → `minecraft:log`, `MONSTER_SPAWNER` → `minecraft:monster_spawner`), but `tags/block_tags.json` uses canonical Minecraft names (e.g. `minecraft:oak_log`, `minecraft:mob_spawner`). Verified: `minecraft:acacia/birch/dark_oak/jungle/log/mob_spawner/oak/spruce` all reference names that can never match a generated key, so those tags resolve to empty sets. The lookup must map through the server's canonical block-name/state mapping instead of Java constant names.
- **ID-only tag model cannot represent metadata-distinguished variants.** `BlockTag` stores only integer IDs and `Block#getBlockTags()` uses `this.getId()`, discarding damage/state. Acacia and dark oak are both `WOOD2` (162), distinguished only by metadata. Even after fixing the name resolution, `minecraft:acacia` and `minecraft:dark_oak` would both match ID 162 and incorrectly tag every WOOD2 variant. The tag model needs to account for block damage/state (or these variants must be excluded).

#### Non-blocking Issues
- `BlockTag#getBlockId()` and `BlockTags#getTagsSet()` expose the actual mutable collections, so external mutation can desynchronize the forward/reverse indexes (e.g. clearing the set returned by `getTagsSet()` removes registered tags for that ID). Return unmodifiable views/copies.
- `BlockTags.register()` publishes the name before resolving membership; if `getBlockId()` throws (unknown vanilla tag), the name stays registered with no reverse mappings and retry fails the duplicate check. Resolve first, then publish both indexes.
- Registry maps are unsynchronized `HashMap`/`Int2ObjectOpenHashMap`; concurrent plugin registrations could lose entries. (Low practical risk since plugin loading is sequential, but worth guarding.)
- `LazyBlockTag`'s lazy field is not `volatile`/synchronized; safe in practice because built-ins load during class initialization, but independently constructed instances are not safely published.
- Despite the name, built-in tags are eagerly loaded during `BlockTags` class init (registration calls `getBlockId()`), and Lombok-generated `equals/hashCode` includes the whole `IntSet`, making `HashSet.contains()` in `Block#hasBlockTag()` hash hundreds of IDs per call.
- `Server.start()` (lines 1482-1491) leaves demo debug logging that creates sample blocks and prints tag info on every server startup — remove or move to tests.

### Recommendations
- Resolve JSON names through the server's canonical block name/ID mapping rather than reflecting over `BlockID` field names.
- Decide the metadata question explicitly: either extend the model to (id, damage) pairs or document that metadata-distinguished tags are unsupported.
- Return unmodifiable collections from all public getters; make registration atomic and thread-safe.
- Remove the debug logging from `Server.start()`.

### Notes
- `block_tags.json` is valid JSON; its 39 keys match the names registered in `BlockTags`; no duplicate members.
- No pre-existing public signatures changed, so no new `@Deprecated` annotations are required.

---
🤖 AI Review

PR #829, opened 2026-08-09
feat(network): support optional IPv6 listener alongside IPv4

lt-name:


lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Re-review after commit `4cd1901f` ("fix(network): abort startup on bind failure and clean up RakNet resources"). The previously reported blocking issue — IPv6 bind failures silently treated as success — is **fixed**: both bind futures now check `isSuccess()` and propagate the Netty `cause()` to abort startup, partially created channels are closed, and the event-loop group is shut down on failure. However, two blocking issues remain around input validation and advertised endpoints.

### Issues Found
#### Blocking Issues
- **`server-ipv6` is not validated as an IPv6 address, and the listener is not guaranteed to be IPv6-only** — `Server.java:3617` keeps the value as a raw string and `RakNetInterface.java:129` binds it through the same family-unspecified channel factory. Consequences: `server-ipv6=127.0.0.1` silently starts a second IPv4 listener; a hostname with both A/AAAA records can resolve to IPv4; on dual-stack systems binding `::` can accept IPv4-mapped traffic because `IPV6_V6ONLY` is never enforced, making the `sendRawPacket()` family-based routing at `RakNetInterface.java:379` unsafe.
- **`server-ipv6-port=0` is accepted but advertised as port 0** — the enablement condition at `Server.java:3620` accepts every non-negative value, so Netty binds an ephemeral port while startup logging and the RakNet advertisement (`RakNetInterface.java:393`) keep publishing the configured value `0`. Clients cannot reach the advertised endpoint. Either reject `0` (require `1..65535`) or read the actual bound port from `ipv6Channel.localAddress()`.

#### Non-blocking Issues
- IPv6 listener is disabled whenever its port equals `server-port`, even for distinct addresses where both binds would be valid — blocks the common `0.0.0.0:19132` + `[::]:19132` setup.
- `/reload` port/address changes update the mutable server fields but do not recreate listeners; subsequent advertisement updates can publish a disabled or `-1` IPv6 port while the old channel stays bound.
- `shutdownGracefully()` is not awaited; event-loop creation and bootstrap construction sit outside the cleanup `try`, leaving a narrow setup-exception leak path (`RakNetInterface.java:73`).
- Removing JVM-wide `java.net.preferIPv4Stack=true` (`Nukkit.java:63`) changes DNS/socket behavior for RCON, metrics, plugins and other JVM networking, not just RakNet. No in-tree regression found, but the compatibility impact should be documented.
- IPv6 administration is incomplete: `/ban-ip` validation and `/pardon-ip` accept only IPv4 literals (`IPStringNode.java:12`, `PardonIpCommand.java:33`).

### Recommendations
- Validate `server-ipv6` resolves to an `Inet6Address` and enforce IPv6-only binding (family-specific channel / `IPV6_V6ONLY`).
- Treat `-1` as disabled and require `1..65535`, or advertise the actual bound port when ephemeral is intentional.
- Preserve actual bound endpoints for logging, advertisement, and outbound query routing.
- Cover all construction in one cleanup scope and optionally await event-loop termination.
- Add mocked coverage for failed IPv4 bind, failed IPv6 bind after IPv4 success, invalid/zero ports, and wrong-family resolution.

### Notes
- Defaults (`server-ipv6-port=-1`) preserve existing IPv4-only configurations.
- Review was static only; no compilation or build was performed.

---
🤖 AI Review

PR #830, opened 2026-08-09
fix(block): correct target block hit vector offset

SemihAkin685:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR adjusts the hit-vector offset used when computing the Target block's redstone power from projectile impacts. It is a one-line edit in `BlockTarget.onEntityCollide`, but it reverses the coordinate compensation that the current `Block.calculateIntercept` implementation requires, which breaks the Target block anywhere away from the world origin.

### Issues Found
#### Blocking Issues
- **`BlockTarget.java:177` — new offset yields world coordinates instead of block-local coordinates, breaking power calculation.** `Block.calculateIntercept` (Block.java:1206-1286) computes intersection points via `getIntermediateWithXValue/YValue/ZValue` against the block's own bounding box (`getMinX()`/`getMinY()`/`getMinZ()` return `x`/`y`/`z`, i.e. world coordinates), then returns `vector.add(this.x, this.y, this.z)` — so `intercept.hitVector` is effectively `worldHit + blockOrigin`. The previous `subtract(x*2, y*2, z*2)` compensated exactly: `worldHit + origin - 2*origin = worldHit - origin`, giving block-local coordinates in [0,1]. The new `subtract(x, y, z)` leaves `worldHit`, so the values fed into the `coords[]` mapping fall outside [0,1] for any block not at the world origin and hit the `coords[i] <= 0 || coords[i] >= 1 → 0` branch — the Target block outputs 0 power and effectively never activates. Verified by tracing the intercept math and simulating hits at (100, 64, 100) and (-5, -3, -2): the old code yields local coordinates in [0,1]; the new code yields raw world coordinates.

#### Non-blocking Issues
- None introduced by this PR.

### Recommendations
- The proper fix is to correct `Block.calculateIntercept` itself: the intersection points are already in world space, so `vector.add(this.x, this.y, this.z)` (Block.java:1285) double-adds the block origin. Removing that addition would make `subtract(x, y, z)` correct. Alternatively, keep the existing `subtract(x*2, y*2, z*2)` compensation as-is.
- Pre-existing (not introduced by this PR): the ray endpoint at lines 166-167 computes `position.add(motion.multiply(2))` where `motion = position + lastMotion`, producing `3*position + 2*lastMotion` — the likely intent was `position + 2*lastMotion`. Worth a follow-up fix.

### Notes
- `BlockTarget` uses a full-block bounding box, so the local-coordinate [0,1] assumption in the `coords[]` mapping is valid.

---
🤖 AI Review

PR #831, opened 2026-08-10
fix(leveldb): accept block states written by PocketMine-MP

w1zardz:
### Problem

A world saved by PocketMine-MP loads on Nukkit-MOT as solid `minecraft:info_update` — every single block, including `minecraft:air`.

PocketMine-MP writes a private `PMMPDataVersion` key into every block state it stores in the LevelDB sub chunk palette. Vanilla palettes do not have that key, and `StateBlockStorage` looks states up as whole NBT maps, so not a single entry matches.

The failure is quiet: `Chunk contains unknown block` is only printed at `debug-level: 2`, so on the default level the load looks perfectly clean and the symptom reads like a broken resource pack.

### Fix

Drop the foreign key right after the palette entry is read, so everything downstream keeps working with plain vanilla states.

### Testing

Opened a world written by PocketMine-MP 5.44.2 on a Nukkit-MOT server:

- before: every block in the world is `info_update`;
- after: the world loads exactly as it was built, no unknown-block lines at `debug-level: 2`.

### DIFF
--- a/src/main/java/cn/nukkit/level/format/leveldb/structure/StateBlockStorage.java
+++ b/src/main/java/cn/nukkit/level/format/leveldb/structure/StateBlockStorage.java
@@ -23,6 +23,7 @@ import lombok.extern.log4j.Log4j2;
 import org.cloudburstmc.nbt.NBTInputStream;
 import org.cloudburstmc.nbt.NBTOutputStream;
 import org.cloudburstmc.nbt.NbtMap;
+import org.cloudburstmc.nbt.NbtMapBuilder;
 import org.cloudburstmc.nbt.NbtUtils;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR correctly fixes the PMMP palette mismatch. `state.toBuilder()` creates an independent builder, `remove()` strips the private `PMMPDataVersion` key, and `build()` produces a new immutable `NbtMap` that is used for all downstream lookups (`hashCode()` caching, `getStateUnsafe()`, state updater, custom-state caching). Palette entries are appended one-for-one, preserving serialized palette indices. All subchunk read paths (V1/V8/V9 serializers) funnel through this method, so the fix covers every block-state load. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- Each PMMP palette entry allocates a builder and copies the map (proportional to palette size, not block positions) — acceptable for a compatibility fix, but slightly heavier on PMMP worlds.
- No automated regression test covering the `PMMPDataVersion` strip path.
- `LevelDBProvider.parseBlockTickingQueue()` performs a separate whole-map block-state lookup; if PMMP also writes `PMMPDataVersion` into ticking-queue `blockState` compounds, that path would need the same normalization (not demonstrated by this PR).

### Recommendations
- Add a regression test: a known state with `PMMPDataVersion`, a known state without the key, two palette entries differing only by the key, and a load→save round-trip verifying the private key is not emitted.
- If other LevelDB block-state readers are confirmed to need the same handling, extract the key-stripping into a small helper.

### Notes
- `NbtMap` is immutable, so no object already used as a hash-map key is mutated; the rebuilt map starts fresh and the explicit `state.hashCode()` caches the sanitized hash.
- Entries without the key retain existing behavior; malformed non-map tags still fail at the pre-existing cast and are caught by the surrounding handler.
- No symmetric `writeToStorage()` change needed: known states resolve to canonical mapping snapshots, so PMMP metadata is not preserved on re-save (as expected).

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

lt-name (approved):

tachibana-kanade-angel:

PR #832, opened 2026-08-12
fix(trident): prevent unintended pickup of returning Loyalty trident

Suoerix:
This PR fixes an issue where a Loyalty trident returning from the ground could be intercepted/picked up by other players along its return path, and resolves a potential `NullPointerException`.

### Cause of Bug
1. **Trident Interception**:
   - When a Loyalty trident hits a player and returns, `hadCollision` is reset to `false`, enforcing the guard check that restricts pickup solely to the thrower.
   - However, when hitting the **ground** (`onHitGround`), `hadCollision` remains `true` throughout the return journey.
   - In `Player.pickupEntity`, the owner validation guard was skipped if `hadCollision == true`. As a result, any player within a 1-block radius during its return flight could intercept and pick up the trident.

2. **Potential NPE**:
   - `shootingEntity.equals(this)` raises a `NullPointerException` if `shootingEntity` is `null` (e.g., launched by dispensers or lost after deserialization).

### Solution
- Refactored `Player.pickupEntity` guard logic: Whenever a thrown trident is in returning state (`isNoClip()`), restrict pickup strictly to its original thrower regardless of `hadCollision`.
- Added a null check for `trident.shootingEntity` before equality evaluation to eliminate the NPE risk.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -23,7 +23,6 @@ import cn.nukkit.entity.mob.EntityWolf;
 import cn.nukkit.entity.passive.EntityHappyGhast;
 import cn.nukkit.entity.passive.EntityVillager;
 import cn.nukkit.entity.projectile.EntityArrow;
-import cn.nukkit.entity.projectile.EntityProjectile;
 import cn.nukkit.entity.projectile.EntityThrownTrident;
 import cn.nukkit.event.block.WaterFrostEvent;
 import cn.nukkit.event.entity.*;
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The fix correctly prevents interception of a returning Loyalty trident by other players. The new guard makes `isNoClip()` authoritative for return-ownership validation: whenever the trident is in its returning state (either after a block hit or an entity hit), only the original thrower (`shootingEntity`) may pick it up, regardless of `hadCollision`. This closes the original hole where a block-hit trident (`hadCollision == true`) skipped thrower validation entirely.

The added `shootingEntity == null` check also eliminates the potential `NullPointerException` in `shootingEntity.equals(this)` for tridents launched by dispensers or whose owner was lost after deserialization.

Behavior verified across all guard states:
| hadCollision | isNoClip() | shootingEntity | Result |
|---|---|---|---|
| false | false | Any | Reject (in flight) |
| true | false | Any | Allow (resting/collided) |
| false | true | null | Reject (was NPE before) |
| false | true | this player | Allow |
| false | true | another entity | Reject |
| true | true | null | Reject (was NPE before) |
| true | true | this player | Allow |
| true | true | another entity | Reject |

Resting-trident behavior (`hadCollision=true, isNoClip=false`, e.g. no Loyalty enchantment) is unchanged — still generally collectible subject to pickup mode.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- No dedicated regression tests were added for the pickup-guard scenarios.
- `shootingEntity` is read twice in the condition; Nukkit entity updates run on the main server thread so there is no practical race, but a local snapshot would make the guard mechanically race-proof.
- `setTridentRope(true)` at EntityThrownTrident.java:419 still dereferences `shootingEntity`; current internal call sites are guarded by `canReturnToShooter()` (which requires a non-null player shooter), so this is not a blocker for this PR.

### Recommendations
Consider adding tests covering: block-hit returning trident crossing another player, entity-hit returning trident crossing another player, successful pickup by the original shooter in both states, null-shooter `noClip` trident returning `false` without throwing, and a resting trident remaining collectible.

### Notes
- The local `thrownTrident` variable is a good readability cleanup; pattern matching (`instanceof EntityThrownTrident thrownTrident`) would be slightly more idiomatic on Java 17.
- Removing the now-unused `EntityProjectile` import is safe.
- No public API signatures or event behavior changed.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

lt-name (approved):

tachibana-kanade-angel:

PR #834, opened 2026-08-13
   fix: weaker effects no longer override stronger effects of the sam…

CoolMagic233:
…e type

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -1088,9 +1088,14 @@ public abstract class Entity extends Location implements Metadatable {
             return;
         }
 
-        if (cause != null) {
-            Effect oldEffect = this.effects.get(effect.getId());
+        Effect oldEffect = this.effects.get(effect.getId());
 
+        if (oldEffect != null && (oldEffect.getAmplifier() > effect.getAmplifier()
+            || (oldEffect.getAmplifier() == effect.getAmplifier() && oldEffect.getDuration() >= effect.getDuration()))) {
+            return;
+        }
+
+        if (cause != null) {
             EntityPotionEffectEvent event = new EntityPotionEffectEvent(
                     this,
                     oldEffect,


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The PR adds an early-return guard in `Entity.addEffect(Effect, Cause)`: a new effect is rejected when the entity already has a stronger effect of the same type (higher amplifier, or equal amplifier with greater-or-equal remaining duration). This matches Minecraft vanilla (1.20+) effect-merge semantics — weaker effects no longer override stronger ones, and equal-amplitude effects keep the longer duration.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- Rejected applications no longer fire `EntityPotionEffectEvent` / `EntityEffectUpdateEvent` nor send a `MobEffectPacket` (`Entity.java:1093`). Semantically reasonable since no effect change actually occurs, but plugins that used the old event as an "attempted application" hook may observe a compatibility change.
- `/effect` command (`EffectCommand.java:112`) always logs success even when the guard rejects a weaker/equal effect — a pre-existing command/API limitation made more visible by this PR.

### Recommendations
- Add focused tests covering the amplifier/duration combinations (including equality), asserting that rejected applications do not fire either effect event and do not send a packet.

### Notes
- All periodic refresh paths remain correct: beacon (220–340 ticks, refresh every 80 ticks), conduit (260 ticks, refresh every 40 ticks), area-effect clouds, arrows and mob attacks. Effect objects stored in `this.effects` have their duration decremented every tick (`Entity.java:2121`), so the remaining-duration comparison is correct.
- Instant effects (healing/harming) bypass `addEffect` in potion/arrow/cloud paths and are unaffected.
- All built-in `addEffect` call sites set a positive duration; none relies on the constructor's default duration of 0.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

lt-name (approved):

tachibana-kanade-angel:

PR #838, opened 2026-08-16
fix(player): sync movement speed attribute on setMovementSpeed

### LINKED ISSUE #836: Player's method setMovementSpeed() doesn't work
# 🐞 Bug Feedback | BUG反馈

## 📝 Bug Description | BUG描述

Player下的setMovementSpeed()方法仅修改了玩家对象下的属性，没有真正产生效果
我尝试使用了速度和缓慢药水效果，他们是可以正常工作的

## 🔍 Steps to Reproduce | 复现方法

```
//首先执行
player.setMovementSpeed(0.0001f);

//然后获取移速
player.getMovementSpeed();//这个值会返回0.0001f，但实际移动速度没有改变

```

## 💻 OS and versions | 操作系统和核心版本
Build: master/b9cffd1
v1.21.114、v1.26.3005均可复现

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes issue #836 by separating the player's base movement speed from the client-bound speed: `setMovementSpeed` now updates the base speed, recomputes `speedToSend` (modifiers stack on top of the base value), and sends the updated attribute to the client. The direction is correct and the added regression tests are meaningful. However, there is one blocking issue where an update can be silently dropped during the first spawn sequence, leaving the client's movement speed permanently out of sync.

### Issues Found

#### Blocking Issues
- **First-spawn `setMovementSpeed` update can be permanently dropped** (`Player.java:7046`, `doFirstSpawn` at `Player.java:1347` / `spawned = true` at `Player.java:1388`): The `if (this.spawned && send)` guard drops the attribute update when `spawned == false`. During login, `sendAttributes()` runs early in `completeLoginSequence` (`Player.java:3492`), *before* `doFirstSpawn()` fires `PlayerRespawnEvent` (`Player.java:1347`) with `spawned` still `false`. If a plugin calls `setMovementSpeed()` from a first-spawn `PlayerRespawnEvent` listener, `speedToSend` is updated but no packet is sent, and nothing re-syncs the attribute after `spawned` becomes `true` at line 1388 — the client keeps the old speed value indefinitely. This is exactly the out-of-sync scenario issue #836 describes, just on the first-spawn path.

#### Non-blocking Issues
- **`sendMovementSpeed()` leaves the attribute's default value at `0.10f`** (`Player.java:7062`): `Attribute.getAttribute(Attribute.MOVEMENT_SPEED).setValue(speedToSend)` does not call `setDefaultValue`, so the serialized default stays the static `0.10f`, while `sendAttributes()` (`Player.java:2714`) sets `setDefaultValue(this.getMovementSpeed())`. If the base speed differs from 0.1, the client receives inconsistent default values across the two paths.
- **Validation accepts `NaN` and positive infinity** (`Player.java:7040`): `if (speed < 0)` is `false` for both `NaN` and `+Infinity`. `NaN` is serialized as-is to the client; `+Infinity` is clamped in individual attribute updates but remains infinity server-side, so a later `sendAttributes()` → `setDefaultValue(getMovementSpeed())` throws `IllegalArgumentException`.

### Recommendations
- Track a pending movement-speed update and flush it after first spawn (e.g., re-send the attribute when `spawned` transitions to `true`), or send the attribute unconditionally in `setMovementSpeed(speed, send)` and let the caller decide.
- Build the movement attribute in a single helper that consistently sets both value and default (used by `sendAttributes()` and `sendMovementSpeed()`).
- Reject non-finite values: `if (speed < 0 || Float.isNaN(speed) || Float.isInfinite(speed))`.
- Add a regression test for `setMovementSpeed()` invoked from a first-spawn `PlayerRespawnEvent`, and a direct packet-capture test for the spawned `setMovementSpeed()` path.

### Notes
- Modifier add/remove (`Player.java:9023` area) recalculates `speedToSend` twice because `super.addMovementSpeedModifier()` calls the overridden `recalculateMovementSpeed()`; harmless but redundant.
- The `respawn()` change at `Player.java:6909` (`recalculateMovementSpeed()` → `setMovementSpeed(DEFAULT_SPEED)`) runs with the client already ready, so sending is safe there.

---
🤖 AI Review

lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Re-review after the fix commit (`f7903aa88`): the previously blocking first-spawn desynchronization is resolved. `setMovementSpeed()` now updates the base speed and recomputes `speedToSend` regardless of spawn state, and `doFirstSpawn()` calls `sendMovementSpeed()` right after `spawned = true` (Player.java:1390), so a speed change made from a first-spawn `PlayerRespawnEvent` listener is flushed to the client. The regression tests for issue #836 are meaningful and the modifier stacking logic (`recalculateMovementSpeedToSend` now starts from `getMovementSpeed()`) is correct. No blocking issues remain.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- **`sendMovementSpeed()` leaves the attribute default at the factory value `0.10f`** (Player.java:7062): it sets the value but not the default, while `sendAttributes()` (Player.java:2716) sets `setDefaultValue(getMovementSpeed())`. If the base speed differs from 0.1, the client sees inconsistent default metadata across the two paths.
- **Validation accepts `NaN` and positive infinity** (Player.java:7039): `if (speed < 0)` is false for both. `NaN` is serialized as-is; `+Infinity` can later make `sendAttributes()` throw when setting the default value. Suggest `if (speed < 0 || !Float.isFinite(speed))`.
- **New test mocks `sendMovementSpeed()`** (PlayerMovementSpeedTest.java): the tests verify which paths call the method but don't capture the actual `spawned=true` flush introduced by the fix — the first-spawn re-sync at Player.java:1390 is not directly asserted. A lifecycle test invoking `setMovementSpeed()` from a first-spawn `PlayerRespawnEvent` would cover the exact regression.
- **`PlayerLegacyLoginInventoryTest` now transitively depends on `Attribute.init()`**: with the new `sendMovementSpeed()` call in `doFirstSpawn()`, that standalone test (which doesn't initialize attributes in its setup) is no longer reliably isolated. Consider adding `Attribute.init()` to its setup.

### Recommendations
- Set `.setDefaultValue(getMovementSpeed())` in `sendMovementSpeed()` for consistency with `sendAttributes()`.
- Reject non-finite values in `setMovementSpeed` and add NaN/infinity tests.
- Add a regression test that changes speed during first-spawn `PlayerRespawnEvent` and captures the post-`spawned=true` attribute packet.
- Initialize `Attribute` in `PlayerLegacyLoginInventoryTest` setup.

### Notes
- The `respawn()` change (`setMovementSpeed(DEFAULT_SPEED)` instead of `recalculateMovementSpeed()`) runs with the client already ready, so sending is safe there.
- Modifier add/remove paths still recalculate `speedToSend` via `recalculateMovementSpeed()` (no duplicated send) — behavior is consistent.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

PR #842, opened 2026-08-22
[Multiversion] Support for 1.26.45

lt-name:


lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. The previously blocking protocol-mapping issue (the `V1_16_200_51(420)` out-of-order entry being shadowed) is fixed by reordering the enum declarations into monotonic order, and the final `GameVersion` source compiles cleanly.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
1. **Public constant source compatibility** — `ProtocolInfo.java:114` removes `ProtocolInfo.v1_26_44`. External plugin *source* referencing it will no longer compile. Consider retaining a deprecated alias: `@Deprecated int v1_26_44 = v1_26_40;`. Existing compiled plugins normally inline interface constants, so this is primarily a source-compatibility concern.
2. **Missing regression coverage for the previously blocking case** — `GameVersionTest.java:17` covers the new duplicate-protocol behavior but not the fixed `420/421` mapping. Add assertions for `V1_16_200_51` so a future enum reordering cannot silently reintroduce the bug.

### Recommendations
- Add regression assertions for `byProtocol(420/421, false)` → `V1_16_200_51` alongside the existing 1.26.x invariants.
- Optionally retain a deprecated `v1_26_44` alias in `ProtocolInfo` for source compatibility.

### Notes
- Login detection correctly distinguishes 1.26.40 and 1.26.44 (both wire protocol `2168`) via the JWT version claim; 1.26.45 uses the real `2169` protocol and does not receive the 1.26.44-only `SetScore` field.
- `Player` propagates the resolved `GameVersion` into session state; palette/item mappings fall back to the existing `2168` resources.
- Falsifying the skin JWT version claim only selects a wire-format compatibility variant — no privilege escalation or protocol-bypass path.
- No sensitive-data exposure, injection, or resource-exhaustion issues found.

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #844, opened 2026-08-22
feat(level): parallel world ticking via per-level GameLoop, with thread-safety hardening

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This re-review covers the updated head (`ed08ed0d59`, "level lifecycle lock, pre-spawn packet queue, shared-state sync", 900+/320- across 34 files) and verifies each of the 12 previously-flagged blocking issues against the new diff.

**Conclusion: needs changes.** The fix commit resolves several narrow defects but does not yet establish the ownership invariant required for safe parallel ticking. Approval is not safe.

### Previously flagged issues — verification
- ✅ **Resolved (#2/#3/#10)**: processor-registry concurrency (DataPacketManager synchronized + lock-free resolution cache), logical tick advance on skipped iterations (GameLoop `tick++` unconditional), and the read-lock→write-lock unload self-deadlock (`Level.close()` detects read-lock hold and defers).
- ✅ **Resolved (#4 original)**: `Player.onUpdate()` returns `true` while `!loggedIn`, so the permanent-removal race is gone.
- ⚠️ **Partially resolved (#1/#5/#6/#8/#9/#11/#12)**: whitelist completion, chunk-request iterate-then-delete, lifecycle fallback on primary, stop/restart queue-drain, notification flags, async provider teardown ordering, GC interruption.
- ❌ **Still present (#7)**: cross-world transfer (`Entity.switchLevel`) still mutates source and target worlds directly without a transfer state machine.

### Blocking issues (verified)
1. **[High] Player disconnect tears down owner state from the primary/network thread** — `Player.close()` (`Player.java:6517-6574`) nulls `inventory`, clears windows, calls `super.close()` and removes the player while the level thread may still be inside `Player.onUpdate()` (`Level.java:1660-1668`). Concurrent dereference of freshly-torn-down state.
2. **[High] Login publication still precedes initialization completion** — `loggedIn=true` at `Player.java:3525`; `spawned=true` at `Player.java:1502` before the rest of `doFirstSpawn()` (`1508-1547`) runs. The level thread can enter the full tick once it observes `loggedIn`, concurrently with the still-running login sequence.
3. **[High] Cross-world transfer mutates two independently owned worlds directly** — `Entity.switchLevel()` (`Entity.java:2719-2748`) removes from source, inserts into target, and runs `afterSwitchLevel()` synchronously; target insertion/init is not queued to the target level thread.
4. **[High] Packet ownership whitelist is still incomplete** — `DataPacket.isLevelSyncPacket()` defaults `false` (`DataPacket.java:41-43`). Confirmed unmarked mutators running on the primary thread while the player ticks on its level thread: `ServerboundDataStorePacket` (client-writable property listeners), `ServerboundDataDrivenScreenClosedPacket` (`removeViewer`, `ServerboundDataDrivenScreenClosedProcessor.java:24`), `SetDifficultyPacket` (global difficulty + broadcast to all worlds, `SetDifficultyProcessor.java:31-35`).
5. **[Medium] `/difficulty` not forced to the primary thread** — `PRIMARY_THREAD_COMMANDS` (`Server.java:1442-1443`) omits `difficulty`; `DifficultyCommand` mutates global difficulty and broadcasts across all worlds from a level thread.
6. **[Medium] Chunk callback iterate-then-delete race** — `chunkRequestCallback()` (`Level.java:4568-4584`) removes the whole queue key after iterating; a concurrent pre-spawn `requestChunk()` can be erased, leaving chunk state stuck.
7. **[Medium] Lifecycle fallback on primary while old owner alive** — `stopLevelThread()` sets `acceptingWork=false` before join; an async callback can fall back to `runOrScheduleOnPrimary()` and mutate provider state (`generateChunkCallback`) while the old level thread is still running.
8. **[Medium] Stop/restart drains the new generation's queues** — `stopLevelThread()` releases `lifecycleLock` before join and drains afterward; a concurrent `startLevelThread()` of the new generation can have its queues cleared by the older stop call.
9. **[Medium] Notification flags check-then-clear loss** — e.g. `needSendFoodLevel` (`Player.java:9108` vs `9137`): a write between owner-read and owner-clear is lost.
10. **[Medium] `/reload` runs queued owner tasks on ForkJoin threads** — `CompletableFuture.runAsync(level::stopLevelThread)` (`Server.java:1522`); if already stopped, `stopLevelThread()` drains `syncTaskQueue` on the ForkJoin worker, executing owner tasks outside their ownership thread.
11. **[Medium] Async chunk-read shutdown doesn't enforce provider/read teardown ordering** — a worker past the `pending.invalidated` check can `readChunkOffThread()` while `Level.close()` tears down the provider (`Level.java:1106-1144`); custom providers may race teardown.
12. **[Medium] GC interruption mutates owner inline** — `GarbageCollectorCommand.java:60` runs `collection.run()` (chunk/entity GC + unload) on the primary thread if interrupted while waiting for the owner task.

### Non-blocking issues
- `SaveCommand` broadcasts success before queued async saves complete.
- Opt-in packet model means future/plugin processors silently default to primary-thread handling.
- `GameLoop` metrics arrays read by status code without a snapshot (diagnostics only).
- New tests cover cache invalidation/ID/packet-decode, not disconnect/login-publication/transfer/lifecycle-restart/queue-drain races.

### Recommendations
- Establish a single ownership invariant and enforce it for: disconnect teardown, login publication (`spawned` should flip only after `doFirstSpawn()` fully completes), cross-world transfer (queue to target level thread + barrier), and the packet whitelist (make defaults safe or route by attribute, and mark `ServerboundDataStore`/`DataDrivenScreenClosed`/`SetDifficulty`).
- Robustness: replace check-then-clear notification flags with atomic consume; hold `lifecycleLock` across join+drain so stop/restart is one atomic lifecycle; route `/difficulty` to the primary thread; make async chunk-read teardown ordering explicit in the provider contract.

### Notes
The pre-spawn packet queue itself is correctly structured (defer under `preSpawnPacketLock`, replay from a primary scheduler task). The lifecycle lock fixes the specific self-deadlock. Remaining issues are the publication/ownership windows, not the queue/lock mechanics. All findings were verified against the current source; no compile/test run per earlier protocol.

🤖 AI Review

PR #845, opened 2026-08-25
fix(level): translate a block state back to its real legacy value, not to a padding row

w1zardz:
### The bug

`legacy_block_data_map.json` pads every block to a power of two by repeating its default variant.
`minecraft:planks` lists oak at data **0, 6 and 7**, `minecraft:stone` lists stone at 0 and 7, and
45 more blocks do the same (`log2`, `stonebrick`, `sapling`, `torch`, `red_flower`, `wooden_door`,
`stone_slab4`, …) — 47 of 380 entries.

All of those rows map to the same block state, but `BlockPalette.registerState` kept the reverse
entry with `putIfAbsent`, so `runtimeIdToLegacy` (and `stateHashToLegacy`) ended up holding
whichever row happened to register first.

When a padding row won, the state translated back to a data value Nukkit has no block for. Oak
planks came back as `5:7` and resolved to `BlockUnknown`: hardness 10, tool type NONE.
`Level#useBreakOn` then measured every hit against that hardness, decided it was far too fast and
cancelled the event before any listener saw it — so the block could not be broken **at all**, with
any tool or by hand, in any game mode, including for operators.

Nothing was logged, because from the server's point of view each refusal is an ordinary
anti-fast-break verdict.

### Why it is hard to reproduce

Which blocks are affected depends on the order states are registered in, which in turn depends on
the set of custom blocks a server registers. On a bare build oak planks resolve to runtime id 14389
and translate back correctly (`5:0`); on our server, which registers a handful of custom blocks,
the same state is 14408 and translates back to `5:7`. The world file itself is fine either way —
the state stored on disk is a plain `minecraft:oak_planks`.

### The fix

The reverse lookup now keeps the **lowest** legacy value that maps to a state. In every padded row
the lowest value is the real one, and picking it makes the reverse lookup independent of
registration order.

Blocks that vanilla merged into another id keep their explicit preset
(`NukkitLegacyMapper#getOverrideLegacyId`), so the lava cauldron round-trip is unchanged.

### Testing

Reproduced and verified on a live 1.26.44 server with custom blocks registered: before the change
oak planks logged `roundtrip=minecraft:oak_planks|rt=14408|back=5:7` and were unbreakable; after it
they translate back to `5:0` and break normally. Stone, logs, stone bricks, torches and flowers —
other blocks with padded rows — were checked as well and behave as before.

### DIFF
--- a/src/main/java/cn/nukkit/level/BlockPalette.java
+++ b/src/main/java/cn/nukkit/level/BlockPalette.java
@@ -209,10 +209,32 @@ public class BlockPalette {
 
         int legacyId = blockId << Block.DATA_BITS | data;
         this.legacyToRuntimeId.put(legacyId, runtimeId);
-        this.runtimeIdToLegacy.putIfAbsent(runtimeId, legacyId);
+        putCanonicalLegacy(this.runtimeIdToLegacy, runtimeId, legacyId);
         int stateHash = Hash.hashBlock(blockState);
-        this.stateHashToLegacy.putIfAbsent(stateHash, legacyId);
+        putCanonicalLegacy(this.stateHashToLegacy, stateHash, legacyId);
         this.legacyToHashId.putIfAbsent(legacyId, stateHash);
+    }
+
+    /**
+     * Reverse lookup keeps the LOWEST legacy value that maps to a state.
+     * <p>
+     * The legacy data table pads every block to a power of two by repeating its default variant:
+     * {@code minecraft:planks} lists oak at data 0, 6 and 7, {@code minecraft:stone} lists stone at
+     * 0 and 7, and 45 more blocks do the same. All of those rows map to one state, so the reverse
+     * entry was decided by registration order — and when a padding row won, the state translated
+     * back to a data value that Nukkit has no block for. The block then read as an unknown one:
+     * hardness 10, no proper tool, and a vanilla break-time check that cancelled every hit, so oak
+     * planks could not be broken at all, by any tool or by hand.
+     * <p>
+     * The lowest value is the real one in every padded row, and picking it makes the reverse
+     * lookup independent of the order states are registered in. Blocks merged by vanilla into
+     * another id keep their own preset (see {@code NukkitLegacyMapper#getOverrideLegacyId}).
+     */
+    private static void putCanonicalLegacy(Int2IntOpenHashMap reverse, int key, int legacyId) {
+        int known = reverse.get(key);
+        if (known == -1 || legacyId < known) {
+            reverse.put(key, legacyId);
+        }
     }
 
     public void lock() {


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a real and important bug: padding rows in the legacy data table (e.g. `minecraft:planks` listing oak at data 0/6/7) could win the previous `putIfAbsent` reverse-lookup race and translate a state back to a data value Nukkit has no block for, making blocks (like oak planks) unbreakable. The intent — making the reverse `runtimeId`/state→legacy lookup deterministic — is sound.

However, **the chosen canonicalization rule (numeric minimum) is not safe**: it also applies to intentional `stateOverload` alias rows, not just padding rows, and regresses blocks whose normal (non-overload) row has a *larger* legacy value than an alias.

### Issues Found
#### Blocking Issues
- **[High] Numeric minimum overwrites the canonical non-overload row.** `putCanonicalLegacy` (BlockPalette.java:233) compares every row for the same runtime/state by numeric value, including deliberate overload aliases. Verified directly in `runtime_block_states_2168.dat`:
  - `minecraft:light_block_14` → runtimeId 13400 has three rows: normal `470:14` (`LightBlock`) plus overload `208:6` and `208:7` (`EndRod`). The overload marker is stripped before hashing (BlockPalette.java:173), so the state contributes no distinction. The minimum rule replaces the correct `470:14` with `208:6`, so `getLegacyFullId*`/`GlobalBlockPalette` reverse lookups now return `EndRod` for a light block.
  - `minecraft:chest` → runtimeId 14584 has normal `data=2` plus overload/padding `data=0,1,6,7`; the minimum rule selects `data=0`, not the designated canonical row.
  - This affects both `runtimeIdToLegacy` and `stateHashToLegacy`, and propagates to callers such as `RuntimeItemMapping` block-state conversion.
  - Root cause: `loadBlockStates` deliberately does two passes — normal rows first, then overloads — so the old `putIfAbsent` kept the canonical (first) row. A plain numeric minimum cannot tell padding rows (where the lowest value is correct) apart from overload alias rows (where it is not). The low-value-is-correct assumption holds only for padding rows, not for aliases where a bigger legacy id is the real block.

  Suggested fix: keep `putIfAbsent` semantics for the first (normal) pass, and only resolve ties among padding rows of the same canonical row, or carry a per-state "is this the designated row" signal (build the reverse map without the minimum rule, then prefer the non-overload entry).

#### Non-blocking Issues
- The Javadoc claim that merged blocks "keep their own preset" is only true for the LevelDB snapshot path (`NukkitLegacyMapper` supplies the cauldron override; `BlockStateSnapshot` honors it). Direct `BlockPalette`/`GlobalBlockPalette` lookups and `NukkitLegacyMapper.runtimeToFullId` still use this numeric reverse map, so the cauldron override does not make the minimum rule generally safe. The comment should scope that guarantee.
- `get` + conditional `put` costs one extra probe vs `putIfAbsent`; negligible (bulk construction only).

### Recommendations
- Preserve the canonical/non-overload row before applying any lowest-value rule, or use explicit per-state canonical metadata. Test specifically with `light_block_14` and `chest`, whose normal rows are not the minimum.

### Notes
- `known == -1` sentinel usage is correct: `defaultReturnValue(-1)` is set in the constructor and valid packed legacy IDs are non-negative.

---
🤖 AI Review

PR #846, opened 2026-08-26
fix(entity): open the horse screen so a rider can reach the saddle and armor slots

w1zardz:
### Problem

`HorseInventory` already exists with a saddle slot, an armor slot and chest storage, and
`NetworkMapping` already resolves `HORSE_EQUIP` item stack requests to it — but nothing ever
opens the window. `InteractPacket.ACTION_OPEN_INVENTORY` is only answered for a chest boat,
so the horse screen is unreachable and the armor slot can never be filled from the client.

On Bedrock this is the only way to armor a horse: once the animal is saddled, tapping it mounts
the player (`EntityHorseBase#onInteract` falls through to the mount branch), and a mounted player
does not send an interaction for the animal underneath. Reported by a player as "I can't put armor
on my horse after saddling it".

### Fix

* `HorseInventory` extends `ContainerInventory`, so opening it sends `ContainerOpenPacket` with
  `type = InventoryType.HORSE` and the horse entity id, and closing answers with
  `ContainerClosePacket` — the same contract every other entity backed container already uses.
* A rider pressing the inventory button gets the horse screen instead of their own bag.
* Crouch + use opens the same screen without mounting. Items the horse itself accepts (feed,
  saddle, horse armor, chest) keep their meaning, so a crouching player can still saddle, feed
  and armor the animal by tapping it.
* The window is removed from its viewers when the horse is closed, so a killed or despawned
  animal cannot leave an open screen behind.

### Compatibility

No new packet fields and no protocol gate: `ContainerOpenPacket` has no version branches,
`ContainerType.HORSE` is 12 on every supported version, and `ContainerSlotType.HORSE_EQUIP` is
already mapped per protocol (it is only rejected for `CRAFTER_BLOCK_CONTAINER` / `DYNAMIC_CONTAINER`,
never for horse slots). Tested on a server serving clients from 1.20 up to 1.26.45.

### Tests

Two cases added to `HorseInventoryTest`:

* `openingTheScreenAnnouncesTheHorseContainer` — opening the inventory sends a `ContainerOpenPacket`
  carrying the horse container type and the horse entity id;
* `crouchingOpensTheScreenWithoutMounting` — a crouching player gets the window and the interaction
  does not consume the held item.

`mvn test -Dtest=HorseInventoryTest` is green on current master (5 tests).

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -21,6 +21,7 @@ import cn.nukkit.entity.item.*;
 import cn.nukkit.entity.mob.EntityWalkingMob;
 import cn.nukkit.entity.mob.EntityWolf;
 import cn.nukkit.entity.passive.EntityHappyGhast;
+import cn.nukkit.entity.passive.EntityHorseBase;
 import cn.nukkit.entity.passive.EntityVillager;
 import cn.nukkit.entity.projectile.EntityArrow;
 import cn.nukkit.entity.projectile.EntityThrownTrident;
@@ -4620,6 +4621,12 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                                 if (this.protocol >= ProtocolInfo.v1_19_0) {
                                     this.addWindow(((InventoryHolder) targetEntity).getInventory());
                                 }
+                            } else if (this.riding instanceof EntityHorseBase horseRide
+                                    && (targetEntity == null || targetEntity == this.riding)) {
+                                // Bedrock has no other way to reach the saddle and armor slots:
+                                // the rider presses the inventory button and expects the horse
+                                // screen, not their own bag.
+                                this.addWindow(horseRide.getInventory());
                             } else if (this.protocol >= 407) {
                                 if (this.inventory.open(this)) {
                                     this.inventoryOpen = true;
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The PR makes the horse equipment screen reachable on Bedrock: `HorseInventory` now extends `ContainerInventory` so opening sends a `ContainerOpenPacket` with the horse container type and entity id, a rider pressing the inventory button gets the horse screen instead of their own bag, and a crouching player can open the same screen without mounting. The tests run green (5/5).

### Issues Found

#### Blocking Issues
- None.

#### Non-blocking Issues
- The sneaking item predicate (`EntityHorseBase.java:144-146`) is broader than "an item the horse itself accepts": it suppresses opening for every chest and every horse-armor item regardless of the concrete horse type or state. A normal horse holding a chest becomes a no-op (it has no chest interaction), and a donkey's chest context or a llama's armor rejection aren't reflected. Continuing to use the item on a horse it can't actually take simply does nothing — minor UX edge, not a defect.
- `EntityHorseBase#close` snapshot of the viewer set avoids a `ConcurrentModificationException`; no duplicate-removal risk. Fine as-is.

### Recommendations
- Consider a capability/state-aware helper for the item predicate so it matches the direct interaction branches (feed, saddle, usable armor, chest) exactly, keeping the screen-opening behavior consistent across horse types.
- The new tests call `inventory.onOpen()` directly and only cover a diamond held item for the crouching path; they don't exercise `Player.addWindow` window registration, horse-close viewer cleanup, or the feed/saddle/armor/chest predicate cases. Adding coverage for those would strengthen the change.
- Optional defensive guard: `Player.java:4624-4629` could check `horseRide.getInventory() != null` and `!horseRide.isClosed()` before opening. Not a current blocker.

### Notes
- Verified `git diff --check` is clean and `HorseInventoryTest` passes all 5 tests (including the two new ones) on this branch.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

PR #847, opened 2026-08-26
fix(inventory): measure an empty slot by the item being added

w1zardz:
### Problem

`Inventory#canAddItem` sizes every **empty** slot by the stack limit of whatever already sits in
it. An empty slot holds air, and air stacks to 64, so 61 golden horse armours "fit" into a single
free slot: the check passes, `addItem` then puts one armour per slot and hands the rest back.

A plugin that charges before it gives takes payment for 61 items and delivers 4. This is how it
surfaced for us: a shop sold 61 pieces of horse armour, the money was taken for all of them and
the overflow silently disappeared. Every non-stackable item is affected — armour, tools, weapons,
potions, buckets, boats, saddles, maps, spawn eggs.

### Fix

Size an empty slot by the **incoming** item, which is exactly what `addItem` already does when it
fills one. Stackable items keep the numbers they had; only the empty-slot arithmetic changes.

### Tests

`BaseInventoryCanAddItemTest` covers both directions (non-stackable overflow is now refused,
stackable capacity is unchanged). `mvn test -Dtest=BaseInventoryCanAddItemTest` is green on
current master.

### DIFF
--- a/src/main/java/cn/nukkit/inventory/BaseInventory.java
+++ b/src/main/java/cn/nukkit/inventory/BaseInventory.java
@@ -327,10 +327,12 @@ public abstract class BaseInventory implements Inventory {
         int count = item.getCount();
         boolean checkDamage = item.hasMeta();
         boolean checkTag = item.getCompoundTag() != null;
+        // The stack limit of the item being added, not of whatever occupies the slot right now:
+        // an empty slot holds air, and air stacks to 64 even when the incoming item does not.
+        int maxStackSize = Math.min(item.getMaxStackSize(), this.getMaxStackSize());
         int i1 = this.getSize();
         for (int i = 0; i < i1; ++i) {
             Item slot = this.getItemFast(i);
-            int maxStackSize = Math.min(slot.getMaxStackSize(), this.getMaxStackSize());
             if (item.equals(slot, checkDamage, checkTag)) {
                 int diff;
                 if ((diff = maxStackSize - slot.getCount()) > 0) {
--- a/src/test/java/cn/nukkit/inventory/BaseInventoryCanAddItemTest.java
+++ b/src/test/java/cn/nukkit/inventory/BaseInventoryCanAddItemTest.java
@@ -1,0 +1,71 @@
+package cn.nukkit.inventory;
+
+import cn.nukkit.MockServer;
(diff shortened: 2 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The change correctly fixes the reported bug: `Inventory#canAddItem` now sizes an **empty** slot by the incoming item's stack limit instead of air (which stacks to 64). This is the right direction and resolves the real issue. However, the refactor also changed the semantics of **matching occupied slots**, introducing a regression for items that share an ID but differ in their stack limit.

### Issues Found
#### Blocking Issues
- **Matching-slot stack limit now uses the incoming item instead of the resident item.** The loop computes `maxStackSize = min(item.getMaxStackSize(), invMax)` once and reuses it for both the matching-slot and empty-slot branches. Previously the matching branch used `min(slot.getMaxStackSize(), invMax)`. Because `item.equals(slot, checkDamage, checkTag)` ignores `meta` when the incoming item has no meta (`checkDamage == false`), an item with the same ID but a different stack limit can match a slot and be counted against the wrong limit. Concretely: adding **empty buckets** (`meta == 0`, max stack 16) to a slot already holding a **filled bucket** (`meta != 0`, max stack 1) — `equals` matches on ID (meta skipped), and the new code incorrectly treats the occupied slot as having capacity for 16 instead of 1, so `canAddItem` can return `true` when it should not.

### Recommendations
- Preserve two distinct limits, as the original code did:
  - **Empty slot:** `min(item.getMaxStackSize(), invMax)`
  - **Matching occupied slot:** `min(slot.getMaxStackSize(), invMax)`
- Add a test covering a 16-stack item and an inventory limit below the item limit, plus a case with differing resident/incoming limits (e.g. filled bucket in slot vs empty bucket being added) to lock in the correct behavior.
- Add an invariant-style test: `canAddItem(item) == true` implies `addItem(item)` returns no remainder on an otherwise identical inventory.

### Notes
- Per-slot `getMaxStackSize()` calls remain cheap; the empty-slot optimization is fine to keep by caching the incoming-item limit for the air branch only.
- `git diff --check` passes; the targeted Maven test could not be completed in the review sandbox.

---
🤖 AI Review

tachibana-kanade-angel:

PR #848, opened 2026-08-26
fix(item): give a block above id 255 its real item form

w1zardz:
### Problem

`Item.get(id)` treats every id as an item id. For a block whose id is above 255 that is wrong:
the item form of such a block is the negative alias (`255 - blockId`), the one the network item
palette knows. `Item.get(559)` therefore returns a bare `Item(559)` that no protocol can encode —
`getNamespaceId()` throws `Unknown legacy2Runtime mapping: id=559`.

Several blocks hand out exactly that as their drop (`BlockBricksQuartz`,
`BlockBricksNetherChiseled`, `BlockBricksNetherCracked`, `BlockBricksResin`,
`BlockCampfireSoul`), so mining quartz bricks yields an item that breaks every inventory path
touching it, and the broken id survives in the player's NBT.

### Fix

Map such an id to its negative alias, one hop, before falling back to a bare `Item`.
447 of the 458 blocks above id 255 without their own item class are fixed by this; the rest keep
the old fallback.

### Tests

Compiles and runs on current master; verified in game on a server serving 1.20 … 1.26.45 clients
(quartz bricks now drop a usable item and survive a relog).

### DIFF
--- a/src/main/java/cn/nukkit/item/Item.java
+++ b/src/main/java/cn/nukkit/item/Item.java
@@ -1070,6 +1070,25 @@ public class Item implements Cloneable, BlockID, ItemID, ItemNamespaceId, Protoc
             Item item;
 
             if (c == null) {
+                // A block id above 255 is not an item id. The item form of such a block is its
+                // negative alias (255 - blockId) -- that is the one the network item palette
+                // knows. Without this a block that hands out Item.get(BlockID.X) as its drop
+                // (quartz bricks, chiseled and cracked nether bricks, resin bricks, soul soil)
+                // produces an item nothing can encode: every getNamespaceId() on it throws
+                // "Unknown legacy2Runtime mapping", and the same broken id comes back from NBT
+                // on every load.
+                Class<?> blockClass = null;
+                if (id > 255) {
+                    if (id >= CustomBlockManager.LOWEST_CUSTOM_BLOCK_ID) {
+                        CustomBlockManager custom = CustomBlockManager.get();
+                        blockClass = custom == null ? null : custom.getClassType(id);
+                    } else if (id < Block.list.length) {
+                        blockClass = Block.list[id];
+                    }
+                }
+                if (blockClass != null) {
+                    return get(255 - id, meta, count, tags);
+                }
                 item = new Item(id, meta, count);
             } else if (id < 256 && id != 166) {
                 if (meta >= 0) {


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR changes `Item.get()` so that a positive id above 255 with no registered `Item` class, but which exists in `Block.list`, is converted to its negative block alias (`255 - id`). The intent is sound — blocks with ids > 255 (quartz bricks, chiseled/cracked nether bricks, resin bricks, soul soil) currently hand out an item form that can't be encoded. However, the implementation conflates the item and block id spaces, which breaks certain legitimate item ids.

### Issues Found

#### Blocking Issues
- **Positive item ids with block-id collisions are misclassified** (Item.java, new `id > 255` branch). The item id space and block id space overlap for ids > 255. When a caller does `Item.get(442)`, `Item.list[442]` is null but `Block.list[442]` is `BIRCH_WALL_SIGN` (BlockID.java:400), so the new code returns the negative alias `-187` (birch wall sign) — even though `442` is a real item id (`ItemID.SPARKLER`, ItemID.java:234). The same collision affects 447, 448, 449, 451, 453, 498, 499, and 645 (`CHEST_BOAT`, ItemID.java:302, vs `DEEPSLATE_TILE_WALL`, BlockID.java:601). A positive id cannot be treated as a block merely because a block happens to share the numeric value. This silently changes what an item request returns and can surface broken/duplicate item stacks at runtime.

#### Non-blocking Issues
- (none specific beyond the above)

### Recommendations
- Restrict the new lookup so it only applies when the caller is genuinely requesting a **block's item form** rather than any positive id. Avoid inferring block-ness from a numeric collision alone. Consider checking an explicit namespace-id/registry source (the same palette the comment references) instead of `Block.list[id]`, or gate the conversion on the id being from the block/material set the PR targets.
- The `id < Block.list.length` guard is present, which prevents an out-of-range read; also note the outermost `try/catch` in `Item.get` will swallow a hypothetical failure and fall back to the raw id (so no crash). Keep that behavior in mind when tightening the branch.
- Add a regression test for `Item.get(442)` (sparkler) and `Item.get(645)` (chest boat) to pin their expected item form, plus one for a genuine block-above-255 (e.g. quartz bricks / soul soil) to confirm the negative alias.

### Notes
- Custom block ids (`>= 10000`) are routed through `CustomBlockManager`; conversion applies to them, which is intended. The theoretical `id >= 65535` case is not newly reachable by this PR's guard logic.

---
🤖 AI Review

PR #849, opened 2026-08-26
fix(inventory): never drop creative UI slots on the floor

w1zardz:
### Problem

Closing an anvil, a grindstone, a smithing table, a stonecutter, a cartography table, an
enchanting table, a trade window or a beacon returns whatever sat in the temporary UI slots to the
player and drops the remainder into the world.

In creative that remainder is free: the items came out of an infinite inventory, so a player with
a full inventory turns every closed station into real item entities on the floor — which a
survival player then picks up. The drop also bypasses the plugin path that decides whether a
player may drop items at all.

### Fix

Creative UI slots are cleared instead of being returned or dropped. The anvil also stops dropping
its two slots unconditionally: like every other station it puts them back into the inventory first
and only drops what does not fit.

Survival behaviour is unchanged — the items still come back to the player, and only the overflow
is dropped.

### Tests

Compiles on current master; verified in game (creative player with a full inventory closing each
of the eight stations leaves nothing on the ground, survival player keeps his items).

### DIFF
--- a/src/main/java/cn/nukkit/inventory/AnvilInventory.java
+++ b/src/main/java/cn/nukkit/inventory/AnvilInventory.java
@@ -31,8 +31,16 @@ public class AnvilInventory extends FakeBlockUIComponent {
         who.resetCraftingGridType();
 
         for (int i = 0; i < 2; ++i) {
-            this.getHolder().getLevel().dropItem(this.getHolder().add(0.5, 0.5, 0.5), this.getItem(i));
+            Item item = this.getItem(i);
             this.clear(i);
+            if (item.isNull() || who.isCreative()) {
+                continue;
+            }
+            for (Item drop : who.getInventory().addItem(item)) {
+                if (!who.dropItem(drop)) {
+                    this.getHolder().getLevel().dropItem(this.getHolder().add(0.5, 0.5, 0.5), drop);
+                }
+            }
         }
     }
 
--- a/src/main/java/cn/nukkit/inventory/BeaconInventory.java
+++ b/src/main/java/cn/nukkit/inventory/BeaconInventory.java
@@ -26,7 +26,7 @@ public class BeaconInventory extends FakeBlockUIComponent {
         super.onClose(who);
 
         // Drop item in slot if client doesn't automatically move it to player's inventory
-        if (!who.isConnected()) {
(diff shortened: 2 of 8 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR clears creative-mode UI slot contents on station close instead of returning/dropping them, preventing creative players from spilling infinite-inventory items onto the floor (which survival players could then pick up). Survival behavior is intended to stay unchanged. Seven of the eight inventory classes handle this correctly, but the anvil's cleanup path has an ordering defect that leaves its core scenario unfixed.

### Issues Found
#### Blocking Issues
- **AnvilInventory.onClose() ordering** (`AnvilInventory.java:31`): `onClose` calls `who.resetCraftingGridType()` *before* the slot-clearing loop. At close time the anvil window is still registered in `Player.windows`, so `resetCraftingGridType()` → `moveBlockUIContents(ANVIL_WINDOW_ID)` runs `inventory.addItem(contents)` and **drops the overflow onto the floor**, only then clearing the slots. The subsequent loop `this.getItem(i)` therefore sees already-emptied slots. Consequence: a creative player with a full inventory closing the anvil can still produce item entities on the ground — the exact scenario this PR claims to eliminate is NOT actually fixed for the anvil. Move `resetCraftingGridType()` after the slot loop (or make the moveBlockUIContents path creative-aware / skip it for anvil).

#### Non-blocking Issues
- **Beacon disconnected-survival path still drops the whole payment** (`BeaconInventory.java`): when quitting while disconnected, the single input slot is dropped outright without first attempting to place it back into the inventory. Pre-existing behavior, but it contradicts the "only overflow is dropped" invariant advertised in the PR description.
- **No regression test** was added for the creative-close behavior across the eight stations; an anvil close test would have exposed the blocking ordering issue above.

### Recommendations
- Reorder AnvilInventory cleanup as described to make the creative no-drop guarantee actually hold.
- Add inventory-close regression tests covering a creative player with a full inventory for each affected station.
- Consider normalizing the beacon disconnected path to attempt inventory-return before dropping.

### Notes
- All eight classes keep public API signatures unchanged; slot indices are in bounds and retrieved items are non-null.
- Survival-return behavior for the seven non-anvil stations is preserved.

---
🤖 AI Review

PR #851, opened 2026-08-26
entity: keep controllable rides moving on 1.21.80 and newer

w1zardz:
### Problem

Since protocol `v1_21_80` the client no longer sends `PlayerInputPacket`, and
`PlayerInputProcessor#isSupported()` stops at that version. The auth input path only forwards the
move vector to minecarts, happy ghasts and boats, so every other `EntityControllable` ride loses
its input entirely: a saddled horse, pig, llama or strider can still be mounted, but it never
moves again on a modern client. Nothing is logged — the rider simply sits still.

### Fix

Forward the auth input move vector to any other `EntityControllable` the player is riding, on the
same terms the legacy packet used: clamped to `[-1, 1]` and only when the stick is actually
pushed, so an idle rider does not pin the mob yaw to the player's every tick.

### Tests

Compiles on current master; verified in game on 1.26.4x — a saddled horse, pig and strider are
controllable again, and older clients still ride through the legacy packet path unchanged.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4079,6 +4079,19 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                             }
                         }
                     }
+                } else if (this.riding instanceof EntityControllable controllable
+                        && this.protocol >= ProtocolInfo.v1_21_80) {
+                    // Since 1.21.80 the client stopped sending PlayerInputPacket, so
+                    // PlayerInputProcessor never runs any more and every controllable ride
+                    // except the boat, the minecart and the happy ghast lost its input: a
+                    // saddled horse could be mounted and then stood still forever. The move
+                    // vector lives in the auth input packet now, and it drives the very same
+                    // EntityControllable hook the legacy packet used to drive.
+                    double moveVecX = NukkitMath.clamp(authPacket.getMotion().getX(), -1, 1);
+                    double moveVecY = NukkitMath.clamp(authPacket.getMotion().getY(), -1, 1);
+                    if (moveVecX != 0 || moveVecY != 0) {
+                        controllable.onPlayerInput(this, moveVecX, moveVecY);
+                    }
                 }
 
                 if (authPacket.getInputData().contains(AuthInputAction.START_SPRINTING)) {


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR fixes controllable-ride movement for protocol `v1_21_80` and newer, where the client stopped sending `PlayerInputPacket` so `PlayerInputProcessor#isSupported()` no longer runs. The new branch in the auth-input handler forwards the auth packet's motion vector to any other `EntityControllable` ride on the same terms as the legacy packet path: clamped to `[-1, 1]` and only when the stick is pushed.

The fix correctly mirrors legacy behavior:
- Axis mapping (`x` = strafe, `y` = forward) matches `PlayerInputProcessor`.
- Clamping to `[-1, 1]` and skipping `(0,0)` matches the legacy processor.
- `this.riding instanceof EntityControllable` is null-safe.
- `else-if` ordering keeps boats, minecarts, and happy ghasts on their existing specialized paths with no duplicate callbacks.
- Protocol gating exactly complements `PlayerInputProcessor.isSupported()` (legacy stops at v800); clients below v1_21_80 remain on the legacy path unchanged.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `NukkitMath.clamp(NaN, ...)` remains NaN, and `NaN != 0` passes, so a custom `EntityControllable` could receive NaN. However, the existing auth boat/ghast handling has the same weakness — follow-up hardening rather than a merge blocker.
- In the deprecated client-authoritative movement mode (v800–817), the auth handler returns early and `PlayerInputProcessor` is disabled, so those clients still have no controllable-ride input path. Only relevant if that deprecated config is still promised.
- The auth path lacks the legacy processor's explicit alive/spawned guard; a stale riding reference could theoretically let a dead player drive a mount. Same pre-existing behavior as minecart/boat/ghast auth paths.
- The PR description overstates llama support: `EntityLlama.onPlayerInput()` is intentionally a no-op. The new branch correctly preserves that legacy semantics, but llamas remain non-controllable.

### Recommendations
- Optionally add a `Float.isFinite` check (or reuse `validateVehicleInput`) on the auth motion values for the controllable path, consistent with the boot-time input validation.
- Consider tracking the deprecated client-auth mode (v800–817) separately; not required for this PR.

### Notes
No sign inversion, null-riding issue, or regression to the boat/minecart/happy-ghast paths was found. Changes are small and isolated to `Player.java` (13 insertions). Safe to merge.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #852, opened 2026-08-26
block: let a flying player mine at normal speed

w1zardz:
### Problem

Bedrock multiplies the mining time by five only while the player is **jumping or falling**.
Flight is a stable state there, and the client keeps mining at the normal rate with it on; the
server does not, so it demands five times the client time and `Level#useBreakOn` cancels the break
as too fast.

A flying player cannot break a wooden plank at all, and nothing reaches the log: from the server
side the refusal is an ordinary anti-fast-break verdict, which makes it very hard to track down.

### Fix

Skip the in-air factor while the player is flying. PocketMine-MP draws the same line
(`SurvivalBlockBreakHandler` skips it when `isFlying()`), and it serves the same clients.

Only the flight case changes: jumping and falling keep the penalty, and the five-tick grace for a
landing player is untouched.

### Tests

Compiles on current master; verified in game — a flying player in survival breaks blocks at the
client-side rate, a jumping or falling one still takes the vanilla penalty.

### DIFF
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -1,5 +1,6 @@
 package cn.nukkit.block;
 
+import cn.nukkit.AdventureSettings;
 import cn.nukkit.Player;
 import cn.nukkit.Server;
 import cn.nukkit.block.custom.CustomBlockManager;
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a flying-player mining bug: the bedrock client keeps mining at the normal rate while flight is on, but the server applied the in-air ×5 penalty and rejected the break as too fast. The change skips the airborne penalty only when `FLYING` is active, while jumping/falling still keep the penalty and the 5-tick landing grace is untouched. The approach matches PocketMine-MP's `SurvivalBlockBreakHandler`.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- No regression test was added covering flying vs. jumping/falling behavior (verified in-game only).

### Recommendations
- Consider adding a test that exercises `calculateBreakTime` with `FLYING` on/off to prevent future regressions.

### Notes
- `git diff --check` passed. Build could not run locally because Maven snapshot dependencies require network access that is unavailable in the review environment.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #853, opened 2026-08-26
fix(portal): make a nether trip land somewhere sane without generating the search box

w1zardz:
Four related fixes on the nether portal path; they build on each other, so they come as one branch.

### 1. Search the nether side by rings, without generating chunks

`findNearestPortal` walked a fixed 257×257 box over every Y level and kept going after a hit
(the inner `break` only left the Y loop), so a single portal step always paid for the full box.
The lookup went through `Level#getBlockIdAt`, which **generates** a missing chunk on access —
roughly 290 chunks generated inside an `AsyncTask` before the player is teleported anywhere. On a
fresh nether that is seconds of waiting per portal.

Now: rings around the mirrored position, returning on the first portal found (which is also the
nearest). Chunks are loaded from disk but never generated — a portal that does not exist yet is
built next to the mirrored position anyway. Block reads go straight to the chunk, and the Y range
follows the world instead of starting at zero.

### 2. Bring the way back to the world the player left, on the ground

`Level#calculatePortalMirror` hands every way back to `Server#getDefaultLevel()`, which is the
overworld only on a server whose single world is also the default one. With a hub as the default
world the player lands on a hub coordinate mirrored eight times out from the build, over terrain
nobody generated. And `spawnPortal` built the frame at the Y taken from the other side, so even in
the right world the frame hung in mid-air with the player standing on its obsidian floor.

The origin world is remembered in the player data when the portal is entered (so a relog or a
restart inside the nether keeps it, and the default world stays the fallback for players who were
already there). The frame is lowered onto the first footing below the mirrored position, or raised
onto the first one above it, and `spawnPortal` returns where it ended up.

### 3. Keep a fresh frame out of lava and give it a floor

A frame spawned in the nether could be placed inside a lava lake or with no floor under it.

### 4. Do not read hundreds of cold chunks for one transfer

The ring search stopped *generating* the target area, but it still loaded every generated chunk in
a 19×19 square from LevelDB. With chunk caching disabled and the Java LevelDB provider, a return
from the nether without a nearby indexed frame spent 30–40 seconds reading up to 361 cold chunks
and scanning ~24 million block cells.

Only cold chunks in the origin ring and its immediate neighbours are loaded now — enough to
recover the portal used for the trip, since coordinate scaling loses at most seven blocks.
Already-loaded chunks are still searched across the vanilla radius, so an active farther portal
keeps linking as before.

### Tests

Compiles on current master; verified in game on a server serving 1.20 … 1.26.45 clients: both
directions land on solid ground in the correct world, and the transfer is immediate instead of
taking tens of seconds on a cold nether.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -2226,9 +2226,14 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
         if (this.server.isNetherAllowed()) {
             if (this.server.vanillaPortals && (this.inPortalTicks == 40 || this.inPortalTicks == 10 && this.gamemode == CREATIVE) && this.portalPos == null) {
-                Position portalPos = this.level.calculatePortalMirror(this);
+                Position portalPos = this.level.calculatePortalMirror(this, this.netherReturnLevel());
                 if (portalPos == null) {
                     return;
+                }
+                if (this.level.getDimension() != Level.DIMENSION_NETHER) {
+                    // Remember which world is being left: the way back has to land here and not in
+                    // whatever world happens to be the default one.
+                    this.namedTag.putString(NETHER_RETURN_WORLD_TAG, this.level.getFolderName());
                 }
 
                 for (int x = -1; x < 2; x++) {
@@ -2265,8 +2270,10 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                             Position foundPortal = BlockNetherPortal.findNearestPortal(portalPos);
                             getServer().getScheduler().scheduleTask(InternalPlugin.INSTANCE, () -> {
                                 if (foundPortal == null) {
-                                    BlockNetherPortal.spawnPortal(portalPos);
-                                    teleport(portalPos.add(1.5, 1, 0.5), TeleportCause.NETHER_PORTAL);
+                                    // The frame may end up at another height than asked for, so the
+                                    // player follows the frame instead of the mirrored position.
+                                    Position spawnedPortal = BlockNetherPortal.spawnPortal(portalPos);
+                                    teleport(spawnedPortal.add(1.5, 1, 0.5), TeleportCause.NETHER_PORTAL);
                                 } else {
                                     teleport(BlockNetherPortal.getSafePortal(foundPortal), TeleportCause.NETHER_PORTAL);
                                 }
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR reworks the nether portal path to avoid generating the whole 257×257 search box on every transfer: ring-based portal lookup, remembered return world, ground-seeking frame placement, and lava-safe sealing. The direction is sound and the improvements are real, but there are **three blocking issues** — a public API binary-compatibility break, an off-main-thread chunk lifecycle problem that defeats the PR's own cold-chunk limit, and a generation race that can build a frame on ungenerated terrain.

### Issues Found

#### Blocking Issues
1. **`spawnPortal` return-type change breaks plugin binary compatibility** — `BlockNetherPortal.java:385`
   The public static method changes from `void spawnPortal(Position)` (master) to `Position spawnPortal(Position)`. The JVM method descriptor includes the return type, so any already-compiled plugin calling `spawnPortal` will fail at runtime with `NoSuchMethodError`. Keep the original `void` method and add a differently-named result-returning method (e.g. `spawnPortalAndGetPosition`), then call the new one from `Player`.

2. **`getChunk(x, z, false)` loads cold chunks from disk in every ring, not just the origin — defeating `PORTAL_DISK_LOAD_CHUNK_RADIUS`** — `BlockNetherPortal.java:343-350`
   In `Level#getChunk`, the `false` only means "do not create a missing chunk"; an existing on-disk chunk is still synchronously loaded, mounted, and initialized (`forceLoadChunk` → `finishChunkLoad`). Because the outer rings call `getChunk(..., false)` unconditionally, a no-hit search can still mount up to 19×19 = 361 persisted chunks — exactly the "reading hundreds of cold chunks" case the PR says it eliminates. Worse, this runs inside an `AsyncTask` worker thread, so `ChunkLoadEvent`, entity/block-entity initialization, and light population fire off the main thread, plus unload-queue and loader notifications mutate shared `Level` state. Use `getChunkIfLoaded()` for the cache-only branch and, for the permitted cold radius, read chunks through the off-thread provider path without mounting/initializing them.

3. **Generation is scheduled but never awaited — a frame can be built on ungenerated terrain** — `Player.java:2239-2248`, `BlockNetherPortal.java:328-364`
   `generateChunk(..., true)` dispatches a `GenerationTask` asynchronously. The ring search skips chunks that are null or not yet generated, then `spawnPortal`/`frameHeightFor` proceed to build the frame — potentially inspecting empty pre-generation terrain, choosing the fallback Y and building in mid-air, and racing against the in-flight generation which can then overwrite the portal. This is most reachable for creative-mode players (only 15 tick prep). Chunk generation must complete before the search/spawn runs, or the spawn should be chained to generation completion.

#### Non-blocking Issues
- `BlockNetherPortal.java:479-499`: `frameHeightFor` caps `start` at `maxY = 122` in the Nether, but a footing found at Y=122 yields a portal floor at Y=123 whose top cap is written at Y=127 — replacing the Nether roof bedrock with mineable obsidian. Edge case (needs solid terrain nearly at the roof), but the complete frame should stay below the roof (footing ≤ 121).
- `BlockNetherPortal.java:304-321`: the Javadoc claims the search returns the *nearest* portal, but it returns the first hit in ring order (Chunk Chebyshev distance ≠ block distance; within a chunk it scans local X/Z then ascending Y and ignores target Y; boundary chunks can match portals slightly beyond 128 blocks). No worse than master, but the "closest" wording is inaccurate.
- `Level.java:5502-5519`: multi-Nether return trips (a `foo` ↔ `foo-nether` pair) remain broken because `getNetherWorld("foo-nether")` resolves to the default `nether`, so `this == nether` is false inside `foo-nether` and the remembered source world is ignored (coordinates get divided and the destination is the default nether). Pre-existing limitation, unchanged by this PR.
- `Player.java:2325-2342`: `NetherReturnWorld` is written before the portal event is accepted and is never cleared after a successful return; cancelled trips and plugin-driven nether travel can reuse a stale world.
- `Level.java:5515-5522`: X/Z scaling handles negative coordinates correctly, but the hard-coded Y clamps remain unsuitable for custom-height dimensions.
- `BlockNetherPortal.java:450-473`: `sealPortalSite` uses raw `setBlockAt`, so protection plugins get no placement/construction event; policy-dependent.
- No regression tests cover negative coordinates, chunk boundaries, multiple portals at different Y, multi-nether pairs, cold chunks, generation delays, or the roof boundary.

### Recommendations
1. Preserve the `void spawnPortal(Position)` overload and return the spawned position from a new method.
2. Restrict disk-backed reads to the origin ring's immediate neighbourhood via the off-thread provider API and use `getChunkIfLoaded()` everywhere else; keep all chunk lifecycle/initialization on the main thread.
3. Chain portal search/spawn after chunk generation completes for the affected footprint.

### Notes
Changes touch `cn/nukkit/Player.java`, `cn/nukkit/block/BlockNetherPortal.java`, and `cn/nukkit/level/Level.java`. `git diff --check` is clean; no binaries or suspicious patterns detected. Static review via Codex, with blocking findings independently verified against the PR head.

---
🤖 AI Review

PR #854, opened 2026-08-26
block: match the vanilla shulker box hardness

w1zardz:
The client computes its own break animation from the vanilla block data, where a shulker box has a
destroy time of `2.0`, while `BlockShulkerBox` reports `2.5`. The server therefore always expects
the break to take 25 % longer than the client shows, marks the finished break as too fast and
cancels it: with a hand, a wooden or a stone pickaxe the box never breaks at all, and with an iron
or better pickaxe it only survives because the 150 ms grace covers the gap.

Report the vanilla value so both sides agree.

Compiles on current master; verified in game — a shulker box now breaks with any tool.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockShulkerBox.java
+++ b/src/main/java/cn/nukkit/block/BlockShulkerBox.java
@@ -59,7 +59,7 @@ public class BlockShulkerBox extends BlockTransparentMeta implements BlockEntity
 
     @Override
     public double getHardness() {
-        return 2.5;
+        return 2;
     }
 
     @Override


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The change lowers `BlockShulkerBox.getHardness()` from `2.5` to `2.0`, matching the vanilla shulker-box destroy time. Since `BlockUndyedShulkerBox` extends `BlockShulkerBox`, both dyed and undyed shulker boxes inherit the corrected value — no further table or override needs updating. This removes the 25% timing mismatch between the client-side break animation and the server-side fast-break validation, so the box now breaks consistently with any tool as described.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- None

### Notes
- Resistance, pickaxe classification, drops, block entities, and creative-mode behavior are unchanged. The correction shortens survival/adventure mining times by ~20%, which is intended; plugins relying on the previous incorrect timing may observe a behavior change.

---
🤖 AI Review

@lt-name 可以合并

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #856, opened 2026-08-26
block: fire cancellable prime events before beds and anchors explode

w1zardz:
### Problem

A bed used in the nether or the end, and a respawn anchor charged in the overworld, explode
directly: `Level#createExplosion` is called without ever firing `BlockExplosionPrimeEvent`. A
plugin guarding land has no way to stop it, so a protected build can be blown up with a bed from
outside the region, and nothing about the blast is visible to plugins.

### Fix

Fire a cancellable `BlockExplosionPrimeEvent` (with the player who triggered it and the vanilla
force of 5) before both explosions. A cancelled event leaves the block and the world untouched.

### Tests

`BlockExplosionPrimeEventTest` covers both blocks, cancelled and not.
`mvn test -Dtest=BlockExplosionPrimeEventTest` is green on current master.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockBed.java
+++ b/src/main/java/cn/nukkit/block/BlockBed.java
@@ -5,6 +5,7 @@ import cn.nukkit.Server;
 import cn.nukkit.blockentity.BlockEntity;
 import cn.nukkit.blockentity.BlockEntityBed;
 import cn.nukkit.entity.Entity;
+import cn.nukkit.event.block.BlockExplosionPrimeEvent;
 import cn.nukkit.item.Item;
 import cn.nukkit.item.ItemBed;
 import cn.nukkit.lang.TranslationContainer;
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. Fires a cancellable `BlockExplosionPrimeEvent` before both bed (nether/end) and respawn anchor (overworld) explosions, giving plugins a way to prevent protected builds from being destroyed. Cancellation leaves the block and world untouched, with unit tests covering both blocks in cancelled and non-cancelled paths.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- `BlockBed.java:120` — When a plugin sets `event.setBlockBreaking(false)`, only the activated bed half is explicitly removed. Skipping `explodeA()` leaves the other half behind as an invalid orphan bed block. Both halves constitute the explosion source and should be removed independently of affected-block destruction, consistent with the respawn anchor consuming itself. This path is currently untested.

### Recommendations
- Consider removing both bed halves when block breaking is disabled.
- Consider adding a test for the `setBlockBreaking(false)` path to cover the orphan-half case.

### Notes
- `git diff --check` passes. Runtime tests could not be completed during review because concurrent workspace branch checkouts interfered with compilation, but the analysis was based on the immutable PR commit diff.

---
🤖 AI Review

@lt-name ready to merge

tachibana-kanade-angel (approved):

PR #857, opened 2026-08-26
fix(inventory): refuse an item stack request with an unhandled action

w1zardz:
### Problem

An action type without a registered processor was logged and skipped, and the rest of the request
kept running on top of a step that never happened. The client had already predicted every action,
so the two sides drifted apart silently: depending on where the skipped action sat in the chain,
the player either kept an item the server never moved or lost one it never received.

### Fix

Fail the request instead. The existing error path already restores the snapshots taken before each
action, resynchronises the actor in full (cursor, every open window, held item, armour, offhand)
and answers `ERROR`, so the client rolls its prediction back to server state.

Every action type of the enum has a processor today, so this only fires when a new protocol brings
a type that is not implemented yet — exactly the case where guessing is worse than refusing.

### Tests

`UnhandledActionRequestTest`. Green on current master.

### DIFF
--- a/src/main/java/cn/nukkit/inventory/request/ItemStackRequestHandler.java
+++ b/src/main/java/cn/nukkit/inventory/request/ItemStackRequestHandler.java
@@ -89,10 +89,16 @@ public final class ItemStackRequestHandler {
                         (ItemStackRequestActionProcessor<ItemStackRequestAction>) PROCESSORS.get(action.getType());
 
                 if (processor == null) {
-                    // 未注册/未实现的 action 类型静默跳过，继续处理同一请求内的后续 action，
-                    // 单条 action 不应令整条 request 失败。
-                    log.warn("{}: unhandled item stack request action {}", player.getName(), action.getType());
-                    continue;
+                    // 未注册/未实现的 action 类型不能静默跳过：同一 request 内的后续 action 会
+                    // 建立在一个从未发生的步骤之上，客户端与服务端随之静默错位。
+                    // An unregistered action type must not be skipped: the remaining actions of the
+                    // same request would keep building on a step that never happened, and client and
+                    // server drift apart with no trace. Fail the request instead, so the snapshots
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. The inventory stack request handler now refuses a request whose action has no registered processor instead of silently skipping it, preventing the client and server from drifting apart.

### Issues Found
#### Blocking Issues
- None.

The rollback is correctly transactional:
- Snapshots are captured before the first mutation of each affected inventory.
- A missing processor now stops the remaining actions and skips the commit.
- Earlier inventory changes are restored, including canonicalized cursor/crafting inventories and bundle contents.
- The actor, open inventories, observers, equipment, and cursor are resynchronized, and the response is `ERROR` so the client discards its prediction.

The new test meaningfully verifies that a successful `TakeAction` is rolled back rather than merely never executing.

#### Non-blocking Issues
- The test uses an action whose `getType()` returns `null`, which is synthetic: unknown wire action IDs are rejected during decoding in `BinaryStream.readItemStackRequest()` and never reach this handler. This PR covers a decoded action with no registered processor, not arbitrary unknown wire action IDs.
- `LabTableCombineActionProcessor` and `CraftNonImplementedActionProcessor` remain registered no-op processors and are still accepted/skipped. That is deliberate existing behavior, but the PR should not be read as refusing every semantically unimplemented action.
- The test temporarily mutates the global processor registry via reflection. It restores it in `finally` and current tests don't run in parallel, but this would be race-prone if parallel testing is enabled later.

### Recommendations
- No urgent actions needed. If parallel test execution is ever enabled, the reflection-based registry mutation in the test should be revisited.

### Notes
- No blocking correctness or security concerns found in the changed path.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #858, opened 2026-08-26
fix(rcon): answer a command whose reply arrives after dispatch

w1zardz:
### Problem

Plugin commands that read a database answer a few ticks after `dispatchCommand` returns. RCON
responded in the same call, so the operator received an empty string and read it as a failure
while the command had actually run.

### Fix

A command that returns without output keeps its connection waiting for up to two seconds and is
answered once its reply stopped growing, so multi-line replies are not cut in half. Commands that
answer straight away are untouched.

The send path also stopped assuming the channel still has a selection key: a deferred answer can
outlive the client, and reading `isValid()` off `null` threw out of `Server#tick`.

Compiles on current master; verified against an async plugin command over RCON.

### DIFF
--- a/src/main/java/cn/nukkit/network/rcon/RCON.java
+++ b/src/main/java/cn/nukkit/network/rcon/RCON.java
@@ -6,6 +6,10 @@ import cn.nukkit.event.server.RemoteServerCommandEvent;
 import cn.nukkit.utils.TextFormat;
 
 import java.io.IOException;
+import java.nio.channels.SocketChannel;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 
 /**
  * Implementation of Source RCON protocol.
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Focuses on a real deficiency: plugin commands that answer asynchronously (e.g. after a DB query) received an empty RCON reply because the response was captured in the same call that returned `dispatchCommand`. The PR defers the reply for those commands so late output is delivered, and also hardens `RCONServer.send()` against a null/invalid `SelectionKey` that previously threw out of the server tick. The change is small, well-scoped, and reads correctly. Verified against an async plugin command over RCON per the author.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- **Deferral only applies to commands with empty initial output** (`RCON.java:75-81`). A command that emits an ack synchronously (e.g. `/debug` printing `Uploading...` and sending the link later) still loses the late portion. The PR description explicitly says "commands that answer straight away are untouched", so this is a stated scope boundary rather than a bug — worth documenting as a known limitation.
- **`RemoteConsoleCommandSender` buffer is not thread-safe** (`RemoteConsoleCommandSender.java:12-29`). The deferred window now reads the `StringBuilder` multiple times across ticks; if an async plugin thread appends concurrently, output can be stale or throw mid-tick. This race predates the PR (the immediate read after `dispatchCommand` already crossed threads), so it is not newly introduced — but the longer read window increases exposure. A synchronized/immutable snapshot would remove it.

### Recommendations
- Document that only empty-initial-output commands are deferred; clarify the current `/debug`-style limitation for operators.
- Consider making `RemoteConsoleCommandSender` thread-safe (synchronized snapshot) to close the cross-thread read.
- Add tests for late output delivery, multi-tick output spanning the 40-tick deadline, and a disconnected client mid-deferral.

### Notes
- The null-key null guard in `RCONServer.send()` correctly targets the crash observed when a deferred reply outlives its client; the check-before-use is sound.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #859, opened 2026-08-26
fix(inventory): propagate event-authored workstation outputs

w1zardz:
### Problem

Workstation transactions (crafting table, enchanting table, grindstone, smithing table, anvil,
and their `ItemStackRequest` counterparts) build the network action list **before** the plugin
event is fired. A plugin that changes the output item inside the event — adds NBT, renames it,
swaps enchantments — sees its change accepted, but the player receives the pre-event clone that
was captured earlier. The event looks honoured and is not.

### Fix

Treat the event result as authoritative: after the event returns, replace every executable
occurrence of the pre-event output in the parsed actions and inventory targets, so the item the
event authored is the item delivered. A cancelled event still cancels the transaction.

Both paths are covered — legacy `InventoryTransaction` and the server-authoritative
`ItemStackRequest` processors.

### Tests

`WorkstationEventOutputRequestTest`, `EventAuthoredLegacyTransactionTest` and
`CakeCraftingReproTest` (regression: ordinary crafting keeps working). Green on current master.

### DIFF
--- a/src/main/java/cn/nukkit/inventory/request/CraftGrindstoneActionProcessor.java
+++ b/src/main/java/cn/nukkit/inventory/request/CraftGrindstoneActionProcessor.java
@@ -64,6 +64,11 @@ public class CraftGrindstoneActionProcessor implements ItemStackRequestActionPro
         if (event.isCancelled()) {
             return context.error();
         }
+        Item eventOutput = event.getNewItem();
+        if (eventOutput == null || eventOutput.isNull()) {
+            return context.error();
+        }
+        result = eventOutput.clone();
 
         int experienceDropped = event.getExperienceDropped();
         if (experienceDropped > 0) {
--- a/src/main/java/cn/nukkit/inventory/request/CraftRecipeActionProcessor.java
+++ b/src/main/java/cn/nukkit/inventory/request/CraftRecipeActionProcessor.java
@@ -46,6 +46,7 @@ public class CraftRecipeActionProcessor implements ItemStackRequestActionProcess
     public static final String RECIPE_NET_ID_KEY = "recipeNetId";
     public static final String ENCH_RECIPE_KEY = "enchRecipe";
     public static final String TIMES_CRAFTED_KEY = "timesCrafted";
+    public static final String EVENT_AUTHORED_MULTI_OUTPUT_KEY = "eventAuthoredMultiOutput";
 
     @Override
     public ItemStackRequestActionType getType() {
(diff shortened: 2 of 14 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a real bug where workstation transactions (crafting table, enchanting table, grindstone, smithing table, anvil — including their `ItemStackRequest` counterparts) built the network action list **before** the plugin `CraftItemEvent`/`EnchantItemEvent`/etc. fired. A plugin that transforms the output item in the event (add NBT, rename, swap enchantments) had its change accepted server-side, but the player received the pre-event clone. The fix treats the event result as authoritative: after the event returns, `applyEventOutputToActions` rewrites every executable occurrence of the pre-event output in the parsed actions and inventory targets, so the authored item is what gets delivered. Both legacy `InventoryTransaction` and `ItemStackRequest` paths are covered.

Verification: `CakeCraftingReproTest` (13), `WorkstationEventOutputRequestTest` (5), and `EventAuthoredLegacyTransactionTest` (6) all pass (24 tests, 0 failures). `git diff --check` is clean. All 3 issues initially flagged by the automated reviewer were independently verified as non-blocking edge cases (see below).

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- `InventoryTransaction.applyEventOutputToActions` matches source/target items via `equalsExact` (ID + damage + NBT + **count**). When the event-authored output merges into an already-stacked destination (e.g. output delivered into a slot that already holds the same item, so the target count exceeds the original output count), the match misses and the pre-event item is delivered in that narrow stacking case. Single-item workstation output slots (the primary path) are unaffected. Consider matching on ID+damage+NBT while updating count separately, or documenting this as a known limit.
- `CraftResultDeprecatedActionProcessor` stores the multi-result list as a singleton after re-authoring. The branch only fires for `recipe instanceof MultiRecipe`, whose concrete subclasses all produce a single output via `toRecipe` (no `getExtraResults`), so no secondary results are actually dropped today. This is a latent constraint rather than a current bug — worth a comment noting that genuine multi-output `MultiRecipe` implementations would need the full-result list rebuilt.
- `CraftingTransaction.setPrimaryOutput` now rejects a count-only change via the added `primaryOutput.getCount() != item.getCount()` check. Master only rejected item inequality (count was ignored). This is the intended conservative boundary (event output should match the recipe's item identity), and it correctly keeps the NBT/rename/enchantment authoring path open — the actual fix target.

### Recommendations
- Consider whether the stacked-destination edge case warrants handling in a follow-up (update target count rather than require exact-count match).
- Add a comment in the multi-result processors clarifying that `MULTI_RESULTS_KEY` assumes single-output `MultiRecipe` implementations.
- The trade-off between `setPrimaryOutput`'s identity/count constraint and allowing NBT authoring is reasonable; keep it documented so plugin authors know count changes are not propagated.

### Notes
- The core regression tests cover the renamed/NBT-authored delivery path for workstations and legacy crafting, and the ordinary-crafting regression case the PR description calls out.

---
🤖 AI Review

@lt-name ready to merge

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #860, opened 2026-08-26
player: trust the requested name for offline lookups

w1zardz:
`Server#getOfflinePlayer(name)` resolves the UUID through the name lookup database and then hands
the UUID to `OfflinePlayer`, which reads `NameTag` out of the saved player data. That tag is
written by the last session, so it goes stale the moment the account is renamed: whitelist, op,
ban and unban all keep operating on the previous name while the console reports the name that was
typed.

Keep the requested name on the tag instead.

Compiles on current master; verified in game — after a rename, `whitelist add <new name>` now
whitelists the account the operator meant.

### DIFF
--- a/src/main/java/cn/nukkit/OfflinePlayer.java
+++ b/src/main/java/cn/nukkit/OfflinePlayer.java
@@ -68,7 +68,12 @@ public class OfflinePlayer implements IPlayer {
         if (uuid != null) {
             this.namedTag.putLong("UUIDMost", uuid.getMostSignificantBits());
             this.namedTag.putLong("UUIDLeast", uuid.getLeastSignificantBits());
-        } else {
+        }
+
+        if (name != null) {
+            // The caller asked for this exact name, so answer with it. The name stored in the
+            // saved data belongs to the last session and goes stale as soon as the player renames
+            // their account, which silently made whitelist, op and ban act on the previous name.
             this.namedTag.putString("NameTag", name);
         }
     }


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The change is a clean, minimal fix to `OfflinePlayer`'s constructor: instead of only writing the requested `NameTag` when the UUID lookup yields `null`, it now always writes the requested name whenever it is non-null. This fixes the stale-name-on-rename problem where whitelist, op, ban and unban kept operating on the previous account name while the console reported the typed name.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- No regression test covers the renamed-account scenario (UUID lookup with a stale `NameTag`), nor the UUID-only / name-only / null combinations. Such tests would make the intended behavior explicit.
- `getOfflinePlayer(UUID)` still has no requested name and can return a stale `NameTag` after a rename. This is outside the PR's stated name-based scope, but worth noting for future work.
- The three-argument constructor now treats a non-null `name` as authoritative even if a caller supplies a mismatched UUID/name pair. Correct for the internal call site, but explicitly documenting that expectation could prevent misuse.

### Recommendations
- Consider adding tests for the constructor's `(uuid, name)`, `(uuid, null)` and `(null, null)` cases.
- Optionally document the authoritative-name contract on the constructor.

### Notes
- `git diff --check` passes; build verification was blocked by the environment's read-only Maven dependency cache, not by an observed source error.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #861, opened 2026-08-26
fix(network): never block the address of a player who already logged in

w1zardz:
A single undecodable batch from a session kicked the player and then blocked his address for sixty
seconds. The block is meant for floods that never become players, so it now stops at `LOGGED_IN`:
a session that passed login, resource packs and spawn is a real player, and one bad batch there is
a bug, not an attack. He reconnects immediately instead of waiting out a ban he was never told
about.

The prefixed-batch warning also carries the compression prefix byte and the negotiated inbound
compression now, so the decode failure itself can be chased on the default debug level.

### Tests

`RakNetPlayerSessionTest`. Green on current master.

### DIFF
--- a/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
+++ b/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
@@ -577,8 +577,10 @@ public class RakNetPlayerSession extends SimpleChannelInboundHandler<RakMessage>
                     this.player
             );
             if (!result.success()) {
-                log.warn("[{}] Failed to decode batch packet ({} bytes, prefixed, raknetProtocol={}, legacyFallback={})",
-                        this.playerLabel(), packetBuffer.length, this.channel.config().getProtocolVersion(),
+                log.warn("[{}] Failed to decode batch packet ({} bytes, prefixed, prefix=0x{}, compressionIn={}, raknetProtocol={}, legacyFallback={})",
+                        this.playerLabel(), packetBuffer.length,
+                        packetBuffer.length > 0 ? Integer.toHexString(packetBuffer[0] & 0xFF) : "-",
+                        this.compressionIn, this.channel.config().getProtocolVersion(),
                         this.state.getSecurity().isLegacyInboundGraceWindow());
                 return false;
             }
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR intends to stop address-blocking for sessions that already reached `LOGGED_IN` (real players), while preserving flood protection for pre-login sessions, and adds the compression prefix byte plus negotiated inbound compression to prefixed-batch warnings. The design goal is reasonable, but the new upper-bound condition has a timing flaw that disables malformed-packet IP blocking for **every** session.

### Issues Found
#### Blocking Issues
- **Malformed-packet IP blocking is effectively disabled for all sessions.** On a decode failure the code calls `disconnect("Sent malformed packet")` (RakNetPlayerSession.java:163) **before** scheduling the delayed block check (lines 164-176). `disconnect()` synchronously sets the login phase to `DISCONNECTED` (line 204). When the delayed task later evaluates `shouldBlockAddressAfterMalformed(phase = DISCONNECTED, address)`, the new upper bound `phase.ordinal() < LOGGED_IN.ordinal()` is `false` (DISCONNECTED = 10, LOGGED_IN = 9), so it always returns false and `blockAddress` is never invoked. The intended 60-second flood protection for sessions between `LOGIN_RECEIVED` and `READY_TO_LOGIN` is silently removed.
  - **Fix**: capture the phase — or the pre-computed `shouldBlock` boolean — *before* calling `disconnect()`, and use that captured value inside the delayed task.
  - The current tests only exercise the helper with explicit phase values (RakNetPlayerSessionTest.java:306-312), bypassing the disconnect→timeout path, so they cannot detect this regression. Add a flow-level test asserting `blockAddress()` is called for `LOGIN_RECEIVED`/`RESOURCE_PACK` and not for `LOGGED_IN`.

#### Non-blocking Issues
- The new Javadoc on `shouldBlockAddressAfterMalformed` claims `LOGGED_IN` means the player "passed … spawn", but that phase is set immediately after sending `StartGamePacket` (Player.java), while `spawned` becomes true later during first-spawn handling. Drop the "passed spawn" claim from the rationale.
- Test name `malformedPreLoginPacketsDoNotTriggerIpBlock` is misleading — it asserts some pre-login phases DO trigger blocking. Consider `blocksMalformedBatchesOnlyDuringLogin`.
- `compressionIn={}` may print an implementation class/hash rather than a human-readable algorithm name (`SNAPPY`/`ZLIB_RAW`); a stable `toString()`/algorithm label would help diagnostics.
- Prefix formatting is inconsistent: values are unpadded lowercase hex, while an empty batch produces `prefix=0x-`. Prefer `0x00`–`0xff` and `prefix=n/a`.

### Recommendations
- Capture the phase before `disconnect()` (blocking fix).
- Add a flow-level regression test in addition to the helper test.
- Tighten the diagnostic formatting consistent with the rest of the project.

### Notes
`git diff --check` passes. The focused Maven test could not complete in the agent sandbox due the read-only dependency cache, but the control-flow ordering above is deterministic from the production code.

---
🤖 AI Review

PR #862, opened 2026-08-26
player: refuse an elytra-less glide instead of kicking for it

w1zardz:
### Problem

A client sends `START_GLIDING` whenever the player double-taps jump, and it does so even when
there are no wings on — the check for elytra lives on the server. With `allow-flight` off the core
answers that request with a kick, so a player whose elytra had just been taken off (by a plugin,
by durability, or by any inventory move) is disconnected for pressing jump twice.

We hit this in combat: a plugin moves the elytra into a backpack when a fight starts, the player
keeps tapping jump to escape, and the third tap throws them off the server with "flying is not
enabled here".

### Fix

Refuse the request instead of kicking. Both call sites already cancel `PlayerToggleGlideEvent`
when the wings are missing, which resyncs the client and leaves `isGliding` false — so a player
who really does rise without wings still meets the ordinary movement check that kicks for flying.
Only the punishment for a legitimate key press is gone.

Compiles on current master; verified in game on 1.20 … 1.26.45 clients.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4141,10 +4141,10 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                     if (chestplate == null || chestplate.getId() != ItemID.ELYTRA || chestplate.getDamage() >= chestplate.getMaxDurability()) {
                         withoutElytra = true;
                     }
-                    if (withoutElytra && !server.getAllowFlight()) {
-                        this.kick(PlayerKickEvent.Reason.FLYING_DISABLED, MSG_FLYING_NOT_ENABLED, true);
-                        return;
-                    }
+                    // A glide request without elytra is refused, never punished: the client asks
+                    // for it whenever the player double taps jump, which happens all the time when
+                    // a plugin takes the elytra off. Cancelling resyncs the client, and a player who
+                    // keeps rising without wings is still caught by the movement check below.
                     PlayerToggleGlideEvent playerToggleGlideEvent = new PlayerToggleGlideEvent(this, true);
                     if (this.riding != null || this.sleeping != null || withoutElytra) {
                         playerToggleGlideEvent.setCancelled(true);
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR correctly removes the immediate kick from both glide-entry paths (auth input `START_GLIDING` around `Player.java:4141` and legacy `ACTION_START_GLIDE` around `Player.java:4454`). Both paths still treat an absent, non-Elytra, or depleted Elytra as `withoutElytra`, pre-cancel `PlayerToggleGlideEvent`, and only call `setGliding(true)` if the event remains uncancelled. Ordinary clients stay connected with `isGliding() == false`; wingless gliding is not granted, and valid-Elytra behavior is unchanged.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- The no-Elytra cancellation can be overridden by a plugin via `setCancelled(false)`. Since the old kick ran before event dispatch, this is a new (plugin-triggerable) way to authorize `setGliding(true)` on a no-flight server. The next tick clears it for a missing/wrong Elytra, so this does not grant sustained flight under normal configurations, but it can open a one-tick window. Plugins are trusted code, so the policy should be documented.
- The existing anti-flight fallback in `onUpdate` (`Player.java:2850`) is delayed and conditional (runs after ~20 air ticks; kicks only after ~150; exact-zero hover packets excluded). This limitation predates the PR, but removing the immediate kick makes it slightly more relevant to anti-cheat coverage.
- No regression tests were added for the packet paths, depleted Elytra, event cancellation, or resulting `isGliding` state.

### Recommendations
- Add tests covering both `START_GLIDING` and `ACTION_START_GLIDE` across: no chestplate, wrong chestplate, depleted Elytra, valid Elytra, and server flight enabled/disabled.
- Decide explicitly whether plugins may override the no-Elytra restriction; if not, re-check `withoutElytra` after event dispatch rather than relying only on the mutable cancellation flag.
- Consider extracting the duplicated Elytra/event logic into a shared helper to keep the two packet paths from diverging.

### Notes
- With `allow-flight` enabled, the old code did not kick for this condition, so behavior there is effectively unchanged.
- The PR also stops emitting `PlayerKickEvent(FLYING_DISABLED)` for accidental glide requests and lets the rest of the auth-input packet be processed.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #863, opened 2026-08-26
fix(level): cap runaway block update cascades and report the load

w1zardz:
### Problem

The normal block update queue is drained with a plain `while (poll() != null)` loop, and an update
performed inside that loop can queue more updates into the very same drain: liquids, falling
blocks, redstone loops and large fills all feed themselves. A cascade that keeps producing work
therefore holds the main thread for as long as it lasts — until the watchdog kills the server —
and nothing in the log says which level was busy or how far behind it fell.

### Fix

Cap how many normal updates one tick may perform (200000 by default,
`-Dnukkit.maxBlockUpdatesPerTick`). Whatever is left stays in the queue and is performed by the
next tick, so no update is dropped and ordinary gameplay — which never reaches the cap — keeps its
exact timing.

Redstone and liquid timing is untouched: the scheduled queue still performs every entry of its
tick, and it cannot loop within one tick because an update scheduled while it runs lands on the
next one.

Report the load when the cap is hit, or when both queues together take longer than 50 ms
(`-Dnukkit.slowBlockUpdateMillis`), at most one line per level per ten seconds, and expose the
counters (normal and scheduled backlog, per-tick counts, cap hits) so a plugin can chart them.
`BlockUpdateScheduler#tick` now returns how many entries it performed.

Compiles on current master; running in production on a server with heavy liquid and fill activity.

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -340,8 +340,35 @@ public class Level implements ChunkManager, Metadatable {
     // Sentinel marking "too many changes, resend the whole chunk"; never populated, so never test it via size()
     private final Int2ObjectOpenHashMap<Object> changeBlocksFullMap = new Int2ObjectOpenHashMap<>();
 
+    /**
+     * Hard ceiling on how many normal block updates one tick is allowed to perform.
+     *
+     * <p>The normal update queue is re-entrant: an update performed while draining it can push
+     * more updates into the very same drain, so a runaway cascade (liquids, falling blocks,
+     * redstone loops, a huge fill) keeps the main thread busy for as long as it produces work -
+     * until the watchdog kills the server and the worlds are saved from whatever state they were
+     * in. The ceiling turns that freeze into a stutter: the remainder simply stays in the queue
+     * and is performed by the next tick, so nothing is lost. Ordinary play never reaches it.
+     */
+    private static final int MAX_NORMAL_UPDATES_PER_TICK =
+            Math.max(1024, Integer.getInteger("nukkit.maxBlockUpdatesPerTick", 200_000));
+
+    /** Block update work slower than this gets a line in the server log. */
+    private static final long SLOW_BLOCK_UPDATE_NANOS =
+            Math.max(1L, Long.getLong("nukkit.slowBlockUpdateMillis", 50L)) * 1_000_000L;
+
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR caps the number of normal block updates performed per tick (default 200,000, min 1,024 via `-Dnukkit.maxBlockUpdatesPerTick`), so a runaway cascade (liquids, falling blocks, redstone loops, large fills) can no longer hold the main thread indefinitely until the watchdog kills the server. Remaining updates stay in the queue and run next tick, so nothing is dropped and ordinary gameplay timing is untouched. It also reports load when the cap is hit or when both queues together take longer than 50 ms (`-Dnukkit.slowBlockUpdateMillis`), rate-limited to one line per level per 10 s, and exposes counters so plugins can chart normal/scheduled backlog, per-tick counts, and cap hits.

Automated review tooling was temporarily unavailable; a manual review of the change was performed.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- The new load counters (`normalBlockUpdatesLastTick`, `scheduledBlockUpdatesLastTick`, `blockUpdateNanosLastTick`, `blockUpdateFuseTrips`, `lastBlockUpdateReport`) are plain non-`volatile` fields written only on the main thread but read through the public getters. A plugin reading them from another thread may observe stale values. These are informational metrics only, so this is cosmetic; making them `volatile` would make cross-thread metric reads more predictable.

### Recommendations
- The backlog counter implementation is correct: every active enqueue goes through `queueNormalUpdate` (increment), and the drain loop decrements on each successful poll, so the tracked size stays consistent with the `ConcurrentLinkedDeque`. The commented-out body inside the `@Deprecated updateAround(int, int, int, int)` method still calls `normalUpdateQueue.add` directly, but it is inert — fine to leave, or clean up.
- No hard-coded values beyond the two documented system properties; the `Math.max(...)` clamps are a sensible guard against misconfiguration.

### Notes
- Verified against the PR head commit (`e6b14c60a`): active references all route through `queueNormalUpdate`, and ClamAV scan of the two changed files returned clean. No binary/archive files or suspicious patterns (command execution, network calls, reflection abuse) present in the diff.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

PR #864, opened 2026-08-26
fix(entity): keep skin NPCs visible on modern Bedrock clients

w1zardz:
### Problem

A non-player `EntityHuman` (statue, world bot, clone) is announced with a temporary player-list
entry, spawned with `AddPlayerPacket` and then removed from the list again. Two parts of that
dance break on recent international clients:

1. **Removing the temporary entry in the same tick as `AddPlayer` disconnects them.** NetEase
   clients already went through a guarded five-tick removal path; international ones did not.
2. **A full persona skin inside the ADD entry is rejected by 1.26.x**, and a rejected entry leaves
   `AddPlayerPacket` with no identity to draw — the doll disappears entirely instead of standing
   with the default look.

### Fix

Every non-player `EntityHuman` takes the delayed removal path (real `Player` skin handling is
unchanged), the ADD entry carries the blank placeholder skin for every client, and the real skin
is sent right **after** `AddPlayerPacket` — the point where the client has an entity to repaint.
A `PlayerSkinPacket` that arrives before the spawn is dropped by the client, which is why the
order matters.

Compiles on current master; verified in game on 1.20 … 1.26.45: NPCs are visible and wear their
own skin on both halves of that range.

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityHuman.java
+++ b/src/main/java/cn/nukkit/entity/EntityHuman.java
@@ -322,8 +322,7 @@ public class EntityHuman extends EntityHumanType {
                 throw new IllegalStateException(this.getClass().getSimpleName() + " must have a valid skin set");
             }
 
-            boolean retainNpcListEntry = !(this instanceof Player)
-                    && PlayerEntitySkinSender.requiresRetainedEntry(player);
+            boolean retainNpcListEntry = !(this instanceof Player);
             if (this instanceof Player) {
                 if (!player.sentSkins.contains(this.uuid)) {
                     this.server.updatePlayerListData(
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The change unifies every skin-bearing non-player `EntityHuman` onto the delayed PlayerList-entry removal path for all clients, sending a blank placeholder skin in the ADD entry and the real `PlayerSkinPacket` only after `AddPlayerPacket`. This matches the documented ordering requirement (a skin packet sent before spawn is dropped by international clients) and is verified across 1.20 – 1.26.45. Logic is coherent and the core timing is correct.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `PlayerEntitySkinSender.sendSkinnedEntryIfAbsent(...)` (added in this PR) is **dead code** — it is never called. `EntityHuman.spawnTo` uses the existing `sendInitialSkinIfAbsent(...)`, not this new method. Remove it or wire it in to avoid an unused, confusingly parallel API.
- `sendInitialSkinIfAbsent` still emits its own `PlayerSkinPacket` before `AddPlayerPacket`, and the new `sendSkinAfterSpawn` emits a second one afterward. For international clients the pre-spawn packet is dropped (harmless redundancy); for NetEase V860 clients the extra post-spawn skin packet is new behavior — please confirm it does not trip the V860 "repeat skin-confirmation hides the entity" rule described in `prepareConfirmSkin`.

### Recommendations
- Drop the unused `sendSkinnedEntryIfAbsent` method (or replace the `sendInitialSkinIfAbsent` call with it and stop double-sending the skin).
- Add a short test or at least a comment documenting the intended single-skin-packet flow per client type.

### Notes
- Codex review was unavailable this run (upstream API quota exhausted); this review was performed manually against `git diff master...HEAD`.
- No `Runtime.exec`, dynamic class loading, external network, or archive/binary payloads introduced — no security concern.

---
🤖 AI Review

@lt-name ready to merge

tachibana-kanade-angel (approved):

PR #865, opened 2026-08-26
build: make the packaged jar reproducible

w1zardz:
Two builds from the same commit produced different jars: archive entries were stamped with the
moment of packaging, and the build time and host name were written into `git.properties` inside
the jar. The contents were byte-for-byte identical, but any hash-based check (a deployment gate,
a mirror, a cache) saw a new artifact every time.

Set `project.build.outputTimestamp` and keep the volatile fields out of `git.properties`, so the
same source produces the same jar.

Compiles on current master; two consecutive builds now hash identically.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,10 +46,8 @@
         <block-state-updater.version>1.21.110-20250915.183018-1</block-state-updater.version>
         <netty.version>4.2.16.Final</netty.version>
         <maven.shade.skip>false</maven.shade.skip>
-        <!-- Воспроизводимая сборка: без фиксированного времени тот же исходник даёт другой jar,
-             потому что записям архива проставляется момент сборки. Мы сверяем ядро по SHA-256,
-             и плавающий хеш останавливал выкатку на каждой пересборке. -->
-        <project.build.outputTimestamp>2026-08-17T00:00:00Z</project.build.outputTimestamp>
+        <!-- Reproducible build: entry timestamps follow the last commit time -->
+        <project.build.outputTimestamp>${git.commit.time}</project.build.outputTimestamp>
     </properties>
 
     <distributionManagement>
@@ -671,9 +669,9 @@
                 <version>3.6.2</version>
                 <dependencies>
                     <dependency>
-                        <groupId>com.github.edwgiz</groupId>
-                        <artifactId>maven-shade-plugin.log4j2-cachefile-transformer</artifactId>
-                        <version>2.8.1</version>
+                        <groupId>org.apache.logging.log4j</groupId>
+                        <artifactId>log4j-transform-maven-shade-plugin-extensions</artifactId>
+                        <version>0.2.0</version>
                     </dependency>
                 </dependencies>
                 <executions>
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. Pinning `project.build.outputTimestamp` and stripping the two most volatile `git.properties` fields makes the packaged jar deterministic for a given build environment — the same source built on the same machine now produces a byte-identical jar, which is exactly the use case described (hash-based deployment gate / mirror / cache seeing a new artifact every rebuild).

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- **Cross-builder reproducibility is incomplete.** The exclude list removes `git.build.time` and `git.build.host`, but `git.properties` still embeds elsewhere-variable fields: `git.build.user.name` / `git.build.user.email` (the OS user running the build), `git.remote.origin.url`, `git.dirty`, and `git.local.branch.ahead`/`behind`. These are constant within a single CI box (so the PR's stated same-environment goal is met) but still differ from builder to builder, so two *different* machines building the same commit still produce different jars. Note the pre-existing `excludeProperty>git.user.*` does **not** match `git.build.user.*`.
- **Log4j2 cachefile transformer not exercised.** The shaded jar uses the Log4j2 `cachefile-transformer`. It is the usual source of residual variance in shaded plugins, so verify empirically with two clean `mvn package` runs that its cache entry honors `outputTimestamp`.
- **Comment language.** The new comments in `pom.xml` are in Russian while the rest of the project's comments are English; consider aligning for maintainability.

### Recommendations
- For full cross-builder reproducibility, replace the blacklist with an `includeOnlyProperties` allowlist restricted to `git.commit.id.abbrev` and `git.branch` — the only two fields the code actually reads via `Nukkit.GIT_INFO`. Removing the blacklist-only approach also drops builder identity, dirty state, and remote-URL variance in one move.
- Run two `mvn package` runs on distinct machines (or at least verify two clean local builds) and diff the two jars' SHA-256 before relying on the reproducibility guarantee.
- `project.build.outputTimestamp=2026-08-17T00:00:00Z` is arbitrary but valid; adding a one-line note about why that date was chosen would help future maintainers.

### Notes
- Repo code only reads `git.commit.id.abbrev` and `git.branch` from `git.properties` (`Nukkit.java`), so restricting the allowlist to those two fields is safe and yields the strictest reproducibility with no runtime impact.
- ClamAV scan of the diff: clean.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

pom.xml
```diff
@@ -46,6 +46,10 @@
         <block-state-updater.version>1.21.110-20250915.183018-1</block-state-updater.version>
         <netty.version>4.2.16.Final</netty.version>
         <maven.shade.skip>false</maven.shade.skip>
+        <!-- Воспроизводимая сборка: без фиксированного времени тот же исходник даёт другой jar,
+             потому что записям архива проставляется момент сборки. Мы сверяем ядро по SHA-256,
+             и плавающий хеш останавливал выкатку на каждой пересборке. -->
+        <project.build.outputTimestamp>2026-08-17T00:00:00Z</project.build.outputTimestamp>
```
lt-name: I think it should be automatically set to the git commit time.

lt-name: Also, please use English comments.

lt-name (changes requested):

tachibana-kanade-angel:

PR #866, opened 2026-08-26
fix(entity): let custom entities work without experiment mode

w1zardz:
### Problem

Registering a custom entity requires `enable-experiment-mode: true`, and that switch does much
more than custom entities: with it on, the server mixes the enabled experiment toggles into
`ResourcePackStackPacket`. The client then re-validates the whole pack stack against them and
rejects packs that do not declare the same experiments — fatal for a server whose resource pack is
mandatory. So the switch has to stay off, and with it off custom entities cannot exist at all.

### Fix

Custom entities need no client-side experiment: the client draws them from the resource pack
alone, as long as the identifier reaches it in `AvailableEntityIdentifiers`. Gate both places on
the thing that actually matters — whether any custom entity is registered — and leave the
experiment toggles out of the pack stack when none are.

The identifiers NBT is also cached per protocol instead of being rebuilt for every login.

Compiles on current master; running in production with custom entities and a mandatory pack on
1.20 … 1.26.45.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -1420,7 +1420,7 @@ public abstract class Entity extends Location implements Metadatable {
         }
     }
 
-    private static int correctEntityIdentifiersProtocol(int protocolId) {
+    public static int correctEntityIdentifiersProtocol(int protocolId) {
         if (protocolId >= ProtocolInfo.v1_19_80) {
             return ProtocolInfo.v1_19_80;
         } else if (protocolId >= ProtocolInfo.v1_19_20) {
--- a/src/main/java/cn/nukkit/entity/custom/EntityManager.java
+++ b/src/main/java/cn/nukkit/entity/custom/EntityManager.java
@@ -80,7 +80,7 @@ public class EntityManager {
                     new ByteArrayInputStream(Entity.getEntityIdentifiersCache(protocol)));
             ListTag<CompoundTag> listTag = compoundTag.getList("idlist", CompoundTag.class);
             for (EntityDefinition entityDefinition : this.identifierToDefinition.values()) {
-                listTag.add(protocol <= 407 ? entityDefinition.getNetworkTagOld() : entityDefinition.getNetworkTag());
+                listTag.add(protocol <= ProtocolInfo.v1_16_0 ? entityDefinition.getNetworkTagOld() : entityDefinition.getNetworkTag());
             }
             compoundTag.putList(listTag);
             return NBTIO.writeNetwork(compoundTag);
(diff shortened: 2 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR removes the requirement that the server must have experiment mode enabled for custom entities to be registered and sent to clients. It refactors `EntityManager`'s network-tag caching from two fixed fields (`networkTagCached` / `networkTagCachedOld`) into a single per-protocol `Int2ObjectMap<byte[]> networkTagCache`, and `AvailableEntityIdentifiersPacket.encode()` now sends the per-protocol tag whenever custom entities exist.

The refactor is clean and behavior is improved: instead of always reading the fixed 582-protocol vanilla tag (`AvailableEntityIdentifiersPacket.TAG`), `createNetworkTag(protocol)` now reads `Entity.getEntityIdentifiersCache(protocol)`, which normalizes via `correctEntityIdentifiersProtocol` — so multiversion clients receive a Vanilla entity list matching their own protocol plus the appended custom entities. The `protocol <= v1_16_0 ? 407 : protocol` split preserves the previous old/new network-tag behavior. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- `getNetworkTagCached(int protocol)` keys the cache by the **raw** protocol argument, while the generated bytes depend on `correctEntityIdentifiersProtocol(protocol)`. Different raw protocols that normalize to the same bucket (e.g. 389, 390, 407 → 340) each get their own cache entry with identical contents — minor memory redundancy, correct behavior. Consider caching by the normalized protocol instead.
- The literal `407` in `AvailableEntityIdentifiersPacket.encode()` equals `ProtocolInfo.v1_16_0`; a named constant reference would be slightly clearer but is not required.

### Recommendations
- Optionally key `networkTagCache` by `correctEntityIdentifiersProtocol(protocol)` to avoid duplicate cache entries across protocol variants.
- Consider a test asserting that a custom entity's network tag is present in the per-protocol cached list for both a legacy (≤407) and a current protocol.

### Notes
- Codex review was unavailable this run due to an upstream API rate limit (403 insufficient balance); the review above is a manual verification of the full diff and its call chains (`Entity.getEntityIdentifiersCache`, `correctEntityIdentifiersProtocol`, `EntityDefinition` network-tag selection).
- The removed `enableExperimentMode` gate in `registerDefinition` is scoped to entities only; custom-*item* registration (`RuntimeItemMapping`, `Item`) still requires experiment mode, which matches the PR's stated purpose.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #869, opened 2026-08-26
network: preserve the transport disconnect reason

w1zardz:
RakNet already knows whether the remote peer closed cleanly, the session timed out or a bad packet
killed it, but `RakNetPlayerSession#channelInactive` discarded that event and reported every
channel close with one generic string (`"Disconnected from Server" // TODO: timeout reason`).

Keep the typed `RakDisconnectReason` from `userEventTriggered` and pass a stable reason into
`PlayerQuitEvent`, so a plugin — and the server log — can tell an explicit client exit
(`transport:closed_by_remote_peer`) from an interrupted connection (`transport:timed_out`).

That distinction is what let us track down a client-side packet parse failure: the client closes
the socket a fraction of a second after the offending packet, and with the generic string those
disconnects were indistinguishable from ordinary quits.

Compiles on current master; running in production.

### DIFF
--- a/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
+++ b/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
@@ -30,6 +30,7 @@ import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import lombok.extern.log4j.Log4j2;
 import org.apache.logging.log4j.message.FormattedMessage;
 import org.cloudburstmc.netty.channel.raknet.RakChildChannel;
+import org.cloudburstmc.netty.channel.raknet.RakDisconnectReason;
 import org.cloudburstmc.netty.channel.raknet.packet.RakMessage;
 import org.cloudburstmc.netty.handler.codec.raknet.common.RakSessionCodec;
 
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR preserves the typed RakNet transport disconnect reason instead of reporting every channel close with a generic "Disconnected from Server" string. It correctly captures `RakDisconnectReason` via `userEventTriggered` and maps it to stable `transport:*` strings so plugins and the server log can distinguish an explicit client exit from an interrupted connection.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- The `switch` over `RakDisconnectReason` is exhaustive (no `default`), which is good — the compile-time exhaustiveness check guarantees new enum values are handled rather than silently falling back to "transport:unknown".

### Notes
- The new `transportDisconnectReason` field is read and written within the Netty event loop thread (`userEventTriggered` and `channelInactive` run on the same channel event loop), so there is no cross-thread visibility concern.
- The `transport:unknown` fallback provides a safe default when the reason was never captured.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #870, opened 2026-08-26
fix(player): let a server keep the session that is already playing

w1zardz:
On a duplicate login the core closes the session that is already in the world and lets the
newcomer in. That hands the world to whoever knocks last: a player carrying loot loses it to a
stranger reusing the nickname, and the owner of the nickname cannot tell an attack from a glitch.

Add `keep-existing-session-on-duplicate-login` to `server.properties`, **false by default**, so
nothing changes for existing servers. With it on the newcomer is refused instead: the one who is
playing keeps playing, and a stale session still times out by itself.

Compiles on current master; running in production with the option on.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3157,6 +3157,14 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         for (Player p : new ArrayList<>(this.server.playerList.values())) {
             if (p != this && p.username != null) {
                 if (p.username.equalsIgnoreCase(this.username) || this.getUniqueId().equals(p.getUniqueId())) {
+                    if (this.server.isDuplicateLoginKeepingExistingSession()) {
+                        // The session that is already in the world stays, and the newcomer is the
+                        // one that is refused: closing the older session hands the world to
+                        // whoever knocks last, so a player carrying loot loses it to a stranger
+                        // reusing the nickname. A stale session still times out by itself.
+                        this.close("", "disconnectionScreen.loggedinOtherLocation");
+                        return;
+                    }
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR adds a new server property `keep-existing-session-on-duplicate-login` (off by default) that changes duplicate-login behaviour: with it enabled, the newcomer is refused and the already-connected session keeps playing; without it, the historical behaviour (close the existing session, let the newcomer in) is preserved.

*Note: Codex was unavailable during this review (API account balance insufficient); a manual review was performed based on the PR diff.*

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- In `Player.processLogin`, the newcomer is closed with reason `disconnectionScreen.loggedinOtherLocation`. This message ("logged in from another location") is slightly inaccurate for the actual scenario (refused because a session with the same name is already playing). Consider a dedicated disconnect screen string for clarity.

### Recommendations
- The getter name `isDuplicateLoginKeepingExistingSession()` is accurate but verbose; a shorter name such as `keepExistingSessionOnDuplicateLogin()` would read more naturally. Not required.
- Consider documenting the new property in the example/config documentation so server admins are aware of the option.

### Notes
- The `return` after closing the newcomer in `processLogin` correctly stops the duplicate-login handling so the existing session is preserved, and the newcomer's stale session will still time out by itself.
- Default `false` means no behaviour change for existing servers. The unique-id match path (`this.getUniqueId().equals(p.getUniqueId())`) is handled by the same branch, which is appropriate.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

PR #874, opened 2026-08-27
perf(skin): cache Skin content fingerprint

Suoerix:
Add a volatile cachedFingerprint to Skin and lazily compute the SHA-256 content fingerprint in getContentFingerprint. Invalidate the cache in setters that affect the fingerprint (skinId, skinData, skinResourcePatch, capeId, capeData and geometryData) so changes recompute correctly. This avoids repeated expensive hashing of large skin payloads on hot paths and reduces main-thread stalls.

### DIFF
--- a/src/main/java/cn/nukkit/entity/data/Skin.java
+++ b/src/main/java/cn/nukkit/entity/data/Skin.java
@@ -88,6 +88,9 @@ public class Skin {
     @OnlyNetEase
     private byte[] bloomData;
 
+    /** 内容指纹缓存（getContentFingerprint 惰性计算；影响摘要的 setter 负责清除）。 */
+    private volatile String cachedFingerprint;
+
     public boolean isValid() {
         return isValid(Server.getInstance().doNotLimitSkinGeometry);
     }
@@ -150,6 +153,7 @@ public class Skin {
             return;
         }
         this.skinId = skinId;
+        this.cachedFingerprint = null;
     }
 
     public void generateSkinId(String name) {
@@ -168,14 +172,17 @@ public class Skin {
     public void setSkinData(SerializedImage skinData) {
         Objects.requireNonNull(skinData, "skinData");
         this.skinData = skinData;
+        this.cachedFingerprint = null;
     }
 
     public void setSkinResourcePatch(String skinResourcePatch) {
         if (skinResourcePatch == null || skinResourcePatch.trim().isEmpty()) {
             this.skinResourcePatch = GEOMETRY_CUSTOM;
+            this.cachedFingerprint = null;
             return;
         }
         this.skinResourcePatch = skinResourcePatch;
+        this.cachedFingerprint = null;
     }
 
     public void setGeometryName(String geometryName) {
@@ -215,6 +222,7 @@ public class Skin {
             capeId = null;
         }
         this.capeId = capeId;
+        this.cachedFingerprint = null;
     }
 
     public void setCapeData(byte[] capeData) {
@@ -231,6 +239,7 @@ public class Skin {
     public void setCapeData(SerializedImage capeData) {
         Objects.requireNonNull(capeData, "capeData");
         this.capeData = capeData;
+        this.cachedFingerprint = null;
     }
 
     public String getGeometryData() {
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR adds a `volatile cachedFingerprint` to `Skin` and lazily computes the SHA-256 content fingerprint in `getContentFingerprint()`, invalidating the cache across the six serialized-image setters. This is a sound optimization that avoids repeated expensive hashing of large skin payloads on hot paths. The six named setters and their `byte[]`/`BufferedImage` overloads correctly delegate invalidation. However, two mutation paths that affect hashed fields bypass invalidation, so the cache can return a stale fingerprint.

### Issues Found

#### Blocking Issues
- `setGeometryName(String)` (Skin.java ~188) directly reassigns `skinResourcePatch`, one of the hashed inputs, without clearing `cachedFingerprint`. A fingerprint warmed before a `setGeometryName(...)` call stays stale indefinitely.
- `generateSkinId(String)` (Skin.java ~159) directly mutates the hashed `skinId` field without clearing the cache. The lazy call from `getSkinId()` is safe during initial computation, but a public call after the cache is warmed leaves it stale.

#### Non-blocking Issues
- `SerializedImage.data` is a public `final byte[]`; setters retain caller-owned arrays and getters return the same instance. Mutating the underlying bytes after the cache is warmed bypasses invalidation. Defensive copying would close this; low likelihood in practice since skins are usually swapped via setters.
- `getContentFingerprint()` reads the volatile field twice. Between a mutation and a cached get, a lost-invalidation race could return a stale fingerprint if `Skin` is accessed from multiple threads; if it is only touched from the server main thread, this stays theoretical.

### Recommendations
- Route `setGeometryName` through `setSkinResourcePatch`, and `generateSkinId` through `setSkinId`, so both inherit cache invalidation (or clear the cache directly in each).
- Read the volatile into a local variable once in `getContentFingerprint()`.
- Consider defensively copying image data in `SerializedImage`, or centralize cache invalidation in a single helper.

### Notes
- `setSkinData(byte[])`/`setSkinData(BufferedImage)`, `setCapeData(byte[])`/`setCapeData(BufferedImage)` all delegate to the serialized-image setters and inherit invalidation; the unsupported cape byte-array size path is a no-op and needs no invalidation.
- Static review; no compilation or runtime tests performed.

---
🤖 AI Review

Suoerix:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR adds a `volatile String cachedFingerprint` to `Skin` and lazily computes the SHA-256 content fingerprint, clearing the cache in every setter that touches a hashed field. This addresses the hot-path cost of repeatedly hashing large 4D skins (hundreds of KB) on the main thread. A follow-up fix commit now routes every hashed-field mutation through cache-clearing code.

### Issues Found
#### Blocking Issues
- None.

Previous review flagged 2 blocking issues; both are confirmed resolved at the current head:
- `generateSkinId` now clears `cachedFingerprint` after updating `skinId` (Skin.java:161).
- `setGeometryName` now delegates to `setSkinResourcePatch`, so `skinResourcePatch` changes invalidate the cache (Skin.java:196).
- All direct assignments to the hashed fields (`skinId`, `skinResourcePatch`, `skinData`, `capeData`, `capeId`, `geometryData`) are covered by cache invalidation. No bypassing assignment remains.

Focused skin lifecycle tests pass: 29/29.

#### Non-blocking Issues
- `SerializedImage.data` remains a publicly exposed mutable array. In-place mutation after fingerprint calculation would bypass cache invalidation and return a stale fingerprint. The PR documents this as a usage contract, so this is accepted as intentional (skins are normally swapped via setters).

### Recommendations
- Consider adding focused tests proving fingerprint invalidation through every hashed-field setter, especially `setGeometryName` and lazy `getSkinId()`/`generateSkinId()`.
- If `Skin` may be mutated concurrently with fingerprint reads, the current `volatile`-only lazy cache could republish an older computation after a setter invalidates the cache; versioning or synchronization would close that gap.

### Notes
- The cache contract (assume hashed bytes are not mutated in place; mutate via setters) is clearly documented in the Javadoc.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name (approved):

tachibana-kanade-angel:

PR #875, opened 2026-08-28
feat: match vanilla boat health

Kanelucky:
https://www.minecraft.net/en-us/article/minecraft-bedrock-edition-26-40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ server.log
 server.properties
 white-list.txt
 nukkit-mot.yml
+scoreboard.json
 cache
 cache/*
 worlds
--- a/src/main/java/cn/nukkit/entity/item/EntityBoat.java
+++ b/src/main/java/cn/nukkit/entity/item/EntityBoat.java
@@ -65,8 +65,8 @@ public class EntityBoat extends EntityVehicle {
     public EntityBoat(FullChunk chunk, CompoundTag nbt) {
         super(chunk, nbt);
 
-        this.setMaxHealth(40);
-        this.setHealth(40);
+        this.setMaxHealth(4);
+        this.setHealth(4);
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The change corrects the boat's max/current health from 40 to 4 to match vanilla. However, this goal is not fully achieved because the existing damage scaling in `attack()` prevents the boat from actually surviving 4 hit points. An unintended zero-byte `scoreboard.json` runtime file is also accidentally included in the PR.

### Issues Found
#### Blocking Issues
- **Effective health is still not vanilla (4 → effectively 2)** — [`EntityBoat.java`](/MemoriesOfTime/Nukkit-MOT/blob/03b5bb532/src/main/java/cn/nukkit/entity/item/EntityBoat.java#L116-L130). `attack()` doubles incoming damage (`source.setDamage(source.getDamage() * 2)`). With max health now 4, a bare-hand hit (1 damage) deals 2, so the boat breaks in two hits instead of four. To truly "match vanilla boat health", the `* 2` multiplier should be removed, or the change needs vanilla hit-count justification.
- **Unrelated empty file added** — `scoreboard.json` is a zero-byte runtime data file produced by `JSONScoreboardStorage` (not valid source or JSON). It is unrelated to boat health and should be removed from the PR; consider adding it to `.gitignore`.

#### Non-blocking Issues
- None.

### Recommendations
- Remove the `scoreboard.json` file from the commit.
- Add `scoreboard.json` to `.gitignore`.
- Evaluate whether the `* 2` damage multiplier should be removed so the boat actually has 4 HP as in vanilla.

### Notes
- Boat base health 4 matches vanilla.
- No compile or tests run, as requested.

---
🤖 AI Review

Kanelucky:
i have already tested it in game. This AI review is wrong

tachibana-kanade-angel:

PR #876, opened 2026-08-28
feat: EntityFallEvent

Kanelucky:
Credit: [Lumi](https://github.com/KoshakMineDEV/Lumi/commit/55e3353c3b518f67a9fb9ce2f4bcc262e04e297f)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ server.log
 server.properties
 white-list.txt
 nukkit-mot.yml
+scoreboard.json
 cache
 cache/*
 worlds
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,6 @@ server.log
 server.properties
 white-list.txt
 nukkit-mot.yml
-scoreboard.json
 cache
 cache/*
 worlds
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2556,6 +2556,12 @@ public abstract class Entity extends Location implements Metadatable {
                 Block down = this.level.getBlock(this.chunk, this.getFloorX(), this.getFloorY() - 1, this.getFloorZ(), 0, true);
                 int floor = down.getId();
 
+                EntityFallEvent event = new EntityFallEvent(this, down, fallDistance);
+                if (event.isCancelled()) {
+                    return;
+                }
+                fallDistance = event.getFallDistance();
+
                 if (!this.noFallDamage) {
                     float damage = (float) Math.floor(fallDistance - 3 - (this.hasEffect(Effect.JUMP) ? this.getEffect(Effect.JUMP).getAmplifier() + 1 : 0));
 
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2557,6 +2557,7 @@ public abstract class Entity extends Location implements Metadatable {
                 int floor = down.getId();
 
                 EntityFallEvent event = new EntityFallEvent(this, down, fallDistance);
+                this.server.getPluginManager().callEvent(event);
                 if (event.isCancelled()) {
                     return;
                 }
--- a/src/main/java/cn/nukkit/event/entity/EntityFallEvent.java
+++ b/src/main/java/cn/nukkit/event/entity/EntityFallEvent.java
@@ -1,0 +1,31 @@
+package cn.nukkit.event.entity;
+
+import cn.nukkit.block.Block;
+import cn.nukkit.entity.Entity;
+import cn.nukkit.event.Cancellable;
+
+/**
+ * Called when an entity falls.
+ */
+public class EntityFallEvent extends EntityEvent implements Cancellable {
+    private Block blockFallOn;
+    private float fallDistance;
+
+    public EntityFallEvent(Entity entity, Block blockFallOn, float fallDistance) {
+        this.entity = entity;
+        this.blockFallOn = blockFallOn;
+        this.fallDistance = fallDistance;
+    }
+
+    public Block getBlockFallOn() {
+        return blockFallOn;
+    }
+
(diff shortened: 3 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds a new `EntityFallEvent` event that is created during fall handling in `Entity.fall()`. The intent is to let plugins cancel fall damage and/or modify the fall distance before it is converted into damage. The overall structure is consistent with existing entity events (`extends EntityEvent`, `implements Cancellable`), but the event is currently **non-functional** — it is constructed but never dispatched, and the event class is missing the mandatory `HandlerList`. Two blocking issues must be fixed before merge.

### Issues Found
#### Blocking Issues
- **[Entity.java:2559] The event is never dispatched.** `EntityFallEvent` is constructed and the cancellation flag is checked immediately, but there is no `callEvent(event)` / `event.call()` call. As a result, listeners never receive the event, `event.isCancelled()` is always `false`, and `setFallDistance()` can never affect normal execution. Dispatch the event before the cancellation check:
  ```java
  EntityFallEvent event = new EntityFallEvent(this, down, fallDistance);
  this.server.getPluginManager().callEvent(event);
  if (event.isCancelled()) {
      return;
  }
  fallDistance = event.getFallDistance();
  ```
- **[EntityFallEvent.java] Missing static `HandlerList`/`getHandlers()`.** `PluginManager` resolves an event's handlers reflectively via `getDeclaredMethod("getHandlers")` (see `PluginManager.java:613`). Without the standard static block, listener registration/lookup fails even after dispatch is added. Add:
  ```java
  private static final HandlerList handlers = new HandlerList();

  public static HandlerList getHandlers() {
      return handlers;
  }
  ```

#### Non-blocking Issues
- **Cancellation also suppresses farmland trampling.** The early `return` skips the later farmland-trampling branch and its interaction event. Decide whether cancellation means "cancel all landing effects" or only "cancel fall damage", and document accordingly.
- **Javadoc is broader than actual behavior.** The event is only fired for falls where `fallDistance > 0.75`, non-water landings, and entities without slow falling. "Called when an entity lands after a qualifying fall" would be more accurate than "Called when an entity falls".
- **`fallDistance` setter doesn't update the public `Entity.fallDistance` field.** After dispatch is fixed, changing the distance via the event alters this landing's damage, but leaves `Entity.fallDistance` untouched. Clarify the intended contract.
- **No validation on the setter.** `setFallDistance()` permits negative, `NaN`, and infinite values; consider validating or documenting.
- **Style:** `blockFallOn` would be better `final`; `getBlockFallOn()` is a slightly awkward name — `getBlock()` or `getBlockFallenOn()` reads more naturally.

### Recommendations
- Add the missing `callEvent(event)` dispatch and the static `HandlerList`/`getHandlers()` block (both blocking).
- Clarify the cancellation semantics (all landing effects vs. fall damage only).
- Consider adding a listener-registration test covering dispatch, cancellation, reset behavior, and distance mutation — such a test would have caught both blocking problems.

### Notes
- Nukkit requires every concrete directly-registered event to declare its own static `HandlerList` and `getHandlers()`. `EntityFallEvent` currently lacks both, so the event is inoperative until fixed.

---
🤖 AI Review

Kanelucky:
ready to review

lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The follow-up commit fixes both previously-reported blocking issues: the event is now dispatched via `callEvent()` (Entity.java:2560) and `EntityFallEvent` now declares its static `HandlerList`/`getHandlers()`. The dispatch/cancellation/fall-distance flow is now functional. However, a new blocking correctness issue remains regarding the block reported by `getBlockFallOn()`.

### Issues Found
#### Blocking Issues
- **`getBlockFallOn()` reports the block BELOW the actual landing surface for partial-height blocks** (Entity.java:2556). The event is constructed with `down = level.getBlock(..., getFloorY() - 1, ...)`. For full-height blocks this is correct, but when the entity lands on a bottom slab, bed, or carpet (which occupy the `getFloorY()` layer), the reported block is the one beneath the real landing surface. The existing fall-damage branch already accounts for this distinction by checking both the `block` at `getFloorY()` and `floor` at `getFloorY()-1` (e.g. bed handling), so listeners of the event receive misleading data for these common landing surfaces. Fix by resolving the actual landing block (the higher of the two candidate blocks that has a solid face at the entity's feet) before constructing the event.

#### Non-blocking Issues
- **Javadoc is broader than actual behavior.** The event only fires for falls where `fallDistance > 0.75`, non-water landings, and entities without slow falling. "Called when an entity lands after a qualifying fall" would be more accurate than "Called when an entity falls".
- **Cancellation suppresses farmland trampling.** The early `return` on cancel skips the later trampling branch and its interaction event. Decide whether cancellation means "cancel all landing effects" or only "cancel fall damage" and document accordingly.
- **`fallDistance` setter doesn't update the public `Entity.fallDistance` field.** Changing the distance mutates only this landing's damage calc, leaving the entity's public field untouched. Clarify the intended contract.
- **No validation on the setter.** `setFallDistance()` permits negative, `NaN`, and infinite values.
- **Style:** `blockFallOn` could be `final`; `getBlockFallOn()` is an awkward name — `getBlock()` or `getBlockFallenOn()` reads more naturally.
- **No test coverage** for dispatch, cancellation, distance mutation, or partial-height landing blocks. A listener-registration test would have caught the earlier blocking defects.

### Recommendations
- Resolve the actual landing block before constructing the event (blocking).
- Confirm the previous blocking fixes (dispatch + HandlerList) address the event lifecycle correctly with a listener-registration test.
- Add a test covering a landing on a bottom slab / bed / carpet to lock in the landing-block semantics.
- Clarify cancellation semantics and validate the setter input.

### Notes
- Both previously-flagged blocking issues are confirmed resolved; the merge remains blocked by the landing-block accuracy issue above.

---
🤖 AI Review

tachibana-kanade-angel:

PR #877, opened 2026-08-30
fix(entity): keep player death drops in a compact pile

w1zardz:
Player death used Level.dropItem with wideScatter enabled, giving every inventory stack up to 0.5 horizontal velocity. The live PocketMine server uses the regular compact motion (each horizontal component within 0.1), so Nukkit scattered a kill across much more of a PvP arena.

Use regular item motion for player death while preserving the existing 40-tick pickup delay and public pickup rules. A focused test locks both the compact flag and the unchanged delay.

### Tests

`PlayerDeathDropTest` locks the compact motion flag and the unchanged 40-tick pickup delay.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -6824,7 +6824,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             if (!ev.getKeepInventory() && this.level.getGameRules().getBoolean(GameRule.DO_ENTITY_DROPS)) {
                 for (Item item : ev.getDrops()) {
                     if (!item.hasEnchantment(Enchantment.ID_VANISHING_CURSE)) {
-                        this.level.dropItem(this, item, null, true, 40);
+                        this.dropDeathItem(item);
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR makes player death drops use compact item motion (wideScatter=false) instead of the wide radial scatter, so a killed player's inventory stays in a compact pile matching live PocketMine behavior. It refactors the death-drop call into a package-private `dropDeathItem` helper and adds a Mockito unit test locking both the compact flag and the unchanged 40-tick pickup delay.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `PlayerDeathDropTest` tests the helper directly rather than exercising `Player.kill()`. It would stay green even if the death loop were later reverted to the old inline wide-scatter call. A higher-level death-path test or spy would provide stronger regression coverage.

### Recommendations
- Consider a death-path integration test for stronger regression coverage (optional).

### Notes
- `false` is the correct compact-motion selector: it yields independent horizontal components in `[-0.1, 0.1)` with vertical motion `0.2`, matching PocketMine's regular drop behavior (`true` is wide scatter up to ~0.5).
- The five-argument overload `dropItem(this, item, null, false, 40)` preserves the 40-tick pickup delay.
- Death-event cancellation, keep-inventory/game-rule checks, and vanishing-curse filtering are all unchanged.
- Package-private visibility is appropriate — a same-package test seam without expanding the public API.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:
After comparing it with the BDS implementation, I believe the current implementation is consistent with the original BDS design. Therefore, this PM implementation should not be merged into `master` as the default behavior. Please consider implementing the customized behavior through a plugin instead.

PR #878, opened 2026-08-30
block: pour what the bucket in a dispenser actually holds

w1zardz:
A dispenser translated the bucket it fires with getDamageByTarget(), which
maps a BLOCK id to a bucket damage, not the other way round. Fed a bucket
damage it answers with whatever bucket shares that number, so a powder snow
bucket (11) came back as a lava bucket (10): the dispenser poured lava and
kept no snow at all. Water (8) and lava (10) survived only because their
bucket damage and their block id happen to be the same number.

Read the content with getBlockByDamage(), which is the inverse the class
already provides, and place it when it is a liquid or powder snow. A bucket
holding a creature is left to the default behavior: it maps to water too, and
placing that water would swallow the fish, the axolotl or the tadpole inside.

### DIFF
--- a/src/main/java/cn/nukkit/dispenser/BucketDispenseBehavior.java
+++ b/src/main/java/cn/nukkit/dispenser/BucketDispenseBehavior.java
@@ -4,6 +4,7 @@ import cn.nukkit.block.Block;
 import cn.nukkit.block.BlockDispenser;
 import cn.nukkit.block.BlockID;
 import cn.nukkit.block.BlockLiquid;
+import cn.nukkit.block.BlockPowderSnow;
 import cn.nukkit.item.Item;
 import cn.nukkit.item.ItemBucket;
 import cn.nukkit.item.ItemID;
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. The core fix is correct: the dispenser previously used `getDamageByTarget()` (which maps a BLOCK id to a bucket damage) in the wrong direction. Switching to `getBlockByDamage()` properly maps bucket metadata back to block IDs, so a powder-snow bucket (11) now places powder snow instead of incorrectly resolving to a lava bucket (10). Water/lava pour behavior and empty-bucket inventory handling remain intact.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- No regression tests were added. Tests should cover water, lava, powder snow, obstructed targets, and creature buckets.
- Creature buckets (fish, axolotl, tadpole) are deliberately ejected intact rather than placing water and spawning the creature. This preserves existing behavior but is not full vanilla dispenser parity — worth handling in a follow-up.

### Recommendations
- Merge for the powder-snow regression fix.
- Consider adding focused dispenser tests and addressing creature-bucket dispenser behavior in a follow-up.

### Notes
- No runtime, API-breaking, or security issues found.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #879, opened 2026-08-30
feat(inventory): expose requested craft repetitions

w1zardz:
### Problem

`CraftItemEvent` carries the recipe and the output, but not how many times the recipe was actually
run. `CraftRecipeAction` knows the number (`getNumberOfRequestedCrafts`), and the stack request
handler drops it on the floor when it builds the event.

A listener that counts crafts therefore has to guess. Counting the output stack size counts four
sticks as four crafts; hardcoding one undercounts every batch craft the recipe book sends, and the
client sends batches routinely.

### Fix

Keep the requested repetition count on the event. `new CraftItemEvent(transaction)` still answers
`1`, so existing listeners and existing callers are unaffected; the stack request path passes the
real number through.

### Tests

`CakeCraftingReproTest` gains a case that crafts three breads in one request and asserts the event
reports three repetitions while the delivered stack is still three items.

### Note on the base

The first two commits of this branch are #859 — the craft event this change touches is built on
that path. Merge #859 first and this branch rebases down to a single commit.

### DIFF
--- a/src/main/java/cn/nukkit/event/inventory/CraftItemEvent.java
+++ b/src/main/java/cn/nukkit/event/inventory/CraftItemEvent.java
@@ -28,19 +28,27 @@ public class CraftItemEvent extends Event implements Cancellable {
 
     private CraftingTransaction transaction;
 
+    private final int repetitions;
+
     public CraftItemEvent(CraftingTransaction transaction) {
+        this(transaction, 1);
+    }
+
+    public CraftItemEvent(CraftingTransaction transaction, int repetitions) {
         this.transaction = transaction;
         this.player = transaction.getSource();
         this.input = transaction.getInputList().toArray(Item.EMPTY_ARRAY);
         // 取宽类型 transactionRecipe，使 MultiRecipe 等非 CraftingRecipe 能透传到事件
         // Use the wide-typed transactionRecipe so non-CraftingRecipe types (e.g. MultiRecipe) survive
         this.recipe = transaction.getTransactionRecipe();
+        this.repetitions = Math.max(1, repetitions);
     }
 
     public CraftItemEvent(Player player, Item[] input, Recipe recipe) {
         this.player = player;
         this.input = input;
         this.recipe = recipe;
+        this.repetitions = 1;
     }
 
     public CraftingTransaction getTransaction() {
@@ -57,5 +65,10 @@ public class CraftItemEvent extends Event implements Cancellable {
 
     public Player getPlayer() {
         return this.player;
+    }
+
+    /** Number of recipe executions represented by this event, not the output stack size. */
+    public int getRepetitions() {
+        return repetitions;
     }
 }
--- a/src/main/java/cn/nukkit/inventory/request/CraftGrindstoneActionProcessor.java
+++ b/src/main/java/cn/nukkit/inventory/request/CraftGrindstoneActionProcessor.java
@@ -64,6 +64,11 @@ public class CraftGrindstoneActionProcessor implements ItemStackRequestActionPro
         if (event.isCancelled()) {
             return context.error();
         }
+        Item eventOutput = event.getNewItem();
+        if (eventOutput == null || eventOutput.isNull()) {
+            return context.error();
+        }
+        result = eventOutput.clone();
 
         int experienceDropped = event.getExperienceDropped();
         if (experienceDropped > 0) {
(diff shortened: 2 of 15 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR propagates the requested craft repetition count through `CraftItemEvent` and makes the crafting path event-author its output (the stack request handler now passes `numberOfRequestedCrafts` through to the event instead of always answering `1`). The retained single-argument constructors still default to `1`, preserving listener compatibility. The ordinary `CraftRecipeAction` path, plus the smithing, grindstone, repair, enchant, and stonecutter request paths, handle the change correctly.

However, three blocking issues remain — most importantly a breaking regression on automatic (recipe-book / shift-click) crafting of dynamic recipes.

### Issues Found
#### Blocking Issues
1. **Recipe-book auto-crafts never propagate the requested repetition count.** `CraftRecipeAutoProcessor.java:57-58` still calls the one-argument `CraftItemEvent` constructor, which defaults to `1` at `CraftItemEvent.java:33-44`. This ignores both `numberOfRequestedCrafts` and `timesCrafted` available on `AutoCraftRecipeAction`. Batch crafts via the recipe book therefore still report one repetition, and this path also keeps using `recipeResult` instead of the event-authored output (`CraftRecipeAutoProcessor.java:125-131`).

2. **Automatic `MultiRecipe` crafts now always fail at the deprecated-results action (breaking regression).** `CraftResultDeprecatedActionProcessor.java:46-49` unconditionally requires `EVENT_AUTHORED_MULTI_OUTPUT_KEY`, which is only populated by the ordinary path at `CraftRecipeActionProcessor.java:125`. The auto-craft path instead stores `MULTI_RESULTS_KEY` at `CraftRecipeAutoProcessor.java:88-101`. When the request dispatcher later reaches the `CraftResultsDeprecatedAction`, the missing key forces `context.error()` and rolls back the whole request. Auto-crafting dynamic recipes — repairs, fireworks, map/book cloning, banner recipes — is broken.

3. **Legacy event-authored outputs are silently lost when delivery stacks or splits the result.** `InventoryTransaction.applyEventOutputToActions()` only rewrites actions whose complete source or target stack is `equalsExact(originalOutput)` (`InventoryTransaction.java:157-171`). For a result of `bread ×1` delivered into an existing `bread ×10`, the real slot action is `×10 → ×11`; neither side equals `×1`, so the event-added NBT is not applied. Split deliveries hit the same problem. The new legacy tests only create `AIR → exact output` actions (`EventAuthoredLegacyTransactionTest.java:193-195`), so they cannot detect this. Delivery must be rewritten around the output quantity delta/role, or rejected and resynchronized when the event output can no longer merge safely.

#### Non-blocking Issues
- The repetition test does not prove "repetitions rather than output stack size." `CakeCraftingReproTest.java:277-300` crafts three breads producing three items, so both values are `3`; a recipe yielding multiple items per execution would distinguish the semantics.
- Missing coverage for the most regression-prone paths: no `AutoCraftRecipeAction` event/repetition or dynamic-auto test; the dynamic `MultiRecipe` test requests only one craft (`CakeCraftingReproTest.java:318-324`); smithing covers transform recipes only, not trim; legacy transaction tests override `canExecute()` and only deliver to empty slots.
- `getRepetitions()` returns the client-requested, clamped value, not necessarily a completed craft count (the ordinary event fires before recipe/consume-plan validation). The javadoc at `CraftItemEvent.java:70` should say "requested" explicitly, since listeners may use it to multiply irreversible side effects.

### Recommendations
- Fix the auto-craft path (issue #1): pass `timesCrafted` through to the event in `CraftRecipeAutoProcessor` and honor the event-authored output.
- Fix the deprecated-results regression (issue #2) before merge — either populate `EVENT_AUTHORED_MULTI_OUTPUT_KEY` on the auto path or fall back to `MULTI_RESULTS_KEY` when the event auth key is absent.
- Rework `applyEventOutputToActions` to match by output role/count-delta rather than exact stack equality, or reject + resync on unsafe merge.
- Add tests for the auto/recipe-book path and a multi-item-per-execution recipe to actually prove repetition semantics.

### Notes
- Changeset is clean and well-scoped; the ordinary array path and all workstation request paths are correctly implemented.
- Required follow-ups above must land before this is safe to merge, per blocking issues #1–#3.

---
🤖 AI Review

PR #880, opened 2026-08-30
generator: let a server read how many objects a count populator places

w1zardz:
The amount a biome asks for can only be overwritten, never adjusted: there is no
way to read it back. A server that wants, say, half the trees of the vanilla
forest therefore has to hardcode a number per biome and drift away from upstream
the moment those numbers change.

Both fields already exist and are only missing their getters.

### DIFF
--- a/src/main/java/cn/nukkit/level/generator/populator/type/PopulatorCount.java
+++ b/src/main/java/cn/nukkit/level/generator/populator/type/PopulatorCount.java
@@ -23,6 +23,26 @@ public abstract class PopulatorCount extends Populator {
         this.baseAmount = baseAmount;
     }
 
+    /**
+     * How many objects this populator always places in a chunk.
+     *
+     * <p>Without a getter the amount can only ever be overwritten, never adjusted: a server that
+     * wants half the trees of a biome has no way to learn how many that biome asked for.
+     */
+    public final int getBaseAmount() {
+        return this.baseAmount;
+    }
+
+    /**
+     * The exclusive upper bound of the extra amount rolled on top of {@link #getBaseAmount()}.
+     *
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR adds two simple getters (`getBaseAmount()` and `getRandomAmount()`) to `PopulatorCount`, letting servers read how many objects a count populator places. The implementation is trivial and behavior-preserving: `getBaseAmount()` returns the existing field, and `getRandomAmount()` returns the exact exclusive bound handed to `nextBoundedInt()`.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `getRandomAmount()` does not round-trip the argument accepted by `setRandomAmount()`: `setRandomAmount(2)` makes the getter return `3`. This is mathematically correct for the documented exclusive-bound contract, but differs from normal setter/getter expectations and could cause off-by-one errors when a plugin reads, scales, and writes the value back. The Javadoc already makes the behavior explicit, so this is an API-design concern rather than a runtime bug. Consider storing the caller's value and adding one only in `populate()`, or renaming to something like `getRandomAmountBound()`.
- The `getBaseAmount()` Javadoc says the value is how many objects are "always" placed, but it is only the additive baseline; random amounts and existing negative baselines mean the populator may place fewer or no objects. Rephrase for precision.

### Recommendations
- Consider clarifying the `getRandomAmount()` API contract (round-trip semantics) before treating it as long-term public API.
- Minor: rephrase the `getBaseAmount()` Javadoc to avoid implying the value is always placed.

### Notes
- Small, focused change; no tests added, which is acceptable given the trivial scope.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

PR #881, opened 2026-08-30
player: keep pre-login level changes packet-free

w1zardz:
PlayerLoginEvent runs before StartGame, which makes it the only safe hook for
moving a player away from a saved position that crashes their client. A level
change from that hook currently calls Player.afterSwitchLevel and emits spawn,
time, weather and gamerule packets before StartGame. Crossing dimensions also
emits ChangeDimension. Besides being useless, that packet order is invalid for
clients which have not entered the world yet.

Keep the server-side level and position change, but defer all client-facing
level-change packets until the player is actually online. StartGame is built
after PlayerLoginEvent and therefore reads the relocated level and coordinates.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2719,7 +2719,7 @@ public abstract class Entity extends Location implements Metadatable {
         this.level.addEntity(this);
         this.chunk = null;
 
-        if (this instanceof Player) {
+        if (this instanceof Player player && player.isOnline()) {
             this.afterSwitchLevel();
         }
         return true;
@@ -3102,7 +3102,7 @@ public abstract class Entity extends Location implements Metadatable {
                 this.z = pos.z;
 
                 // Dimension change
-                if (this instanceof Player player && newLevel.getDimension() != oldLevel.getDimension()) {
+                if (this instanceof Player player && player.isOnline() && newLevel.getDimension() != oldLevel.getDimension()) {
                     player.setDimension(newLevel.getDimension());
                 }
 


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR guards `afterSwitchLevel()` and `setDimension()` inside `setPosition()`/`switchLevel()` behind `player.isOnline()` so that pre-login (`PlayerLoginEvent`) level changes don't emit client-facing packets (spawn, time, weather, gamerules, ChangeDimension) before `StartGame`. The intent — defer client packets until the player is online — is sound and the PR does eliminate the invalid packet order. However, the implementation skips the **entire** `afterSwitchLevel()` hook rather than only its client-facing packets, dropping server-side state that the hook is also responsible for. This needs to be resolved before merging.

### Issues Found

#### Blocking Issues
- **Skipping `afterSwitchLevel()` loses a real server-side state reset, not just packets.** `Player.afterSwitchLevel()` (Player.java:8061) resets `timeSinceRest = 0` (Player.java:8081) in addition to sending packets. With the new `isOnline()` guard, a player relocated during `PlayerLoginEvent` never gets this reset, so the saved insomnia counter survives login and phantoms can spawn immediately afterward instead of the counter being reset as on every other level switch.
- **The world-spawn/compass packet suppressed here is not regenerated by the login sequence.** `afterSwitchLevel()` sends `SetSpawnPositionPacket` (TYPE_WORLD_SPAWN) for the target level, but `completeLoginSequence()` builds `StartGame` with the player's current coordinates as `spawnX/Y/Z` — not `targetLevel.getSpawnLocation()`. So the target world's actual spawn/compass position may never reach the client.
- **Legacy protocols lose initial weather state.** For protocols using `completeLegacyLoginSequence()`, `rainLevel`/`lightningLevel` are not populated and no later `sendWeather()` runs. A pre-login move into a raining world can lose the initial weather state.

#### Non-blocking Issues
- No style or naming concerns.
- No regression tests cover packet ordering, the final `StartGame` dimension/coordinates, online cross-dimension transitions, or disconnects during a switch. These should be added once the state/packet handling is corrected.

### Recommendations
- Preserve the no-packets-before-`StartGame` behavior, but retain `afterSwitchLevel()`'s server-side state effects (`timeSinceRest` reset).
- Expressly replay the client-facing pieces that `StartGame` does not cover (world-spawn/compass for the target level; weather for legacy protocols) after the player is actually online — either by deferring/queuing them until login completes or by separating the state-reset portion from the packet-sending portion of the hook.
- Add regression tests for the pre-login relocation path.

### Notes
- Server-side level and position mutation still completes; normal online transitions retain their old behavior.
- The race assessment: in the normal Nukkit lifecycle login and level changes run on the server thread, and `dataPacket()` also checks `connected`, so no new runtime exception or security issue is evident.

---
🤖 AI Review

tachibana-kanade-angel:

PR #882, opened 2026-08-30
generator: honor the layer a populator writes into

w1zardz:
SimpleChunkManager.setBlockAtLayer accepted a layer argument and silently
dropped it: every write went to layer 0. A populator that placed a block
and then cleared layer 1 (to strip water) erased its own block instead -
the layer-1 air landed in layer 0 on top of it. No warning was logged;
the block was simply gone before the chunk was ever saved.

Route the write through FullChunk.setBlockAtLayer, which honors layers.

### DIFF
--- a/src/main/java/cn/nukkit/level/generator/SimpleChunkManager.java
+++ b/src/main/java/cn/nukkit/level/generator/SimpleChunkManager.java
@@ -65,7 +65,7 @@ public abstract class SimpleChunkManager implements ChunkManager {
 
         FullChunk chunk = this.getChunk(x >> 4, z >> 4);
         if (chunk != null) {
-            return chunk.setBlock(x & 0xf, y, z & 0xf, id, data);
+            return chunk.setBlockAtLayer(x & 0xf, y, z & 0xf, layer, id, data);
         }
         return false;
     }


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The fix is correct and sound: `SimpleChunkManager.setBlockAtLayer()` now forwards the `layer` argument through to `FullChunk.setBlockAtLayer(...)` instead of silently dropping it to layer 0. This lets populators write into a specific layer (e.g. clearing layer 1 to strip water) without erasing their own blocks in layer 0.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- No regression test covers layer-aware writes, so the dropped-layer bug could silently return.
- No unified validation of illegal layer numbers (negative / > 1). Existing callers only use the valid layer 1, so this does not affect the current PR.

### Recommendations
- Add a regression test: place a block in layer 0 and water in layer 1, then clear layer 1 via `SimpleChunkManager`; assert layer 0 is unchanged and layer 1 is empty, and verify the first modification returns `true` while a repeated one returns `false`.
- Unified illegal-layer validation could be a follow-up improvement.

### Notes
- `FullChunk.setBlockAtLayer(int,int,int,int,int,int)` exists and writes to the requested layer, returning `true` only when the block state actually changes.
- `setBlockAt()` still delegates to layer 0 explicitly — behavior unchanged.
- No callers were found that rely on the old "ignore layer" behavior.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #883, opened 2026-08-30
custom-block: translate legacy render method names for new clients

w1zardz:
Clients since the 1.21.80 material overhaul silently ignore the legacy
"alpha_test" render method name in a network-defined block: the material
falls back to opaque and every transparent texel renders as a black
pixel. No warning appears anywhere - the texture itself still applies,
so the block looks almost right and the defect surfaces only on the
first custom block with actual transparency in its texture.

Translate the render method per protocol in the definition serializer,
next to the existing face_dimming/ambient_occlusion conversions: for
1.21.110+ clients "alpha_test" becomes "alpha_test_single_sided". Old
clients keep the legacy name they understand.

### DIFF
--- a/src/main/java/cn/nukkit/block/custom/serializer/impl/CustomBlockDefinitionSerializer843.java
+++ b/src/main/java/cn/nukkit/block/custom/serializer/impl/CustomBlockDefinitionSerializer843.java
@@ -48,6 +48,14 @@ public class CustomBlockDefinitionSerializer843 extends CustomBlockDefinitionSer
                     material.remove("face_dimming");
                     material.putByte("packed_bools", (byte) (faceDimming ? 0x1 : 0x0));
                 }
+                // 1.21.80+ material overhaul: the legacy "alpha_test" name is silently
+                // ignored by new clients and the material falls back to opaque, so every
+                // transparent texel renders black. Translate to the modern name here;
+                // older clients keep the legacy name they understand.
+                if (material.contains("render_method")
+                        && material.getString("render_method").equals("alpha_test")) {
+                    material.putString("render_method", "alpha_test_single_sided");
+                }
             }
         }
     }


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR adds translation of the legacy `alpha_test` render method name to `alpha_test_single_sided` for newer Bedrock clients (1.21.80+ material overhaul), because new clients silently ignore `alpha_test` and the material falls back to opaque, making transparent texels render black. The translation is correctly scoped (only material entries whose `render_method` is exactly `alpha_test`), and older clients are not regressed. However, the fix only actually runs for protocol 843+ (1.21.110), not for the full 1.21.80–1.21.100 range the comment claims.

### Issues Found
#### Blocking Issues
- **Translation doesn't run for 1.21.80–1.21.100 clients.** Although the comment at `CustomBlockDefinitionSerializer843.java:51` says "1.21.80+", this serializer is only selected for protocol >= 843 (1.21.110) at `CustomBlockDefinitionSerializer.java:35-36`. Protocols 800 (1.21.80), 818 (1.21.90), 819 (1.21.93), and 827 (1.21.100) — all present in `SUPPORTED_PROTOCOLS` (`ProtocolInfo.java:125-126`) — bypass serialization entirely via `return nbt` at `CustomBlockDefinitionSerializer.java:38`, so those supported clients still receive `alpha_test` and keep the reported rendering defect.

#### Non-blocking Issues
- **No tests** cover the new conversion or its protocol boundary. Existing fixtures only use `opaque` and `blend` (`CustomBlockDefinitionSerializerTest.java:108`), so the missed 1.21.80 cutoff is undetected.

### Recommendations
- Introduce a translation-only serializer at protocol 800, then have serializer 843 add the `face_dimming` conversion on top. Do **not** simply lower the existing 843 dispatch threshold, because that would incorrectly apply the `face_dimming` → `packed_bools` conversion to 1.21.80–1.21.100 clients (that conversion belongs to the 1.21.110+ family).
- Add boundary tests covering: 1.21.70 retains `alpha_test`; 1.21.80/1.21.100 produce `alpha_test_single_sided`; 1.21.110+ inherit the translation; top-level and permutation materials; and that `opaque`, `blend`, `double_sided`, and already-modern values remain unchanged.

### Notes
- The translation itself is correctly scoped: only `minecraft:material_instances.materials` entries are processed, material compounds are guarded, and the rewrite is exact/case-sensitive. Top-level and permutation materials are both covered.
- Older clients (< 843) currently bypass the serializer, and for newer clients the definition is cloned before mutation, so serializing for a new client cannot poison the legacy definition later sent to an older client. Later serializers inherit the conversion via `CustomBlockDefinitionSerializer898` and `CustomBlockDefinitionSerializer975`.
- The Maven test run could not be completed in the review sandbox (Maven/JGit setup could not inspect the managed `.git` state).

---
🤖 AI Review

lt-name:
Based on available information, the two appear to be able to coexist.

PR #885, opened 2026-08-30
generator: look up block flags by id so custom blocks cannot crash population

w1zardz:
`Block.solid` and `Block.transparent` are plain arrays sized `MAX_BLOCK_ID`
(2048). Custom blocks live above `CustomBlockManager.LOWEST_CUSTOM_BLOCK_ID`
(10000) and are kept in the manager instead, which is why `Block` already
exposes `isBlockSolidById` / `isBlockTransparentById` for exactly this case.

Generators still indexed the raw arrays. The moment a populator reads a
neighbour cell that holds a custom block, the read throws
`ArrayIndexOutOfBoundsException: Index 10023 out of bounds for length 2048`
inside the asynchronous `PopulationTask`. The chunk is then never marked
populated, so it is never sent to the client: the player teleported there
keeps `teleportPosition` set, movement stays locked, the world around them
stays empty, and the vanilla air-time check eventually kicks them with
"Flying is not enabled on this server". Nothing in the server log points at
the chunk.

Every generator read now goes through the id-aware accessors, which return
the manager-cached flag for custom ids and the array value for vanilla ones.
Behaviour for vanilla blocks is unchanged.

### DIFF
--- a/src/main/java/cn/nukkit/level/generator/object/mushroom/BigMushroom.java
+++ b/src/main/java/cn/nukkit/level/generator/object/mushroom/BigMushroom.java
@@ -174,7 +174,7 @@ public class BigMushroom extends BasicGenerator {
                                 if (position.getY() >= position.getY() + i - 1 || meta != ALL_INSIDE) {
                                     Vector3 blockPos = new Vector3(l1, l2, i2);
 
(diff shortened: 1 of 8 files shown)


tachibana-kanade-angel:

PR #888, opened 2026-08-30
entity: stop AI mobs from jittering and pacing back and forth

w1zardz:
Three defects made every AI entity twitch and pace whenever it had no
player to chase (creative, or the player too far to be targeted), and
made a chasing mob turn back on itself every ten ticks.

* checkTarget() re-rolled a random wander point with a 5% chance on
  every single tick while the entity was resting, so a standing mob
  changed facing about once a second.
* The walking wander radius was 30 blocks with no minimum distance, so
  a mob that did start moving ran across half a chunk and back.
* updateMove() consumed route.next() in the same breath as it started
  an asynchronous route search. The search resets the node list and the
  cursor first, so the entity picked up node one of a freshly built
  path - a point right next to itself - and turned around.
* hasArrivedNode() compared node coordinates for exact equality, which
  a fractionally moving entity never reaches, so the normal per-arrival
  node advance never fired at all.

Body rotation is now rate limited to 30 degrees per tick instead of
snapping, and the random coin flip that rotated a resting entity on
half of all ticks is gone.

### DIFF
--- a/src/main/java/cn/nukkit/entity/BaseEntity.java
+++ b/src/main/java/cn/nukkit/entity/BaseEntity.java
@@ -797,6 +797,34 @@ public abstract class BaseEntity extends EntityCreature implements EntityAgeable
         return null;
     }
 
+    /**
+     * Насколько градусов за такт разрешено поворачивать тело сущности с ИИ.
+     *
+     * <p>Мгновенный разворот на произвольный угол клиент показывает рывком: он не сглаживает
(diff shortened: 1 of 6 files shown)


tachibana-kanade-angel:
Thanks for the fix, @w1zardz! 🎉 The AI mobs will no longer jitter and pace back and forth. This has been merged into Nukkit-MOT. Appreciate the contribution!

PR #889, opened 2026-08-30
fix(inventory): decode a stack request the way the live network sends it

w1zardz:
Two gaps in the item stack request decoder turn a normal craft into a decode
failure, and a decode failure used to cost the player the whole session.

An ingredient descriptor that names an item the server cannot resolve went
straight into Item.fromString(...).getId(), so an identifier we do not know threw
a NullPointerException out of the decoder. An unknown identifier now decodes as
air and the request is judged by the handler, which already answers ERROR and
resynchronises the client.

The recipe book containers (64, 65, 66) were missing from ContainerSlotType, so a
slot info that names one of them left fromId() empty and the decoder threw. They
are present in the protocol fork the production network runs on. Older protocols
are untouched: their branches already reject every id above DYNAMIC_CONTAINER,
and the encoder refuses to write the new ids below v1_21_20.

### DIFF
--- a/src/main/java/cn/nukkit/network/protocol/types/inventory/ContainerSlotType.java
+++ b/src/main/java/cn/nukkit/network/protocol/types/inventory/ContainerSlotType.java
@@ -75,7 +75,10 @@ public enum ContainerSlotType {
     /**
      * @since v712
      */
-    DYNAMIC_CONTAINER(63);
+    DYNAMIC_CONTAINER(63),
+    RECIPE_FOOD_CONTAINER(64),
+    RECIPE_BLOCKS_CONTAINER(65),
+    RECIPE_FURNACE_ITEMS_CONTAINER(66);
 
     private final int id;
     private static final Int2ObjectArrayMap<ContainerSlotType> VALUES = new Int2ObjectArrayMap<>();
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. Two gaps in the item-stack request decoder are addressed: unknown identifiers now decode as air instead of failing the whole batch, and the recipe-book container IDs (64–66) are added to `ContainerSlotType`. The protocol gating (`< v1_21_20` throws) keeps older versions safe.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- [P2] Protocol gate uses `v1_21_20`, while the pinned Cloudburst codec introduces IDs 64–66 in `Bedrock_v944` (1.26.10). Protocols 712–943 will now encode/decode values that did not exist in those versions, and NetEase bypasses the guard entirely. Consider gating the new IDs at `v1_26_10` to match the live network capability.
- [P2] The unknown-identifier fallback (`descriptorItem == null ? AIR : ...`) is effectively a no-op: `Item.fromString()` returns an AIR item (not `null`) for an unresolved identifier. Exceptions originating inside parsing (e.g. an out-of-range metadata reaching `Short.parseShort()`) still escape before the null check. Decoder-failure containment would need a non-throwing registry lookup or a guarded parse call.

### Recommendations
- Reconsider the protocol version at which the new container IDs are enabled.
- If decode-failure containment is the objective, guard the parse path so exceptions cannot fail the whole batch.

### Notes
- `git diff --check` passed. A full build could not run because Gradle could not initialize networking in the sandbox.

---
🤖 AI Review

@lt-name ready to merge

tachibana-kanade-angel (approved):

PR #890, opened 2026-08-30
player: scale movement sanity limits for authorized flight

w1zardz:
### Problem

A player with `ALLOW_FLIGHT` and `FLYING` set and a raised fly speed moves further per packet than
two hardcoded sanity limits allow: 100 (squared distance per move packet) and 225 (squared distance
per tick). At a fly speed around 20x vanilla, normal flight trips them constantly and the player is
rubber-banded back.

Both checks run before `PlayerInvalidMoveEvent` is fired, so a plugin cannot allow the movement it
authorized itself by raising the speed in the first place.

### Fix

Scale the two limits by the square of the configured speed ratio, and only when the player is
genuinely an authorized flier: `ALLOW_FLIGHT` and `FLYING` both set and `getFlySpeed()` a finite
value above the vanilla default. A walking player, a player who may fly but is not flying, vanilla
fly speed and a non-finite multiplier all keep the original thresholds untouched.

### Tests

`PlayerFlightMovementLimitTest` covers all four cases.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -640,6 +640,27 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         return this.adventureSettings.get(Type.ALLOW_FLIGHT);
     }
 
+    /**
+     * Scale the packet and tick movement sanity limits for an authorized player who is actually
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:

PR #891, opened 2026-08-30
entity: let a server see which item breeds or tames an animal

w1zardz:
Breeding and taming are the only ways a player creates new mobs out of nothing,
and both of them are invisible until they already happened. Every breeder knows
its own food - EntityWalkingAnimal has isFeedItem, EntityWolf has its own
isBreedingItem, the rabbit and the dolphin compare item ids inline in
onInteract - and nothing asks the entity from the outside. A server that wants
to keep a farm from growing without limit has to copy those three lists and
watch them drift, or refuse after the food has already been eaten.

BaseEntity gains isBreedingItem(Item), answered by the lists the entities
already carry, and EntityTameable gains isTamingItem(Item), answered by the wolf
with the bone it already looks for. Both are pure questions with no side
effects, so PlayerInteractEntityEvent becomes enough: a plugin sees the intent
while the item is still in the player's hand and refusing costs the player
nothing.

No behaviour changes: the rabbit and the dolphin now call the list they used to
compare inline, and every other entity answers false as before.

### DIFF
--- a/src/main/java/cn/nukkit/entity/BaseEntity.java
+++ b/src/main/java/cn/nukkit/entity/BaseEntity.java
@@ -488,17 +488,11 @@ public abstract class BaseEntity extends EntityCreature implements EntityAgeable
     }
 
     /**
-     * Does feeding this item to this entity start breeding.
-     *
-     * <p>Every breeder in the tree already knows the answer and every one of them keeps it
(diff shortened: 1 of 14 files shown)


tachibana-kanade-angel:

PR #893, opened 2026-09-01
feat(sound): add Sound.fromName() to resolve sounds by definition name

Suoerix:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR adds a static `Sound.fromName(String)` lookup method that resolves a `Sound` enum constant by its sound definition name (e.g. "note.harp"), iterating the enum values and returning null if no match is found. The implementation is functionally correct for the current enum and follows existing enum-lookup naming conventions.

### Issues Found
#### Blocking Issues
None.

#### Non-blocking Issues
- **[P2] Reverse lookup is a linear scan** — `Sound` has 1,254 constants. Each call allocates a new array via `values()` and performs up to 1,254 string comparisons, taking the worst-case path for unknown/null/empty names. Consider a one-time immutable `Map<String, Sound>` (guard null explicitly), consistent with `LevelSoundEventMap` and other reverse-lookup maps.
- **[P3] Duplicate-name behavior is an implicit invariant** — no duplicates exist today, but if one is added later the first declaration silently wins. Enforce uniqueness at map construction (reject or `putIfAbsent`).
- **[P3] Nullable contract not annotated** — Javadoc documents it, but consider JetBrains `@Nullable` on the return (and optionally the param) to match codebase conventions like `GameVersion.byName`.
- **[P3] No tests cover the new public API** — recommend known-value, round-trip, unknown/null/empty, and uniqueness tests.

### Recommendations
- Convert to a cached static reverse-lookup map; this is the only non-trivial change worth considering.
- Add focused unit tests for the new method.
- Optionally annotate with `@Nullable`.

### Notes
- Matching is exact and case-sensitive with no trimming, appropriate for definition identifiers.
- The Javadoc correctly distinguishes definition names from Java enum names used by `valueOf`.

---
@lt-name ready to merge
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #894, opened 2026-09-01
entity: recognize a totem by its item id

w1zardz:
Inventory deserialization and plugin-created stacks can carry the canonical
totem id in a plain Item object instead of ItemTotem. Entity.attack checked the
Java subclass, so such a real totem stayed in the player's hand while the
player died.

The item id is the stable protocol and inventory identity. Use it for both
hands and for the void/suicide client guard, while still rejecting empty
stacks. Consume exactly one item so even an invalid plugin-created stack cannot
lose every totem at once. Typed ItemTotem stacks keep the same behavior.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -6870,7 +6870,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                 }
             } else {
                 // 发包给客户端清除不死图腾，防止影响自杀等操作
-                if (this.getOffhandInventory().getItemFast(0) instanceof ItemTotem) {
+                if (Entity.isTotem(this.getOffhandInventory().getItemFast(0))) {
                     InventorySlotPacket pk = new InventorySlotPacket();
                     pk.slot = 0;
                     pk.item = Item.AIR_ITEM;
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The PR replaces the `instanceof ItemTotem` checks in `Entity.attack()` with an ID-based `isTotem(Item)` helper, so totems restored from inventory serialization (plain `Item` with ID 450) are correctly recognized. It also switches totem consumption from `clear(slot)` to `decreaseCount(slot)`, so a stacked totem only loses one item instead of the whole stack. A new test class covers the helper's ID/count semantics.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- `Player.kill()` keep-inventory cleanup still uses `instanceof ItemTotem` (`Player.java:6873` and `Player.java:6887`). A plugin-created plain `Item` totem held in the main hand is now recognized and consumed by `attack()` on normal lethal damage, but on SUICIDE/VOID deaths the client-side hide packets in `Player.kill()` will miss it, leaving a possible client "ghost" revival state. This behavior predates the PR and is an edge case, but for consistency the same ID-based check should eventually be applied there.
- `EntityTotemTest.consumesOnlyOneTotemFromEitherHand` asserts against source-code strings rather than runtime behavior. It would pass even if the wrong slot were decremented or an inventory event blocked consumption. A real inventory-level test would be more robust (understandable given the harness cost).

### Recommendations
- Consider sharing the ID-based totem check with `Player.kill()` (e.g. reuse `Entity.isTotem` or an equivalent) so all totem recognition sites behave identically.
- Offhand precedence and the null/zero-count guards in `isTotem` are correct; no changes needed there.

### Notes
- `decreaseCount` routes through normal inventory events and packets; a count-1 stack still ends up cleared via `setItem` → `clear`, so vanilla behavior is unchanged.
- The SUICIDE/VOID bypass branch correctly hides a plain ID-450 offhand stack without removing it server-side.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #895, opened 2026-09-01
item: keep totems single and preserve old overstacks

w1zardz:
Direct container writers could store two typed totems in one slot because
BaseInventory.setItem does not enforce the item's stack limit. Item.get also
falls back to a plain Item after a constructor failure, giving the canonical
totem id the base limit of 64.

Make the base Item report a limit of one for the totem id, so inventory
transfers and addItem keep the vanilla invariant even for a fallback object.
Do not truncate an already persisted totem overstack during NBT decoding: the
host migration needs the original count to split it across slots without
silently deleting paid or looted items. Other invalid stacks retain the old
clamp.

### DIFF
--- a/src/main/java/cn/nukkit/item/Item.java
+++ b/src/main/java/cn/nukkit/item/Item.java
@@ -1652,7 +1652,7 @@ public class Item implements Cloneable, BlockID, ItemID, ItemNamespaceId, Protoc
     }
 
     public int getMaxStackSize() {
-        return 64;
+        return id == TOTEM ? 1 : 64;
     }
 
     final public Short getFuelTime() {
--- a/src/main/java/cn/nukkit/nbt/NBTIO.java
+++ b/src/main/java/cn/nukkit/nbt/NBTIO.java
@@ -83,7 +83,11 @@ public class NBTIO {
             item.setCount(count);
         }
 
-        if (item.count > item.getMaxStackSize()) {
+        // Old direct container writes could persist two totems in one slot. Keep their exact
+        // count until the host migration can split them into separate slots; clamping here
+        // silently deletes the paid or looted tail before any inventory code can see it.
+        // Other malformed stacks keep the original defensive clamp.
+        if (item.count > item.getMaxStackSize() && item.getId() != Item.TOTEM) {
             item.count = item.getMaxStackSize();
             tag.putByte("Count", item.getMaxStackSize());
         }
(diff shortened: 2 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR enforces the vanilla single-stack invariant for totems on the base `Item` fallback and preserves legacy totem overstacks during NBT decoding so the host migration can split them instead of silently deleting items. The changes are small, focused, and well-documented. A new test covers the fallback stack limit, the overstack-preservation path, and the retained defensive clamp for other malformed stacks.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- **BaseInventory.setItem/setItemForce do not enforce stack limits** (pre-existing gap): a direct `setItem(index, item)` with a 2-count `ItemTotem` still succeeds — the write path remains unvalidated, and the new NBT exception makes such stacks persist across reloads. This was the original source of the overstack bug. Consider enforcing or splitting stacks at the inventory write path in a follow-up, plus a regression test that exercises an actual inventory write (not only `getMaxStackSize()`).
- **Global totem exemption in NBTIO.getItemHelper**: `getItemHelper` is a generic decoder with ~44 call sites (trade recipes, loot, dropped items, bundles, equipment, containers). Exempting every totem from the clamp globally means malformed/crafted NBT in any of those contexts can carry an overstacked totem (up to a byte-sized count). Preservation is only needed for the legacy container-migration path; consider scoping the exemption so newly created or malformed NBT in unrelated contexts (e.g. trade recipes, loot) can't smuggle in oversized totem stacks.
- **Whole-slot clear on totem use**: `Entity.attack` clears the entire hand/offhand slot when a totem activates, so a preserved 2-stack would be consumed entirely for one activation until the host migration splits it. Preserved overstacks should be migrated before gameplay, or consumption should decrement the count instead of clearing the slot.

### Recommendations
- Scope the NBTIO totem exemption to the container-migration path (e.g. an explicit preserve flag) rather than a global item-id check.
- Validate or split overstacks at `BaseInventory.setItem`/`setItemForce` so the direct-write path cannot create new overstacked totems.

### Notes
- Findings were verified against the code: `BaseInventory.setItem` performs no max-stack check; `Entity.attack` clears the whole slot; `getItemHelper` has 44 call sites. All flagged issues are either pre-existing gaps or documented design trade-offs (preservation is intentionally deferred to the host migration described in the PR body).
- The defensive clamp for other malformed stacks is retained and covered by tests.

---
🤖 AI Review

@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

PR #896, opened 2026-09-01
fix(level): suppress synchronous drops from creative placement

w1zardz:
A block can place itself, fail its support check during the nested update, and call useBreakOn before useItemOn returns. Creative mode does not consume the source item, so spawning that automatic drop prints a real item on every click. Moss carpet on a ladder is one deterministic example.

Track the exact world, cell and block id only while Block.place is on the stack. Suppress automatic drops from that same block when the placement started in creative mode. Nested non-creative placement masks an outer scope, player-authored breaks and neighbouring blocks remain untouched, and shulker boxes keep their NBT-bearing drop.

The scope is closed with try-with-resources so a false return or exception cannot leak creative provenance into later drops.

### DIFF
--- a/src/main/java/cn/nukkit/level/CreativePlacementDropGuard.java
+++ b/src/main/java/cn/nukkit/level/CreativePlacementDropGuard.java
@@ -1,0 +1,103 @@
+package cn.nukkit.level;
+
+import cn.nukkit.Player;
+import cn.nukkit.block.Block;
+import cn.nukkit.block.BlockShulkerBox;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Identifies an automatic break caused synchronously by the block currently being placed.
+ *
+ * <p>The stack is thread-local because a placement can re-enter item use through a plugin. Every
+ * placement pushes a scope, including non-creative ones, so an inner placement can never inherit
+ * an outer creative scope.
+ */
+final class CreativePlacementDropGuard {
+
+    private record Placement(Level level, boolean creative, int blockId, int x, int y, int z) {
+    }
+
+    private static final ThreadLocal<Deque<Placement>> ACTIVE =
+            ThreadLocal.withInitial(ArrayDeque::new);
+
+    private CreativePlacementDropGuard() {
+    }
+
+    static Scope enter(Level level, Player player, Block block) {
+        return enter(
+                level,
+                player != null && player.isCreative(),
+                block.getId(),
+                block.getFloorX(),
+                block.getFloorY(),
+                block.getFloorZ());
+    }
+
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a creative-mode item duplication: when a block places itself and fails its support check during the nested update (e.g. moss carpet on a ladder), the automatic break spawns a real item drop even though creative placement never consumed the source item. A thread-local scope (`CreativePlacementDropGuard`) records the world, cell, and block id while `Block.place` is on the stack and suppresses the automatic drop from that same block in `Level.useBreakOn`. The design is sound: nested non-creative placement masks an outer creative scope, player-authored breaks (`player != null`) and neighbouring blocks are unaffected, shulker boxes keep their NBT-bearing drop, and try-with-resources guarantees the scope cannot leak on failure or exception. Tests cover the guard's core semantics, including nesting and failure cleanup. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- Tests exercise only the guard's primitive overloads; they do not cover the production overloads or the actual `Level.useItemOn` → `Block.place` → `Level.useBreakOn` path, so wiring regressions in `Level.java` could pass all new tests.
- `ACTIVE.get().peek()` in `suppresses` creates and retains an empty deque on every thread that performs an ordinary break without an active placement. Small on the main thread, but avoidable for plugin executor threads (a nullable `ThreadLocal<Deque<Placement>>` would fix this).
- The shulker test supplies `MOSS_CARPET` with `shulkerBox=true`, a combination the production overload cannot produce; it verifies the boolean branch but not the `instanceof BlockShulkerBox` wiring.
- Matching the pre-placement block ID intentionally excludes placements that substitute another runtime block ID (e.g. standing-sign items placing wall-sign blocks). Does not affect the reported moss-carpet case, but the limitation should be documented if the guard is meant as a general mechanism.

### Recommendations
- Add a regression test for the motivating behavior: creative moss carpet placed over a non-solid, non-air support (ladder) should leave no dropped item; survival should preserve existing drop behavior.
- Add focused tests using the production overloads, including a real shulker block instance.
- Consider a nullable `ThreadLocal<Deque<Placement>>` so read-only misses do not install empty deques.
- Document the shulker exemption and same-runtime-ID limitation directly in the guard's javadoc.

### Notes
- ClamAV scan of the PR diff: clean.
- Security review: no network, process, reflection, or file-system concerns; scope is strictly limited to drop suppression.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:
Superseded by https://github.com/MemoriesOfTime/Nukkit-MOT/pull/897

PR #897, opened 2026-09-01
feat(level): fire a cancellable event for automatic block drops

w1zardz:
Delayed block physics can remove compound blocks long after placement has returned. Expose the exact source, whole drop batch and every main-layer cell synchronously replaced by the break chain so plugins can apply provenance rules and finish bookkeeping once at the outermost event. Direct drops from Block.onBreak are filtered at their exact source for both world- and player-started chains, while the player primary batch and block-entity contents remain on their existing paths.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockDoor.java
+++ b/src/main/java/cn/nukkit/block/BlockDoor.java
@@ -232,7 +232,6 @@ public abstract class BlockDoor extends BlockTransparentMeta implements Faceable
                 Block up = this.up();
 
                 if (up instanceof BlockDoor) {
-                    this.getLevel().setBlock(up, Block.get(BlockID.AIR), false);
                     this.getLevel().useBreakOn(this, getToolType() == ItemTool.TYPE_PICKAXE ? Item.get(ItemID.DIAMOND_PICKAXE) : Item.get(Item.WOODEN_PICKAXE)); // Drop iron doors
                 }
 
--- a/src/main/java/cn/nukkit/block/BlockPistonBase.java
+++ b/src/main/java/cn/nukkit/block/BlockPistonBase.java
@@ -219,11 +219,10 @@ public abstract class BlockPistonBase extends BlockSolidMeta implements Faceable
             List<Block> destroyBlocks = calculator.getBlocksToDestroy();
             for (int i = destroyBlocks.size() - 1; i >= 0; --i) {
                 Block block = destroyBlocks.get(i);
-                this.level.useBreakOn(block, null, null, true);
-
                 if (Server.getInstance().dropSpawners && block instanceof BlockMobSpawner){
-                    this.level.dropItem(block.add(0.5, 0.5, 0.5), Item.get(Item.MONSTER_SPAWNER, 0, 1));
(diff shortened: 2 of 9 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR introduces `AutomaticBlockDropsEvent` to expose drops from automatic (non-player) block break chains, with a trace (`AutomaticBlockBreakTrace`) that records every main-layer block replaced during the chain and a guard (`CreativePlacementDropGuard`) that suppresses drops from creative placement side effects. The overall design is sound: the event has a proper static `HandlerList`, is dispatched via `callEvent`, and the terminal/direct drop split is well documented. One correctness issue in nested break chains needs to be addressed before merge.

### Issues Found
#### Blocking Issues
- `Level.java:2683-2693`, `Level.java:2943-2973`, `AutomaticBlockBreakTrace.java:27-39` — Nested automatic breaks reclassify their primary drops as direct `onBreak` drops. When an inner automatic break runs inside an outer break's `onBreak`, the inner scope closes and the inner terminal event fires first, but the outer scope is still active. Because `record()` copies every change into *all* active scopes, the inner target block is present in the outer scope's changed map, so when the inner (already filtered) drops spawn via the public `dropItem` path at `Level.java:2971`, `directDropSource` matches them again and fires a second `AutomaticBlockDropsEvent` with an empty changed-block list and `isOutermostBreak() == false`. Consequences: (1) listeners receive duplicate events for one physical drop, and (2) a plugin that cancels direct/secondary events can suppress legitimate nested primary drops that already passed the inner terminal event. This contradicts the stated goal of finishing bookkeeping once per drop. Verified against the code: inner drop spawning happens after the inner scope's `close()` but while the outer scope remains active.

#### Non-blocking Issues
- `BlockSnowLayer.java:88-92` — Coverage gap (pre-existing path, not in this diff): `onUpdate` calls `useBreakOn` and then drops a snowball via `level.dropItem` *after* the trace has closed, so no `AutomaticBlockDropsEvent` is fired for that drop. No regression — plugins had no event before — but similar "break then manually drop" callers bypass the new API. Worth a follow-up to route such drops through the event or document them as out of scope.
- `AutomaticBlockDropsEvent.java:17-18`, `BlockPistonBase.java:222-225` — The javadoc states the source block "has already been removed from the level", but direct-drop events (and the piston spawner drop via `dropItemFromAutomaticBlock`) can fire before removal. The contract should clarify that only terminal events guarantee post-removal state.

### Recommendations
- Guard the direct-drop reclassification so drops spawned from an inner break's terminal batch are not re-tagged while an outer scope is active (e.g. only treat a `dropItem` call as a direct drop when it originates from inside `Block.onBreak`, or track drops already emitted by a terminal event).
- Add a test covering a nested automatic break (break A → break B) asserting exactly one event per physical drop batch.

### Notes
- Event class correctly declares a static `HandlerList` and is dispatched through the plugin manager; listener registration works.
- Player-authored breaks intentionally emit an empty terminal batch (primary drops stay on `BlockBreakEvent`); block-entity contents are intentionally excluded via `enterBlockEntityDrops`. Both design choices are documented and consistent.
- New tests (`AutomaticBlockDropsEventTest`, `AutomaticBlockBreakTraceTest`, `CreativePlacementDropGuardTest`) cover the basic paths but not the nested-break scenario above.

---
🤖 AI Review

PR #898, opened 2026-09-01
fix(entity): bound client absorption capacity on 1.26

w1zardz:
Bedrock 1.26 renders the advertised absorption maximum as HUD capacity.
Nukkit sent Float.MAX_VALUE, so one enchanted golden apple created thousands
of empty hearts and made Android clients stall even though the real absorption
value remained 16.

Keep the server-side attribute range unchanged for plugins and older clients.
Only the per-player packet for 1.26+ gets a finite capacity, never smaller
than vanilla enchanted-apple absorption or the actual custom value.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2485,10 +2485,21 @@ public abstract class Entity extends Location implements Metadatable {
         return absorption;
     }
 
+    static Attribute clientAbsorptionAttribute(float absorption, int protocol) {
+        Attribute attribute = Attribute.getAttribute(Attribute.ABSORPTION);
+        if (protocol >= ProtocolInfo.v1_26_0) {
+            // Bedrock 1.26 builds one HUD heart slot per advertised capacity point. Sending the
+            // server-side Float.MAX_VALUE range therefore freezes mobile clients. Preserve
+            // custom values above vanilla's 16 absorption points instead of clipping them.
+            attribute.setMaxValue(Math.max(16.0f, absorption));
+        }
+        return attribute.setValue(absorption);
+    }
+
     public void setAbsorption(float absorption) {
         if (absorption != this.absorption || (this instanceof Player player && player.protocol >= ProtocolInfo.v1_21_60)) {
             this.absorption = absorption;
-            if (this instanceof Player player) player.setAttribute(Attribute.getAttribute(Attribute.ABSORPTION).setValue(absorption));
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a client freeze on Bedrock 1.26 where sending `Float.MAX_VALUE` as the absorption attribute max value caused mobile clients to render thousands of empty hearts. The change adds a `clientAbsorptionAttribute()` helper that caps the advertised HUD capacity at `Math.max(16, absorption)` for 1.26+ clients while leaving the server-side range and older protocols untouched. The fix is targeted and well-documented, and the helper tests cover full, damaged, and custom absorption values. Vanilla behavior (≤16 absorption points) is fully correct, and the freeze root cause is fixed.

### Issues Found

#### Blocking Issues
- None.

#### Non-blocking Issues
- **Custom absorption capacity shrinks as damage is taken** (`Entity.java:2494`): the advertised max derives from the *current* absorption value. A custom 24-point shield damaged to 23 advertises `(max=23, value=23)`, so a 1.26 client HUD shows full capacity instead of 23/24. Only affects plugin-set custom values above 16; vanilla absorption (≤16) is unaffected since the floor of 16 keeps the capacity stable.
- **Client-facing capacity is still unbounded for arbitrary plugin inputs**: `Math.max(16.0f, absorption)` passes through very large or non-finite (`NaN`/`Infinity`) values — `Attribute.setMaxValue()`/`setValue()` do not reject non-finite floats, so a plugin setting a huge or NaN absorption could still advertise an unbounded/malformed capacity to 1.26 clients. Edge case, plugin-controlled only; vanilla gameplay is unaffected.
- Tests exercise only the helper; no packet-level regression test for 1.26 vs older protocols, damage path, or invalid/oversized inputs.

### Recommendations
- Consider tracking an advertised capacity independently from the current absorption value so custom shields retain their original HUD capacity while damage consumes them.
- Consider sanitizing non-finite/absurdly large absorption values before advertising to 1.26 clients.

### Notes
- Tests could not be executed in the review environment; the review is based on static analysis of the diff.
- The two findings above only affect plugin-driven custom absorption (>16 points). The vanilla freeze scenario this PR targets is fixed, and older protocols are explicitly left unchanged.

---

@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #977, which contains the same changes (the 1.26 client absorption capacity bound) along with additional fixes, rebased onto the current master.

PR #901, opened 2026-09-01
fix(entity): keep pathfinding body yaw while tracking a target

w1zardz:
Walking mobs used lookAt(followTarget) before turning towards the current route node. That reset body yaw to the player every tick, so the 30-degree smoothing limit could never reach a lateral detour and obstacle checks inspected the wrong facing.

Track the live target with head yaw only after the body turns towards the route. Keep the old body-and-head behavior for wandering without a live target, and cover both orientation paths.

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityLiving.java
+++ b/src/main/java/cn/nukkit/entity/EntityLiving.java
@@ -592,6 +592,17 @@ public abstract class EntityLiving extends Entity implements EntityDamageable {
     }
 
     public void lookAt(Vector3 target) {
+        this.lookAt(target, true);
+    }
+
+    /**
+     * Turn only the entity's head towards a target while preserving the body's route-facing yaw.
+     */
+    protected void lookHeadAt(Vector3 target) {
+        this.lookAt(target, false);
+    }
+
+    private void lookAt(Vector3 target, boolean turnBody) {
         double dx = this.x - target.x;
         double dy = this.y - target.y;
         double dz = this.z - target.z;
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes walking mobs resetting their body yaw to the follow target every tick by introducing head-only tracking (`lookHeadAt`) while the body turns toward the current route node. The orientation logic itself is internally consistent and both paths are covered by the new `EntityWalkingOrientationTest`. However, two integration problems with the surrounding code were confirmed during review.

### Issues Found
#### Blocking Issues
- **Ranged mobs now aim along the route instead of at the target.** `EntityWalkingMob.onUpdate()` runs `updateMove()` and then `attackEntity()` in the same tick. After this change, `updateMove()` leaves `yaw` facing the route node and puts the target angle only in `headYaw`, but `attackEntity()` in `EntityPillager` (`EntityPillager.java:63-71`), `EntityPiglin`, `EntityDrowned`, `EntitySnowGolem` and `EntityWitch` builds the projectile spawn position and motion from `this.yaw`/`this.pitch`. Whenever the path deviates laterally from the target (exactly the scenario this PR enables), arrows/tridents/snowballs/potions fly along the detour direction and can be blocked by obstacles on the route. Before this PR, `lookAt(followTarget)` kept body yaw on the target every tick, so ranged aim was correct.
- **Head-only yaw changes are never sent to clients.** `lookHeadAt` changes `headYaw` while `yaw` stays fixed, but `Entity.updateMovement()` only emits a movement packet when the position or yaw/pitch changes (`Entity.java:2220-2248`); `headYaw` is written to `lastHeadYaw` but is never part of the trigger condition. A stationary or blocked mob therefore tracks the target server-side while clients keep rendering the previous head direction.

#### Non-blocking Issues
- None confirmed.

### Recommendations
- Derive ranged aim from the attack target (or from the target-facing `headYaw`) instead of the route-facing body yaw, e.g. compute yaw/pitch toward the target inside `attackEntity`.
- Include head-yaw changes in the `Entity.updateMovement()` emission condition so head tracking is visible for stationary mobs, and add a regression test for that case.

### Notes
- Static review only; no build or tests were run locally.
- The new test covers both orientation paths introduced by this PR.

---
🤖 AI Review

tachibana-kanade-angel:
This PR has been superseded by #977, which contains the same changes (the pathfinding body-yaw fix) along with additional fixes, rebased onto the current master.

PR #903, opened 2026-09-01
fix(player): consume one arrow per crossbow charge

w1zardz:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR reworks crossbow click handling in `Player` to fix repeated arrow consumption caused by trailing `CLICK_AIR` packets Bedrock sends after a charge completes. A slot-scoped state machine (`decideCrossbowClick`) plus a 12-tick tail-suppression window replaces the previous direct `onClickAir` dispatch. The overall direction is reasonable and the `ItemCrossbow` consumption path (creative bypass, offhand preference, Multishot consuming one projectile, no Infinity effect) is untouched and correct. However, two logic issues in the new state machine need to be addressed before merge.

### Issues Found
#### Blocking Issues
- **Tail-suppression window swallows legitimate shots and can extend indefinitely** (`Player.java`, `decideCrossbowClick` / `IGNORE_TAIL`). `ItemCrossbow.launchArrow` only requires `getTick() - loadTick > 10` (ItemCrossbow.java:161), so a charge completed at tick T can legally be fired at T+11. The new code suppresses every click up to T+12, and each suppressed click re-extends the window by another 12 ticks. A player who clicks immediately after the charge completes (normal gameplay) has the shot silently ignored, and continuous rapid clicking can prevent a charged crossbow from ever firing. Consider using `UseItemData.triggerType` (`PLAYER_INPUT` vs `SIMULATION_TICK`, already parsed in `InventoryTransactionPacket`) to distinguish trailing simulation packets from real input instead of a purely time-based window.
- **Charge state is keyed to the hotbar slot, not the item stack** (`Player.java:5593`, `PlayerInventory.java:110`). `resetCrossbowInputState()` only runs when `itemInHandIndex` changes. Replacing the crossbow in the *same* slot (inventory transaction, plugin, etc.) mid-charge leaves `usingItem`/`startAction` and the tail window intact, so the replacement crossbow can finish charging using the previous item's elapsed time and consume an arrow without being held for its own charge duration; a charged replacement likewise inherits the old tail suppression. Reset on item-stack change (or validate stack identity) in addition to slot change.

#### Non-blocking Issues
- `PlayerCrossbowInputTest` simulates the decision table but never exercises real arrow consumption — it performs `arrows--` itself whenever `FINISH_LOADING` is returned, so it would pass even if production consumed zero or multiple arrows. Consider coverage against the real `ItemCrossbow.onUse`/inventory path for: survival vs creative, offhand ammunition, Multishot single consumption, no-Infinity, insufficient charge time, no ammunition, same-slot item replacement, and an intentional shot inside the suppression window.

### Recommendations
- Gate tail suppression on `triggerType == SIMULATION_TICK` where available, falling back to the time window on older protocols.
- Track the charged item stack identity (or reset on same-slot replacement) so charge/tail state cannot transfer between items.

### Notes
- The `PlayerInventory.setHeldItemSlot` refactor preserves viewer notification via `setHeldItemIndex(...).sendHeldItem(viewers)`, behavior is equivalent.
- Spectator handling is still guarded earlier in the `CLICK_AIR` path (`useClientSpectator` early-return + cancelled `PlayerInteractEvent`), so the removed crossbow-specific spectator check does not open a hole.

---
🤖 AI Review

PR #904, opened 2026-09-01
fix(block): match enchanting table and cauldron collisions

w1zardz:
The client treats the enchanting table as twelve pixels high and the
cauldron as a hollow frame with a five-pixel bottom. The server used a
full cube for both, so movement onto either block was corrected back and
forth against geometry the client could not see.

Expose composite collision boxes to the movement helper, match both
shapes to Bedrock, and cover the solid and hollow parts with regressions.

### DIFF
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -1150,6 +1150,15 @@ public abstract class Block extends Position implements Metadatable, Cloneable, 
         return collidesWithBB(bb, false);
     }
 
+    public void addCollisionBoxesToList(AxisAlignedBB bb, List<AxisAlignedBB> collidingBoxes) {
+        if (this.collidesWithBB(bb)) {
+            AxisAlignedBB boundingBox = this.getBoundingBox();
+            if (boundingBox != null) {
+                collidingBoxes.add(boundingBox);
+            }
+        }
+    }
+
     public boolean collidesWithBB(AxisAlignedBB bb, boolean collisionBB) {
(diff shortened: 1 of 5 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes correctly model the enchanting table as 12/16 blocks high and the cauldron as a 5/16-high bottom plus four 2/16-thick walls, matching Bedrock client geometry. The default `Block.addCollisionBoxesToList` preserves existing single-box behavior, and the stair-specific virtual shape predicate is retained.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- `BlockCollisionShapeTest.java`: the cauldron test samples positions but does not verify exact dimensions — a 4px bottom or 1px rim would still pass the current samples. Add boundary assertions immediately below/above 5/16 and 2/16.
- `BlockCollisionShapeTest.java`: tests call `addCollisionBoxesToList` directly but never exercise `CollisionHelper.getCollisionCubes`, which is the production integration changed in `CollisionHelper.java`. A level-backed test should verify the cavity returns no collision cubes while bottom and rim do.

### Recommendations
- Tighten cauldron test boundaries around 5/16 (bottom) and 2/16 (rim).
- Add an integration test through `CollisionHelper.getCollisionCubes`.

### Notes
- Static review only; the build was not compiled or run.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:
Superseded by https://github.com/MemoriesOfTime/Nukkit-MOT/pull/967

PR #905, opened 2026-09-01
player: bound Frost Walker radius before scanning blocks

w1zardz:
The movement handler trusted an item enchantment level when choosing the nested block-scan radius. An imported or otherwise malformed Frost Walker level could therefore turn one movement tick into billions of block reads before any cancellable event was fired.

Clamp the level to the vanilla maximum before deriving the radius. Keep the calculation in a pure helper so regression tests prove that ordinary levels remain unchanged and extreme integer values cannot enlarge the scan.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a real DoS vector: an item carrying a malformed/imported Frost Walker level (e.g. `Integer.MAX_VALUE`) is trusted when deriving the block-scan radius in `handleEnchantmentInMove`, turning one movement tick into billions of block reads. The extraction into a pure helper `frostWalkerRadius(int)` plus the regression test (`PlayerFrostWalkerRadiusTest`) is a clean approach, and the test was verified to pass.

However, the chosen bound overrides an intentional, documented server configuration.

### Issues Found
#### Blocking Issues
- **Hard-coded vanilla cap silently overrides `forced-safety-enchant=false`** (`Player.java`, `frostWalkerRadius`). When `forced-safety-enchant=false` is set, the server deliberately preserves administrator-created enchantment levels above the normal maximum (`Enchantment.setLevel(level, safe)` skips the max-level clamp; NBT loading in `Item.java:1337/1399` passes this flag through). With this PR, a legitimate Frost Walker X on such a server shrinks from radius 12 to radius 4 with no warning. The scan must be bounded, but the bound should be a separate **safe maximum radius** (e.g. cap the radius at a fixed safe value such as 16), not the enchantment's vanilla level cap — so configured over-level behavior is preserved while the unbounded-scan vulnerability stays fixed.

#### Non-blocking Issues
- None.

### Recommendations
- Keep the pure-helper + regression-test structure, but change the helper to clamp the resulting radius to a safe absolute bound instead of clamping the enchantment level to `MAX_FROST_WALKER_LEVEL` (2), e.g. `Math.min(radius, SAFE_MAX_RADIUS)`.
- Add a test case documenting the behavior for over-level enchantments under the chosen bound.

### Notes
- The underlying vulnerability (unbounded nested block scan from a trusted NBT enchantment level) is real and worth fixing promptly.
- The dedicated test was executed against the PR head and passes.

---
🤖 AI Review

tachibana-kanade-angel:

PR #906, opened 2026-09-03
fix(block): give crying obsidian the Bedrock hardness 35

w1zardz:
BlockCryingObsidian returns hardness 50, copied from Java Edition. Bedrock uses 35 — the same value BlockObsidian already returns.

Because of that a plain diamond pickaxe finishes the block in ~6.6 s while the server expects ~9.4 s, so every break is classified as a fast break and cancelled: the block reappears and nothing drops. Reproduced on every protocol from 786 to 2168, so it is the block data and not a client version issue.

Value confirmed against PocketMine-MP `VanillaBlocksInputs` (`35.0 /* 50 in Java */`).

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockCryingObsidian.java
+++ b/src/main/java/cn/nukkit/block/BlockCryingObsidian.java
@@ -20,10 +20,6 @@ public class BlockCryingObsidian extends BlockSolid {
 
     @Override
     public double getHardness() {
-        // Bedrock parity: crying obsidian has hardness 35 on Bedrock (50 in Java), the same as
-        // obsidian. With 50 the server expected 9.4 s for a plain diamond pickaxe while every
-        // client finished in 6.6 s, so each break was refused as too fast and the block
-        // reappeared. Confirmed against PocketMine-MP VanillaBlocksInputs (35.0 /* 50 in Java */).
         return 35;
     }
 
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR changes `BlockCryingObsidian#getHardness()` from 50 (Java Edition value) to 35 (Bedrock value), matching `BlockObsidian`. With hardness 50 the server computed a ~9.4 s break time for a plain diamond pickaxe while the client finished in ~6.6 s, so every break was flagged as a fast break and cancelled — the block reappeared and dropped nothing. With 35, `Block#calculateBreakTimeNotInAir()` yields ~6.56 s (hardness × 1.5 ÷ 8), aligning server-side timing with the client and eliminating the false fast-break cancellations. `getResistance()` (1200) is untouched, so blast resistance is unaffected. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- Consider adding a focused regression test for crying obsidian's hardness and diamond-pickaxe break time if block timing tests exist.

### Notes
- Value verified against PocketMine-MP `VanillaBlocksInputs` (35.0) and consistent with `BlockObsidian` in this codebase.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #907, opened 2026-09-03
fix(block): keep the end portal frame 13/16 tall once the eye is in

w1zardz:
BlockEndPortalFrame raises its collision box to a full block as soon as the eye bit is set. That is Java Edition behaviour: on Bedrock the frame stays 13/16 tall with or without the eye.

The client therefore keeps standing at y + 0.8125 while the server believes the floor is at y + 1. Survival players do not notice, but a creative or flying player is corrected on every move and gets thrown back and forth as long as he stands on a filled frame.

Value confirmed against PocketMine-MP `EndPortalFrame`, which returns `AxisAlignedBB::one()->trim(Facing::UP, 3 / 16)` with no eye branch at all.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockEndPortalFrame.java
+++ b/src/main/java/cn/nukkit/block/BlockEndPortalFrame.java
@@ -63,12 +63,17 @@ public class BlockEndPortalFrame extends BlockTransparentMeta implements Faceabl
 
     @Override
     protected AxisAlignedBB recalculateBoundingBox() {
+        // Bedrock parity: the frame is 13/16 tall with or without the eye. Java raises the box to
+        // a full block once the eye is in; Bedrock clients keep standing at 0.8125, so the server
+        // saw every creative player 0.1875 deep inside the block and pushed them back out on each
+        // move. Confirmed against PocketMine-MP EndPortalFrame
+        // (AxisAlignedBB::one()->trim(Facing::UP, 3 / 16), no eye branch).
         return new SimpleAxisAlignedBB(
                 this.x,
                 this.y,
                 this.z,
                 this.x + 1,
-                this.y + ((this.getDamage() & 0x04) > 0 ? 1 : 0.8125),
+                this.y + 0.8125,
                 this.z + 1
         );
     }


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

The change removes the eye-state branch in `BlockEndPortalFrame.recalculateBoundingBox()` (`src/main/java/cn/nukkit/block/BlockEndPortalFrame.java:65`), keeping the collision box at `0.8125` (= 13/16) whether or not an eye is inserted. This matches Bedrock behavior (and PocketMine-MP's `EndPortalFrame`, which trims 3/16 from the top with no eye branch) and fixes the client/server height mismatch that caused creative/flying players to be pushed back on every move while standing on a filled frame. Empty-frame behavior is unchanged.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `this.y + 0.8125` would read more clearly as `this.y + 13.0 / 16.0`, matching dimensional expressions elsewhere in the codebase. The current value is exact and correct.
- No regression test covers the two metadata states. A small test asserting `maxY - minY == 13 / 16` for damage values with and without the eye bit (0x04) would protect this behavior.
- The six-line implementation comment is somewhat verbose for a straightforward invariant; a shorter note would be easier to maintain.

### Recommendations
- Merge as-is, optionally adding the metadata-state bounding-box test.
- Manually verify standing/flying over a filled frame and projectile passage through the upper 3/16, since both use the same generic AABB.

### Notes
- Eye insertion and portal completion are independent of geometry and remain unchanged (`BlockEndPortalFrame.java:107`).
- Comparator output still reads the eye metadata bit directly (`BlockEndPortalFrame.java:97`), so it stays 15 with an eye and 0 without.
- Stronghold-generated frames with eyes use the same metadata bit and receive the corrected shape automatically.
- Ray intersection and block picking do not depend on the old full height; block-pick uses client-supplied coordinates and `toItem()`.
- Face solidity is derived from the AABB (`Block.java:1298`), so a filled frame's top/side faces are no longer treated as full. This may affect redstone dust support (`BlockRedstoneWire.java:247`) and explosion fire placement (`Explosion.java:313`), but it removes a state-dependent behavior caused by the previously incorrect full-height box and is consistent with Bedrock.
- Safe-spawn lookup is unaffected: `Level.isFullBlock` checks `isSolid()` before consulting the AABB.
- `git diff --check` is clean.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #908, opened 2026-09-03
fix(entity): restore vanilla melee knock-back and the Knockback enchantment push

w1zardz:
Melee knock-back has been 0.3 since 8dca6ffd, and the Knockback enchantment only adds 0.1 per level on top of it. Vanilla Bedrock — and PocketMine-MP, which mirrors it — pushes with 0.4 and applies the enchantment as a **separate** knock-back of 0.5 per level right after the hit lands (`Living::DEFAULT_KNOCKBACK_FORCE`, `KnockbackEnchantment::onPostAttack`).

The gap is large enough to change how melee feels: a bare hit pushes 25% less far and, because the vertical cap equals the force, lifts the victim noticeably less, while a Knockback I sword ends up weaker than a bare vanilla hit (0.4 total against 0.4 plus a 0.5 push).

This PR restores the vanilla numbers:

* `EntityLiving.DEFAULT_KNOCKBACK` is 0.4 and is used by the short `knockBack` overload and by both `EntityDamageByEntityEvent` constructors that default it;
* a melee hit from a player asks for that force instead of 0.3;
* `EnchantmentKnockback` pushes the victim again with `level * 0.5` from `doPostAttack`, the way the enchantment works in vanilla, instead of nudging the base force.

The horizontal direction and the vertical cap keep their existing meaning, so this is a value fix, not a new mechanic.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -5709,11 +5709,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                             Map<DamageModifier, Float> damage = new EnumMap<>(DamageModifier.class);
                             damage.put(DamageModifier.BASE, itemDamage);
 
-                            float knockBack = 0.3f;
-                            Enchantment knockBackEnchantment = item.getEnchantment(Enchantment.ID_KNOCKBACK);
-                            if (knockBackEnchantment != null) {
-                                knockBack += knockBackEnchantment.getLevel() * 0.1f;
-                            }
+                            float knockBack = (float) EntityLiving.DEFAULT_KNOCKBACK;
 
                             EntityDamageByEntityEvent entityDamageByEntityEvent = new EntityDamageByEntityEvent(this, target, DamageCause.ENTITY_ATTACK, damage, knockBack, enchantments);
                             entityDamageByEntityEvent.setBreakShield(item.canBreakShield());
--- a/src/main/java/cn/nukkit/entity/EntityLiving.java
+++ b/src/main/java/cn/nukkit/entity/EntityLiving.java
@@ -55,6 +55,11 @@ public abstract class EntityLiving extends Entity implements EntityDamageable {
     protected int attackTime = 0;
     protected int knockBackTime = 0;
 
+    /**
+     * Vanilla Bedrock knock-back force. Both the horizontal push and the vertical cap use it.
+     */
+    public static final double DEFAULT_KNOCKBACK = 0.4;
+
     protected float movementSpeed = 0.1f;
 
     protected int turtleTicks = 0;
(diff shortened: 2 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR restores vanilla Bedrock melee knock-back: the base force moves from 0.3 to 0.4 via a shared `EntityLiving.DEFAULT_KNOCKBACK` constant, and the Knockback enchantment is reworked from a +0.1/level nudge into a separate `0.5 * level` push fired from `doPostAttack`, matching vanilla/PocketMine-MP behavior. The main player attack path (`Player.java`) is updated correctly: the push fires only after a successful, non-cancelled `attack()`, the direction math reuses the existing `knockBack` semantics, and null/type safety is handled by the `instanceof EntityLiving` guard.

However, two other attack paths that also invoke `EnchantmentKnockback.doPostAttack` still use the old `0.3 + level * 0.1` base force, causing the enchantment to be double-counted there.

### Issues Found
#### Blocking Issues
- `InteractProcessor_v113.java:138-141` — The v1.1.x client attack path still computes `knockBack = 0.3f + level * 0.1f`, then calls `enchantment.doPostAttack(player, target)`. With this PR, enchanted hits through this path get both the obsolete `+0.1/level` base bonus and the new separate `+0.5/level` push, while bare hits stay at the old 0.3 instead of 0.4. This path needs the same update as `Player.java` (use `EntityLiving.DEFAULT_KNOCKBACK`, drop the per-level adjustment).
- `ItemSpear.java:103-106` — Spear jab attacks have the identical stale logic (`0.3f + level * 0.1f`) and also call `doPostAttack`. This is especially relevant because `EnchantmentKnockback.canEnchant()` explicitly permits spears, so a Knockback spear now double-counts the enchantment and a bare spear keeps the old 0.3 base.

#### Non-blocking Issues
- None.

### Recommendations
- Update `ItemSpear` and `InteractProcessor_v113` to use `EntityLiving.DEFAULT_KNOCKBACK` and remove the `level * 0.1f` adjustment, so all three melee paths behave identically.
- Consider a quick in-game sanity check that a Knockback II sword now produces a noticeably larger push than a bare hit (0.4 base + 2 x 0.5 separate pushes).

### Notes
- Ordering, event-cancellation handling, and NPE safety in the new `doPostAttack` implementation are correct for the main attack path.
- The `damageArmor` path in `EntityHumanType` also calls `doPostAttack` for armor enchantments, but Knockback cannot legitimately appear on armor, so this is not a practical concern.

---
🤖 AI Review

w1zardz:
Added a second commit: cap the vertical part of a knock-back at the vanilla limit. With the enchantment push restored, `knockBack` capped the vertical component at the force it was given, so Knockback I lifted the victim by 0.5 and Knockback II by 1.0, several times a jump. Bedrock (and PocketMine-MP's `Living::DEFAULT_KNOCKBACK_VERTICAL_LIMIT = 0.4`) keeps the vertical ceiling separate from the force. The five-argument overload keeps its old meaning for every force at or below 0.4, so an ordinary hit, a shield bash and an event-authored knock-back move exactly as before.

tachibana-kanade-angel:
This PR has been superseded by #982, which contains the same changes (the vanilla melee knock-back restoration) along with additional fixes, rebased onto the current master.

Closing this PR to avoid double-applying these changes. If any part of this PR is not covered by #982, feel free to reopen or comment and we will split it out.

PR #909, opened 2026-09-03
fix(resourcepack): order the client stack the way the server declared it

w1zardz:
`ZippedResourcePackLoader` builds the pack stack from `listFiles()`, which returns filesystem order. The stack sent to every client therefore depends on inode order: two packs that override the same texture atlas key can swap priority after a redeploy, and an archive left in the folder outranks the packs the owner listed on purpose. Nothing is logged either way.

Read the `resource_stack` list from `resource_packs.yml`, keep its order, and append the packs that are on disk but not declared after it with a warning.

### DIFF
--- a/src/main/java/cn/nukkit/resourcepacks/loader/ZippedResourcePackLoader.java
+++ b/src/main/java/cn/nukkit/resourcepacks/loader/ZippedResourcePackLoader.java
@@ -1,7 +1,9 @@
 package cn.nukkit.resourcepacks.loader;
 
+import cn.nukkit.Nukkit;
 import cn.nukkit.Server;
 import cn.nukkit.resourcepacks.ResourcePack;
+import cn.nukkit.utils.Config;
 import cn.nukkit.resourcepacks.ZippedResourcePack;
 import com.google.common.io.Files;
 import lombok.extern.log4j.Log4j2;
@@ -73,7 +75,7 @@ public class ZippedResourcePackLoader implements ResourcePackLoader {
     public List<ResourcePack> loadPacks() {
         var baseLang = Server.getInstance().getLanguage();
         List<ResourcePack> loadedResourcePacks = new ArrayList<>();
-        for (File pack : path.listFiles()) {
+        for (File pack : orderedPackFiles()) {
             if (shouldIgnoreFile(pack.getName())) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The PR makes the resource pack stack deterministic by reading the declared `resource_stack` list from `resource_packs.yml`, preserving its order, and appending on-disk-but-undeclared packs afterwards with a warning. The comparator logic is correct (declared index first, undeclared appended after, name as tie-breaker), and the no-config fallback now yields a stable alphabetical order instead of filesystem order. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- **Declared-but-missing packs are silently omitted** (`ZippedResourcePackLoader.java:128-135`): reconciliation only iterates files found on disk, so a declared entry that matches nothing (typo, or a pack deleted from the folder) produces no diagnostic. Impact is limited — a misspelled entry usually still surfaces indirectly, because the actual on-disk pack is then reported as undeclared — but a warning here would close the loop.
- **Explicitly empty `resource_stack` is indistinguishable from a missing key** (`ZippedResourcePackLoader.java:136`): when the key exists but is an empty list, the `!declared.isEmpty()` guard suppresses the undeclared-pack warnings. Behavior stays deterministic (alphabetical), so this is an observability gap, not an ordering bug.
- **`getStringList` silently coerces or drops malformed entries** (`ZippedResourcePackLoader.java:153`): non-scalar YAML entries (maps/lists) are dropped without warning, and numbers/booleans are stringified. An entirely malformed list degrades to the no-key path, suppressing all warnings. This is pre-existing `Config` behavior, not introduced here.
- **Undeclared warning fires for entries that will never load** (`ZippedResourcePackLoader.java:137-141`): every directory entry except `packs.yml` — including unsupported files and manifest-less directories — gets "applied below every declared pack", which is misleading and may be followed by a second unknown-format warning.

### Recommendations
- Track key presence separately (e.g. `Config#exists`) so an explicitly empty `resource_stack` still emits undeclared-pack warnings.
- Log a warning for declared entries that match no file on disk.
- Restrict the undeclared warning to entries that are actually loadable (`.zip`/`.mcpack` files or directories containing a manifest).

### Notes
- Verified against head commit `b0282526`: the sort is deterministic; servers without `resource_packs.yml` keep working, now with stable alphabetical order (matching the fallback log message).
- The PR body accurately describes the fixed problem: filesystem (`listFiles()`) order previously decided pack priority.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:
Why create a separate `resource_packs.yml`? We already have `packs.yml` handling CDN and encryption; I think we can just continue using `packs.yml`.

PR #910, opened 2026-09-03
fix(player): keep creative players out of unbroken floors

w1zardz:
Creative instant-break prediction can move the client down before an authoritative block update restores a block that was not actually removed: the player ends up inside the floor and stays there.

Probe the thin target-feet collision shape before `fastMove`, independently of the global `allow-flight` setting, and keep the last safe server position when that live collision remains.

Survival, spectators, vehicles, air, safe landings and side-only collisions keep their existing behaviour. Corrections use the existing version-stable NORMAL/RESET path, so nothing new goes on the wire.

Covered by `PlayerCreativeDownwardCollisionTest`.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -662,6 +662,11 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         return vanillaLimitSquared * speedRatio * speedRatio;
     }
 
+    static boolean shouldProbeCreativeDownwardCollision(
+            boolean creative, boolean noClip, boolean riding, double dy) {
+        return creative && !noClip && !riding && dy < -1.0E-7;
+    }
+
     public void setAllowModifyWorld(boolean value) {
         this.adventureSettings.set(Type.WORLD_IMMUTABLE, !value);
         this.adventureSettings.set(Type.MINE, value);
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
The guard is limited to non-mounted, non-no-clip creative players moving downward, preserving survival, spectator, vehicle, and air behavior. However, the contracted feet probe misses valid floor collisions near block edges, so the original clipping bug remains possible. The tests mock collision results and therefore do not exercise this geometry.

### Issues Found
#### Blocking Issues
- `src/main/java/cn/nukkit/Player.java:2451` — Shrinking the target footprint by `0.05` can hide a real supporting-floor collision. For example, a player standing near a block edge may overlap the floor by less than 0.05 blocks; the full player bounding box collides after the predicted downward move, but `targetFeet` does not. `fastMove` then either bypasses collision under global `allow-flight` or has its correction accepted by `handleMovement`, leaving the player inside/below the live floor — the exact bug this change intends to prevent.

#### Non-blocking Issues
- None

### Recommendations
- Detect a vertical floor obstruction using the player's full horizontal footprint, while distinguishing it from lateral collisions through collision-shape position/direction rather than a fixed horizontal inset.
- Add a test with real or captured block/AABB geometry for a player supported near a block edge. The current `Level.hasCollision` stubs at `src/test/java/cn/nukkit/PlayerCreativeDownwardCollisionTest.java:157` return synthetic booleans and cannot reveal this false negative.

### Notes
- Verdict: **NEEDS CHANGES**

---
🤖 AI Review

w1zardz:
Superseded by #913. This implementation probes every tiny downward creative movement and reverts the full XYZ frame, which causes horizontal rubber-banding on partial blocks because normal Bedrock float round-trips can produce a small negative Y delta. #913 scopes the correction to an actually denied support break, sweeps the attempted vertical path, and preserves X/Z movement.

PR #911, opened 2026-09-03
fix(block): match Bedrock collision boxes for nine more blocks

w1zardz:
A server-side collision box that is taller than the box the client draws makes every creative or flying player collide with geometry he cannot see and get corrected back frame after frame. Eight blocks were still on Java values or on a plain full cube:

* **dirt path** and **farmland** — client walks one pixel below the top;
* **soul sand** — two pixels below the top;
* **stonecutter** — nine pixels high;
* **hopper** — has a bowl a player drops into;
* **snow layer** — steps two pixels per layer; the server had zero up to two layers, half a block up to seven and a full block only at eight;
* **thin blocks (panes/bars)** with no neighbours — fell into the branch that widens the box to the whole cell instead of keeping the two-pixel centre post;
* **stairs** — exposed only the lower slab through `getBoundingBox`, so the movement resolver never saw the step half.

Each shape is covered by a regression in `BlockCollisionShapeTest`.

⚠️ Stacked on #904 (enchanting table and cauldron collisions), whose commit introduces that test class and is the first commit here. Happy to rebase to a single commit once #904 is merged.

### DIFF
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -1157,6 +1157,15 @@ public abstract class Block extends Position implements Metadatable, Cloneable, 
         return collidesWithBB(bb, false);
     }
 
+    public void addCollisionBoxesToList(AxisAlignedBB bb, List<AxisAlignedBB> collidingBoxes) {
+        if (this.collidesWithBB(bb)) {
+            AxisAlignedBB boundingBox = this.getBoundingBox();
+            if (boundingBox != null) {
+                collidingBoxes.add(boundingBox);
+            }
+        }
+    }
+
     public boolean collidesWithBB(AxisAlignedBB bb, boolean collisionBB) {
         AxisAlignedBB bb1 = collisionBB ? this.getCollisionBoundingBox() : this.getBoundingBox();
         return bb1 != null && bb.intersectsWith(bb1);
--- a/src/main/java/cn/nukkit/block/BlockCauldron.java
+++ b/src/main/java/cn/nukkit/block/BlockCauldron.java
@@ -12,14 +12,17 @@ import cn.nukkit.level.biome.Biome;
 import cn.nukkit.level.particle.SmokeParticle;
 import cn.nukkit.level.vibration.VibrationEvent;
 import cn.nukkit.level.vibration.VibrationType;
+import cn.nukkit.math.AxisAlignedBB;
 import cn.nukkit.math.BlockFace;
 import cn.nukkit.math.MathHelper;
+import cn.nukkit.math.SimpleAxisAlignedBB;
 import cn.nukkit.nbt.tag.CompoundTag;
 import cn.nukkit.nbt.tag.Tag;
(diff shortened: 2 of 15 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR aligns server-side collision boxes with what the Bedrock client renders for eight block families: dirt path/farmland (15/16), soul sand (14/16), stonecutter (9/16), hopper (bowl + rim), snow layer (2px per layer), lone thin panes/bars (2-pixel centre post), and stairs (both slab and step halves exposed via a new `addCollisionBoxesToList` hook). The `CollisionHelper` now collects multi-part boxes instead of a single bounding box. All blocks that override `collidesWithBB` also override `addCollisionBoxesToList`, so the new code path stays consistent. The new regression suite `BlockCollisionShapeTest` passes locally (10/10).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `BlockHopper.java` / `BlockStairs.java`: newly added comments are written in Russian; the codebase convention is English.
- `BlockHopper` / `BlockCauldron`: `recalculateCollisionBoxes()` allocates 5 `SimpleAxisAlignedBB` objects on every collision query. Hoppers are common in farms and this runs on the movement hot path; caching per block position/state would reduce allocation pressure.
- `BlockStairs.collisionParts()` duplicates the same geometry already encoded in `collidesWithBB`; the two could share a single source of truth to avoid future drift.

### Recommendations
- Translate the Russian comments to English.
- Consider caching the collision box array for hopper/cauldron (e.g. lazily, keyed by block state) instead of reallocating per query.
- This PR is stacked on #904 (the first commit, enchanting table + cauldron, belongs to #904) — merge only after #904 lands, or rebase to a single commit as the author offered.

### Notes
- The plain `collidesWithBB(bb)` override for hopper/cauldron intentionally changes behaviour for other call sites (`Player` in-block check, `EntityBoat`), making them consistent with the multi-part shape — this matches the PR's goal of matching client-side geometry.
- Snow layer with a single layer (meta 0) now has zero collision height and `canPassThrough() == true`, matching Bedrock behaviour.
- The `BlockThin` change only alters the "no neighbours" case; all connected cases produce identical boxes to before.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

w1zardz:
Added one more block to this PR: the shelf.

A shelf occupies five pixels next to its wall: a lower board across the whole depth, a three-pixel back panel and a top board above the niche. The server used a full cube, so one player stood on a shelf as if it were a solid block while another fell through it, and every step near a shelf was corrected against geometry the client cannot see.

The new commit gives `BlockShelf` its composite collision boxes through the same `addCollisionBoxesToList` path as the other blocks here, keeps the bounding box on the band the shelf really occupies, and adds `BlockShelfCollisionTest` for both the free space in front of it and its boards.

lt-name:
Superseded by https://github.com/MemoriesOfTime/Nukkit-MOT/pull/967

PR #912, opened 2026-09-03
fix(inventory): move what fits when a stack merge overflows

w1zardz:
## Problem

A `TAKE`/`PLACE` whose destination cannot hold the whole moved amount is answered with `ERROR`, and nothing is moved:

```java
int destCount = destItem.isNull() ? 0 : destItem.getCount();
if (destCount + count > sourceItem.getMaxStackSize()) {
    return context.error();
}
```

The client rolls such a request back, so both slots stay exactly as they were. To the player the two stacks simply refuse to merge.

It shows up first on items with a small stack limit — ender pearls, eggs, snowballs, honey bottles — where a merge overflows almost every time. It was reported on a production server as "ender pearls do not stack": the stacks were byte-identical (same id, same damage, no NBT), they just could never be merged.

Reproduced against `master` by calling the processor directly:

```
15 + 2, requested 2 -> success=false, slots stay 15 and 2
15 + 2, requested 1 -> success=true,  slots become 16 and 1
 4 + 3, requested 3 -> success=true,  slots become 7 and empty
```

## Fix

Vanilla moves as much as the destination can hold and leaves the remainder on the source side. `TransferItemActionProcessor.doTransfer` now does the same:

* clamp the transferred count to the room left in the destination (`min(source max stack size, destination inventory max stack size)`);
* recompute `fullTransfer` after the clamp, so a partial move no longer clears the source slot;
* keep answering `ERROR` only when the destination is already full and nothing can move at all.

Amounts are conserved: the source loses exactly what the destination gains. The response echoes both containers from the real inventory state, so the client resynchronises to the partial result.

## Tests

`PartialStackMergeTest` covers the three cases: an overflowing merge fills the destination and keeps the remainder, a merge that fits moves everything, and a merge into a full stack is still refused with both slots untouched.

Full suite is green: `Tests run: 16129, Failures: 0, Errors: 0, Skipped: 2`.

### DIFF
--- a/src/main/java/cn/nukkit/inventory/request/TransferItemActionProcessor.java
+++ b/src/main/java/cn/nukkit/inventory/request/TransferItemActionProcessor.java
@@ -91,9 +91,19 @@ public abstract class TransferItemActionProcessor<T extends TransferItemStackReq
             return context.error();
         }
 
+        // A merge the destination cannot hold in full is vanilla behaviour, not an error: the
+        // client moves as much as fits and keeps the remainder on the source side. Refusing the
+        // whole request instead makes stacks with a small limit — ender pearls, eggs, snowballs —
+        // look like they cannot be merged at all, because the client rolls the request back and
+        // both slots stay as they were.
         int destCount = destItem.isNull() ? 0 : destItem.getCount();
-        if (destCount + count > sourceItem.getMaxStackSize()) {
+        int stackLimit = Math.min(sourceItem.getMaxStackSize(), dstInv.getMaxStackSize());
+        if (destCount >= stackLimit) {
             return context.error();
+        }
+        if (destCount + count > stackLimit) {
+            count = stackLimit - destCount;
+            fullTransfer = sourceItem.getCount() == count;
         }
 
         if (!isSlotCompatible(dstInv, dstSlot, sourceItem)) {
--- a/src/test/java/cn/nukkit/inventory/request/PartialStackMergeTest.java
+++ b/src/test/java/cn/nukkit/inventory/request/PartialStackMergeTest.java
@@ -1,0 +1,132 @@
+package cn.nukkit.inventory.request;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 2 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Fixes a real vanilla-parity bug: a TAKE/PLACE whose destination cannot hold the whole stack was answered with ERROR, so the client rolled the request back and stacks with small limits (ender pearls, eggs, snowballs) appeared unmergeable. The fix clamps the transferred count to the room left in the destination (`min(item max stack size, destination inventory max stack size)`), recomputes `fullTransfer` after the clamp, and only returns ERROR when the destination is already full. Amounts are conserved, the stackNetId allocation strategy stays consistent (merge into a non-empty slot keeps the dest id; partial move to an empty slot assigns a fresh id), and the response echoes both containers from real inventory state so the client resynchronises to the partial result. Verified locally: `PartialStackMergeTest` — Tests run: 3, Failures: 0, Errors: 0. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- In the clamp branch, `fullTransfer = sourceItem.getCount() == count` is provably always false (clamped count < requested count <= source count), so it could be simplified to `fullTransfer = false`. Harmless as-is; the general form just never evaluates to true.
- The clamp also applies when placing into an empty slot with a requested count above the stack limit (previously an ERROR, now a partial move). This is vanilla-consistent, but it is a behaviour change beyond the merge scenario — worth confirming it is intentional.

### Recommendations
- Consider adding a test case for the empty-destination clamp path (requested count above the stack limit) to lock in that behaviour.

### Notes
- Creative CREATED_OUTPUT transfers return before the clamp and are unaffected.
- Over-full destination stacks (destCount >= stackLimit) are still refused, matching previous behaviour for that case.

---
@lt-name ready to merge

🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #976, which contains the same changes (the partial stack merge overflow handling) along with additional fixes, rebased onto the current master.

PR #913, opened 2026-09-04
Prevent clipping after denied creative breaks without repeated floor corrections

w1zardz:
## Problem

A creative client predicts an instant block break before the server validates it. When interaction checks or a block-break event deny the break, the restored support can disagree with movement that the client has already sent. Reverting an entire movement frame causes horizontal rubber-banding, while restoring the previous airborne Y can leave the player suspended above a partial block.

Resting contact can also place Bedrock's reported feet about 0.005 blocks below the real surface. Repeatedly correcting that harmless offset creates a correction/acknowledgement loop and can reject the next jump.

## Changes

- Arm a five-tick correction only after a denied creative break below the player, covering both server-authoritative breaking and the legacy inventory transaction path.
- Preserve the existing level, expiry, distance, creative-mode, riding, no-clip, and live-support checks. Unrelated denied breaks cannot replace the pending support.
- Resolve the crossed top using the denied block's actual collision predicate, including both halves of stairs. The search is bounded to 40 probes of that one block; ordinary creative movement does not gain a world-wide floor sweep.
- Land on the actual crossed surface while preserving accepted X/Z. Check horizontal overlap at the crossing instant so diagonal movement below an overhang is accepted. A taller replacement retains the previous protection without pushing the player upward.
- Normalize small downward contact locally so it cannot accumulate below the authoritative floor. The 0.01-block tolerance controls correction packets only and does not enlarge the geometric support search.
- Let an exactly stationary movement packet acknowledge pending corrections in both input formats, allowing the next jump.

`SkinDimensionValidationTest` initializes its mock server explicitly to remove its existing dependency on test order.

## Validation

`mvn -q -B -Dtest=SkinDimensionValidationTest,PlayerFlightMovementLimitTest,PlayerFallCheckTest,PlayerDeniedCreativeBreakContactTest test`

59 tests passed with zero failures or errors. Coverage includes real trapdoors, end portal frames, lower/upper slabs and both stair heights; one-packet falls; horizontal edge exits; diagonal overhang and pit-rim movement; recorded resting-contact cycles; penetration accumulation; stationary acknowledgements; and the original denied-break guard regressions. Packet tests cover protocols 1.20.0, 1.21.90, 1.26.40 and 1.26.45 across five protocol/movement-mode combinations.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -155,6 +155,9 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     public static final int SPECTATOR = 3;
     public static final int VIEW = SPECTATOR;
 
+    private static final double CREATIVE_BREAK_POSITION_EPSILON = 1.0E-4;
+    private static final int CREATIVE_BREAK_CORRECTION_TICKS = 5;
+
     public static final int CRAFTING_SMALL = 0;
     public static final int CRAFTING_BIG = 1;
     public static final int ANVIL_WINDOW_ID = 2;
@@ -256,6 +259,10 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     protected long randomClientId;
 
     protected Location forceMovement = null;
+
+    private Level deniedCreativeBreakLevel;
+    private BlockVector3 deniedCreativeBreakBlock;
+    private int deniedCreativeBreakExpiresAtTick;
 
     protected Location teleportPosition = null;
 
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a server-authoritative movement exploit where a creative client predicts an instant block break, and if the break is denied, a movement packet can place the player below the still-existing support block (clipping through floors).

The fix arms a 5-tick Y-clamp correction only after a denied creative break below the player, covering both the server-authoritative block-breaking path and the legacy inventory-transaction path. It revalidates block collision, level, distance, and expiry before correcting, sweeps the attempted vertical path so a single packet cannot skip below the support, clamps only Y while preserving accepted X/Z movement, and ignores float round-trip noise via a small position epsilon.

The approach is well-scoped and avoids a blanket movement rollback (which would cause horizontal rubber-banding on partial blocks). Regression coverage is comprehensive: partial-block horizontal movement, correction acknowledgement, float noise, one-packet vertical skips, re-arm attempts, block replacement, expiry, distance, and non-creative mode. All targeted tests pass locally (`PlayerDeniedCreativeBreakMovementTest`, `PlayerFlightMovementLimitTest`, `PlayerFallCheckTest`).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- The extreme-distance check and food exhaustion in `handleMovement` are still computed from the raw (uncorrected) client position. A heavily clamped vertical skip is still charged exhaustion for the full raw distance. Impact is negligible, but worth being aware of.
- The `0.00001` Y offset used to force the correction packet is a magic number; a named constant (similar to `CREATIVE_BREAK_POSITION_EPSILON`) would improve readability.

### Recommendations
- Consider extracting the `0.00001` correction offset into a named constant in a follow-up.

### Notes
- Correction state tracks a single block and is overwritten on re-arm; this is covered by tests and acceptable for the targeted scenario.
- State is properly invalidated on gamemode change, level change, riding, no-clip, block replacement, and 5-tick expiry.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

w1zardz:
This supersedes #910. The replacement avoids its global creative-movement probe and full-frame rollback; correction is armed only after a denied support break and clamps Y while preserving X/Z.

PR #919, opened 2026-09-05
fix(player): import PocketMine-MP off-hand and ender chest data

w1zardz:
## Problem

A player profile written by PocketMine-MP stores the off-hand item as a separate `OffHandItem` compound and the ender chest as an `EnderChestInventory` list. Nukkit reads the off-hand from slot `-106` of the shared `Inventory` list and the ender chest from `EnderItems`. The main inventory shares the same layout, so a world migrated from PocketMine-MP loads with hotbar and armour intact while the off-hand item and the whole ender chest silently come up empty. Nothing is logged, and the original data is still in the profile.

## Fix

Translate both PocketMine-MP keys before the inventories are loaded:

* an `OffHandItem` compound becomes slot `-106` of `Inventory`;
* an `EnderChestInventory` list becomes `EnderItems`.

When the profile already carries a Nukkit value for a slot, the Nukkit slot wins and the legacy item is moved to a free slot of the ender chest, then of the main inventory; if there is no room anywhere, the legacy key is left in place so nothing is lost. Once translated, the legacy keys are removed, so a second login does not duplicate anything.

## Tests

`PocketMinePlayerInventoryImportTest` covers the plain import, a conflict on the off-hand slot, a full ender chest, an overflow into the main inventory, and a repeated login.

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityHumanType.java
+++ b/src/main/java/cn/nukkit/entity/EntityHumanType.java
@@ -24,6 +24,12 @@ import java.util.List;
 
 public abstract class EntityHumanType extends EntityCreature implements InventoryHolder {
 
+    private static final String POCKETMINE_OFFHAND = "OffHandItem";
+    private static final String POCKETMINE_ENDER_CHEST = "EnderChestInventory";
+    /** Host-side durable storage must acknowledge the imported ender chest before clearing this. */
+    public static final String POCKETMINE_ENDER_IMPORT_PENDING =
+            "NukkitPocketMineEnderImportPending";
+
     protected PlayerInventory inventory;
     protected PlayerEnderChestInventory enderChestInventory;
     protected PlayerOffhandInventory offhandInventory;
@@ -47,6 +53,8 @@ public abstract class EntityHumanType extends EntityCreature implements Inventor
 
     @Override
     protected void initEntity() {
+        importPocketMineInventories(this.namedTag);
+
         this.inventory = new PlayerInventory(this);
         this.offhandInventory = new PlayerOffhandInventory(this);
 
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes silent data loss when loading a PocketMine-MP player profile: the off-hand item (`OffHandItem` compound) and ender chest (`EnderChestInventory` list) are now translated into Nukkit's layout (slot `-106` of `Inventory` and `EnderItems`) before the inventories are loaded. The conversion is slot-conflict aware (existing Nukkit items win, legacy items overflow to the first free slot), retains untranslatable legacy tags when everything is full so nothing is lost, and removes imported tags so a second login does not duplicate items. All 5 new tests pass locally (`mvn test -Dtest=PocketMinePlayerInventoryImportTest`).

The slot arithmetic was verified against the loading path in `initEntity()`: NBT slots 9–44 map to in-memory indices 0–35, which exactly matches the `firstEmptySlot(inventory, 9, 45)` overflow range, and slot `-106` as a byte is valid.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `POCKETMINE_ENDER_IMPORT_PENDING` is written into player NBT but nothing in the codebase ever reads or clears it, so the flag persists in player data forever. The javadoc references a "host-side durable storage" contract that does not exist in this repo. Consider documenting the intended consumer or dropping the flag.
- When the off-hand slot `-106` is occupied, the legacy item overflows to the main inventory only; the ender chest is never considered as a fallback. Behavior is safe, just worth noting as a deliberate choice.

### Recommendations
- Clarify or remove the unused pending-import flag to avoid permanently polluting player NBT.
- Optionally add a debug log line when legacy tags are retained due to full inventories, so server admins can tell why an item "didn't move yet".

### Notes
- Conversion is idempotent: imported tags are removed, verified by `importIsIdempotent`.
- Retained legacy items keep their original `Slot` values and are re-attempted on the next login — no data loss path identified.
- `ListTag.getAll()` returns a copy, so the remove-while-iterate pattern in `removeEmptyItemsAt` is safe.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #921, opened 2026-09-05
fix(entity): replenish absorption when a stronger effect remains

w1zardz:
## Problem

`Entity.addEffect` returns early when an effect with a higher amplifier or a longer duration is already active. For Absorption that skipped the refill as well: a player still showing enchanted-apple Absorption IV who ate a normal golden apple after the yellow hearts had been used up received no absorption at all, because `Effect.add` is what refills the pool and it never ran.

Bedrock keeps the stronger icon and its duration but still lets the consumed food replenish the hearts it grants.

## Fix

When a food-caused Absorption effect is rejected in favour of a stronger one, raise the absorption to what the consumed item grants (`(amplifier + 1) * 4`) if the current value is lower. The existing effect, its amplifier and its duration stay untouched, and the current absorption is never lowered. Only `EntityPotionEffectEvent.Cause.FOOD` triggers this; a rejected potion or command effect still does nothing.

## Tests

Three cases in `ArmorDamageReductionTest`: a weaker apple refills its four points under Absorption IV, an already higher pool is not reduced, and a non-food cause does not refill.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -1092,6 +1092,18 @@ public abstract class Entity extends Location implements Metadatable {
 
         if (oldEffect != null && (oldEffect.getAmplifier() > effect.getAmplifier()
             || (oldEffect.getAmplifier() == effect.getAmplifier() && oldEffect.getDuration() >= effect.getDuration()))) {
+            // A weaker/shorter absorption effect must not replace the stronger icon or its
+            // duration, but consuming the item still replenishes the amount granted by that
+            // item. Without this, a normal golden apple eaten while enchanted-apple
+            // Absorption IV is still visible gives zero yellow hearts after the old pool was
+            // depleted: addEffect returns before Effect.add can refill anything.
+            if (effect.getId() == Effect.ABSORPTION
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

When a food-caused Absorption effect is rejected in favor of a stronger/longer active effect, the entity now still receives the absorption points granted by the consumed item (`(amplifier + 1) * 4`), matching Bedrock behavior. The stronger effect's icon, amplifier, and duration remain untouched, and the current absorption is never lowered. The granted-amount formula is identical to the one used by `Effect.add`, and the change is scoped to `EntityPotionEffectEvent.Cause.FOOD` only. Three new tests cover the refill case, the no-reduction case, and the non-food rejection case. Local run: `ArmorDamageReductionTest` — 37 tests, 0 failures.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- None

### Notes
- The refill path does not fire `EntityPotionEffectEvent` (consistent with the existing early-return behavior for rejected effects), so plugins cannot cancel the refill. Acceptable since it mirrors vanilla Bedrock behavior.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #922, opened 2026-09-05
fix(resourcepack): serve one immutable archive per runtime

w1zardz:
Deployment replaces resource-pack archives while gameplay servers wait for
their nightly restart. ZippedResourcePack cached the manifest at startup but
read size, SHA-256 and every chunk from the live pathname later. A replacement
therefore mixed the old identity with new bytes, and clients split between a
stale cache and an invalid download.

Snapshot each small archive when the runtime loads it. The running process now
keeps serving one internally consistent pack; the next process picks up the
new archive after the normal restart.

### DIFF
--- a/src/main/java/cn/nukkit/resourcepacks/AbstractResourcePack.java
+++ b/src/main/java/cn/nukkit/resourcepacks/AbstractResourcePack.java
@@ -19,13 +19,19 @@ public abstract class AbstractResourcePack implements ResourcePack {
 
     /**
      * 除了字段存在性，还校验后续解析无条件读取的字段类型（getPackId、getPackVersion、
-     * ZippedBehaviourPack 的 modules 扫描）：构造成功之后的抛异常路径没有清理入口，
-     * 会泄漏快照 channel，所以凡是能通过本校验的 manifest 后续解析绝不能抛异常。
+     * getPackProtocol 的 min_engine_version、ZippedBehaviourPack 的 modules 扫描）：
+     * 构造成功之后的抛异常路径没有清理入口，会泄漏快照 channel，所以凡是能通过本校验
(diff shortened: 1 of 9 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a real consistency bug: `ZippedResourcePack` previously parsed the manifest once at load time but served size, SHA-256 and every chunk from the live file path. If a deployment replaced the archive while the server was running, the server would mix the old identity (manifest/sha) with new bytes, and clients would reject the download. The change snapshots the archive into memory at load time and serves everything (manifest, size, sha256, chunks) from that single immutable byte array. The semantics of manifest lookup priority (root `manifest.json` > root `pack_manifest.json` > first nested manifest at depth ≤ 1) are preserved, and `getPackChunk` behavior is equivalent for all valid offsets. A regression test covering the replace-after-load scenario is included.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- Memory trade-off: the entire archive is now held in heap for the lifetime of the runtime. This is the intended design ("one immutable archive per runtime"), but servers with many large packs will see increased RSS. Given resource packs are typically small, this is acceptable — worth documenting if very large packs are expected.
- `JsonParser().parse(...)` uses a deprecated Gson API (pre-existing, untouched by this PR's intent).

### Recommendations
- Optionally log the pack's byte size at load time to make the per-runtime snapshot memory footprint observable.
- Consider a short comment on `packBytes` noting it is the single source of truth for the runtime's lifetime.

### Notes
- Verified `findManifest` preserves the old lookup priority across all entry orderings.
- `getPackChunk` edge cases (off == size → empty array; off > size → exception, same as before) match previous behavior.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #924, opened 2026-09-05
fix(dispenser): place dyed shulker boxes and keep the slot when a behavior declines

w1zardz:
behavior declines

A dyed shulker box (minecraft:shulker_box, id 218) in a dispenser was consumed without ever
being placed: ShulkerBoxDispenseBehavior created the block entity with the box contents on
an air cell and never wrote the block. The slot lost the box, the entity had no block under
it, and the next garbage collection or block placement discarded it together with up to 27
stacks inside. The undyed box already did this correctly; the dyed one now mirrors it,
including the collision check that used to run against an unpositioned block at 0,0,0.

BlockDispenser.dispense also decremented the count of the exact object it handed to the
behavior. Behaviors that decline (undyed shulker box facing a wall, water bottle on dirt
that refused, shears with nothing to shear) return that same object to say 'put it back',
so the slot came back one item short: an undyed shulker box fired at a wall vanished.
The behavior now receives its own copy; the same-id-and-damage branch keeps working for
behaviors that return a damaged copy (shears).

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockDispenser.java
+++ b/src/main/java/cn/nukkit/block/BlockDispenser.java
@@ -230,7 +230,10 @@ public class BlockDispenser extends BlockSolidMeta implements Faceable, BlockEnt
         original = original.clone();
 
         DispenseBehavior behavior = DispenseBehaviorRegister.getBehavior(original);
-        Item result = behavior.dispense(this, facing, original);
+        // The behavior gets its own copy. Several behaviors answer "nothing happened, put it
+        // back" by returning the very item they were given; when that was the same object whose
+        // count is decremented below, the slot lost one item without dispensing anything.
+        Item result = behavior.dispense(this, facing, original.clone());
 
         pk.evid = LevelEventPacket.EVENT_SOUND_CLICK;
 
--- a/src/main/java/cn/nukkit/dispenser/FlintAndSteelDispenseBehavior.java
+++ b/src/main/java/cn/nukkit/dispenser/FlintAndSteelDispenseBehavior.java
@@ -18,14 +18,18 @@ public class FlintAndSteelDispenseBehavior extends DefaultDispenseBehavior {
                 boolean soulFire = down.getId() == Block.SOUL_SAND || down.getId() == Block.SOUL_SOIL;
                 block.level.setBlock(target, Block.get(soulFire ? BlockID.SOUL_FIRE : BlockID.FIRE));
             }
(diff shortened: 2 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Two real fixes: dyed shulker boxes dispensed from a dispenser were never placed as a block (the block entity was created on an air cell and later discarded with up to 27 stacks inside), and behaviors that decline by returning the passed item caused the slot to lose one item. The shulker box fix looks correct — the block is now written with `setBlock`, the color comes from `item.getDamage()`, and the collision check runs against a positioned block instead of one at 0,0,0. However, the `BlockDispenser` change introduces a regression elsewhere.

### Issues Found
#### Blocking Issues
- **`BlockDispenser.dispense`: handing the behavior its own copy breaks behaviors that mutate the passed item in place and return `null`.** `FlintAndSteelDispenseBehavior` calls `item.useOn(target)` (which does `this.meta++` on the item) and returns `null`. Before this PR the item it received was the same object later written back to the slot, so the durability loss persisted. After this PR the damage is applied to a discarded clone: a flint and steel fired from a dispenser now never loses durability and never breaks, which deviates from vanilla behavior and is an infinite-use exploit. Other behaviors are safe (`ShearsDispenseBehaviour` clones internally and returns the damaged copy; `ProjectileDispenseBehavior`'s trident hands the damaged item to the thrown entity), but `FlintAndSteelDispenseBehavior` must be adjusted in the same PR — e.g. have it return the damaged item so the dispenser's result handling writes it back (note its damage differs from the origin, so it will take the `addItem` path; verify this doesn't duplicate the tool when the stack count is 1).

#### Non-blocking Issues
- The `success` flag is left untouched when the shulker box placement is declined due to a non-air target (it keeps whatever value it had). Consider setting `this.success = false` on that early return, matching the semantics of "nothing was dispensed".

### Recommendations
- Fix the flint and steel durability regression before merge (see blocking issue).
- Add a brief regression test or at least a manual test matrix for dispenser behaviors that mutate the passed item: flint and steel, shears, trident, water bottle, buckets.

### Notes
- `mvn compile` passes with these changes.
- The shulker box block-entity path (`Block.get(BlockID.SHULKER_BOX, item.getDamage())` + positioned collision check + `setBlock`) mirrors the undyed variant and looks correct.

---
🤖 AI Review

tachibana-kanade-angel:

PR #925, opened 2026-09-05
fix(level): save changed neighbours together with a chunk saved on unload

w1zardz:
unload

A machine on a chunk border moves items between two chunks within one tick: a hopper feeding
a chest across the border, a double chest whose halves sit in different chunks, a hopper
chain. When one of those chunks was saved on unload while the other stayed loaded, the disk
held the two halves of one transfer from different moments. A crash before the next save then
duplicated or lost everything moved in between: a hopper that had pushed a diamond into a
chest across the border came back after kill -9 with the diamond still inside and the chest
holding one too; a stack moved from the left half of a double chest to the right half came
back on both sides or on neither, depending on which chunk had been written last.

Level.unloadChunk now writes the changed neighbours of the unloading chunk in the same pass.
A neighbour that did not change is skipped, so the extra work is bounded by what the unload
queue would have written moments later anyway. The autosave path already writes every changed
chunk in one loop and is unchanged.

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -4486,6 +4486,7 @@ public class Level implements ChunkManager, Metadatable {
                     if (chunk.hasChanged()) {
                         levelProvider.setChunk(x, z, chunk);
                         levelProvider.saveChunk(x, z);
+                        this.saveChangedNeighbours(levelProvider, x, z);
                     }
                 }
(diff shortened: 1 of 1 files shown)


tachibana-kanade-angel:
This PR has been superseded by #979, which contains the same changes (the neighbour-chunk save on unload) along with additional fixes, rebased onto the current master.

Closing this PR to avoid double-applying these changes. If any part of this PR is not covered by #979, feel free to reopen or comment and we will split it out.

PR #926, opened 2026-09-05
fix(item): track the goat horn cool down on the player

w1zardz:
Two goat horns in the same inventory can be sounded back to back with no wait.

`ItemGoatHorn` kept its cool down in an `AtomicBoolean` field of the item itself, so the cool down belonged to one item stack instead of to the player: the second stack is a different `Item` object with its own flag. On top of that `BaseInventory#getItem` returns a clone, so the flag on the stack in the slot is not reliably the one that was flipped. `PlayerStartItemCoolDownPacket` does not save it either — it is a client-side display effect, and the server checked nothing.

Cool downs are a property of the player, so track them there:

* `Player#setItemCoolDown` records the end tick per item category next to the packet it already sent (and a cool down of `0` or less clears it);
* new `Player#isItemCoolDownEnd(String)` / `Player#getItemCoolDownEnd(String)` answer whether the category may be used again;
* `ItemGoatHorn#onClickAir` asks the player instead of its own field, which also drops the delayed task that was scheduled on every single use.

Clients below `v1_18_10` still get no display packet, but the server-side cool down now applies to them as well.

**How to reproduce on master:** put two goat horns in the hotbar, sound the first one, switch slot and sound the second — it plays immediately instead of waiting the 7 s cool down.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -310,7 +310,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     protected final Map<UUID, Player> hiddenPlayers = new HashMap<>();
     /** Server tick at which the cool down of an item category ends. */
-    protected final Map<String, Integer> itemCoolDownEnds = new HashMap<>(2);
+    protected final Map<String, Integer> itemCoolDownEnds = new ConcurrentHashMap<>(2);
 
     /**
      * 已向本观察者下发 PlayerList(ADD) 的玩家型实体 UUID，用于去重防网易客户端隐形；
@@ -6452,7 +6452,8 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             return 0;
         }
         if (this.server.getTick() >= end) {
-            this.itemCoolDownEnds.remove(itemCategory);
+            // remove(key, value): an unconditional remove could clobber a cool down recorded concurrently
+            this.itemCoolDownEnds.remove(itemCategory, end);
             return 0;
         }
         return end;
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Moves the goat horn cool down from a per-item `AtomicBoolean` (broken by item-stack cloning and by carrying two horns) to per-player tracking: `Player#setItemCoolDown` now records the end tick server side, and `ItemGoatHorn#onClickAir` queries the new `Player#isItemCoolDownEnd`. This also removes the delayed task scheduled on every use. Two files, +58/-15. Verified compiling cleanly with `mvn compile`. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- `getItemCoolDownEnd` lazily evicts expired entries inside a getter — acceptable, slightly surprising for a query-style method.
- Cool down state is per player session and not persisted across re-login; consistent with the intent of the fix, noted for completeness.

### Recommendations
- Follow-up: other vanilla items with shared cool downs (e.g. ender pearl, chorus fruit) could adopt `isItemCoolDownEnd` for server-side enforcement.

### Notes
- `setItemCoolDown` semantics changed from "client-side display only" to also recording server-side state. The only other existing caller (shield break in `Player`) merely gains a tracked entry — nothing queries the `shield` category, so no behavior change there.
- Clients below `v1_18_10` receive no display packet but are now covered by the server-side cool down, as described in the PR.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #929, opened 2026-09-05
feat(network): add NetEase zlibstream compression

CDS114514:
This change adds support for NetEase's proprietary "zlibstream" compression (protocol algorithm 2), controlled by the new use-netease-zlibstream-compression setting. zlibstream is a compression scheme unique to the NetEase ecosystem and is fundamentally different from the per-frame compression used in standard Bedrock — the two wire formats are incompatible and cannot be mixed, and this server previously did not support it.

zlibstream treats the data of an entire session as a single never-ending raw deflate stream: compression is performed without a zlib header or trailer, and each transmitted segment is finished with a SYNC_FLUSH — the output is aligned to a byte boundary and decodable up to that point, yet the stream is never closed, so the sliding-window dictionary is shared across segments and later data can still reference earlier content for better compression. On the receive side, a single persistent Inflater is fed continuously and no end-of-stream marker is required. In the implementation, every session entering this mode keeps a persistent Deflater/Inflater pair, one for each direction; after compressing each segment, all pending output is fully drained so that one message always corresponds to exactly one complete batch plus a sync point. Sessions in this mode never carry a compression prefix byte, and the receive path handles the stream accordingly; pre-compressed cached payloads such as chunks and biome definitions are first restored to plaintext and then fed through the session stream so the peer's stream state is never corrupted.

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -578,6 +578,7 @@ public class Server {
      * Using Snappy compression
      */
     public boolean useSnappy;
+    public boolean useNeteaseZlibStream;
     /**
      * 1.19.30+ Using Client Spectator Mode
      * Because some servers may require the use of the inventory in spectator mode
@@ -3887,6 +3888,7 @@ public class Server {
         // Network
         this.networkCompressionThreshold = config.networkSettings().compressionThreshold();
         this.useSnappy = config.networkSettings().useSnappyCompression();
+        this.useNeteaseZlibStream = config.networkSettings().useNeteaseZlibStreamCompression();
         this.rakPacketLimit = config.networkSettings().rakPacketLimit();
         this.networkLoginTimeoutMilliseconds = config.networkSettings().timeoutMilliseconds();
         this.rakCookieMode = parseRakCookieMode(config.networkSettings().rakCookieMode());
--- a/src/main/java/cn/nukkit/network/NetEaseZlibStreamCompression.java
+++ b/src/main/java/cn/nukkit/network/NetEaseZlibStreamCompression.java
@@ -1,0 +1,63 @@
+package cn.nukkit.network;
+
+import cn.nukkit.utils.BinaryStream;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.zip.Deflater;
+import java.util.zip.Inflater;
+
+public class NetEaseZlibStreamCompression implements CompressionProvider {
+
+    private static final int CHUNK = 8192;
+    private static final int MAX_DECOMPRESSED_BYTES = 3 * 1024 * 1024;
+
+    private final Deflater deflater = new Deflater(7, true);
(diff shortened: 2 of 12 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds support for NetEase's proprietary zlibstream compression (algorithm 2): a persistent raw-deflate stream per session with SYNC_FLUSH after each segment, negotiated only for NetEase clients (raknet protocol 8, netEaseMode, protocol >= V1_20_50_NETEASE) and gated behind a new `use-netease-zlibstream-compression` config option (default off). The implementation is clean and well-scoped: the new `NetEaseZlibStreamCompression` provider keeps a persistent Deflater/Inflater pair, the v554 network-settings processor negotiates it correctly, prefixed compression is properly excluded for these sessions, and pre-compressed cached batch payloads are transparently re-inflated and re-compressed through the session stream so peer stream state is never corrupted. Compile verified (`mvn compile` passes, new class artifacts produced) and antivirus scan is clean. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `NetEaseZlibStreamCompression.compress(BinaryStream, int level)` ignores the `level` parameter; the Deflater is hardcoded to level 7, so the existing `networkCompressionLevel` setting has no effect in this mode.
- The per-session `Deflater`/`Inflater` are never `end()`ed; native zlib memory is held until GC. Consider releasing them when the session disconnects.
- `inflateStandaloneBatch` probes formats by trial (raw → zlib → snappy) and swallows `IOException`s between attempts. If none match (e.g. an unexpected uncompressed payload), the exception surfaces in `sendBatchPacket`, which logs an error and silently drops the batch. Probing by trial on non-deflate data also carries a small theoretical mis-detection risk; tracking the cached payload's original format would be more robust.
- `decompress` breaks out of the loop returning partial data if the inflater reports `needsDictionary()`; in practice this cannot happen on a healthy session stream, but a defensive check would make failures louder.
- `PacketCompressionAlgorithm.java` lost its trailing newline.

### Recommendations
- Either honor the `level` argument in `compress` or document that zlibstream mode always uses level 7.
- Add a `close()`/cleanup path for the Deflater/Inflater on session teardown.
- Consider logging at a clearer level (or failing the send explicitly) when a cached batch cannot be re-inflated, so dropped packets are easier to diagnose.

### Notes
- Codex review was unavailable (API balance), so this is a manual review based on the full diff; compile was verified in an isolated worktree.
- Security: no suspicious patterns; ClamAV scan of the diff is clean. The 3 MB default decompression cap matches the network batch limit, and the max-size guard against zip-bomb-style inflation is present.
- `CompressionProvider.from()` does not handle `ZLIBSTREAM`, but the PR correctly bypasses it; wire serialization uses the enum ordinal (2), matching the NetEase algorithm ID.

@lt-name 可以合并

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #931, opened 2026-09-08
fix: measure melee reach to large entities from their bounding boxes

w1zardz:
Melee interaction measures reach and facing against an entity's position. For large or scaled entities, a player can stand next to the visible body while that position remains beyond reach or behind the attacker, so the server silently rejects the hit.

Measure non-player targets to the nearest point on their bounding box from the attacker's eyes, using that same point for the facing check. Preserve the existing centre-based checks for players, plain positions, and entities without a bounding box; PvP reach is unchanged.

Validation: all 7 PlayerEntityReachTest cases pass, covering near/far faces, standing inside the body, facing away, vertical reach, player targets, and the fallback path.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3170,12 +3170,38 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     }
 
     private boolean canInteractEntity(Vector3 pos, double maxDistance) {
-        if (this.distanceSquared(pos) > Math.pow(maxDistance, 2)) {
+        double pointX = pos.x;
+        double pointZ = pos.z;
+        double distanceSquared;
+        if (pos instanceof Entity entity && !(entity instanceof Player) && entity.boundingBox != null) {
+            // A big body puts its visible head and tail far away from the centre point: the
+            // bounding box can extend well beyond the entity position. Measured from the
+            // centre, a hit on the head was farther than the
+            // reach even though the sword touched the model, and the facing check refused a
+            // player who faced the head while the centre was behind his back. The hit is
+            // measured to the nearest point of the bounding box instead; a player standing
+            // inside the body is at distance zero. Players keep the centre measurement so the
+            // PvP reach does not change by a single block.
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Melee reach and facing checks for non-player entities are now measured against the nearest point of the entity's bounding box instead of its centre position. This fixes the case where a player stands next to a large entity's visible body but the hit is silently rejected because the centre is out of reach or behind the attacker. Players, plain positions, and entities without a bounding box keep the original centre-based logic, so PvP reach is unchanged. Verified locally: full build succeeds and all 7 `PlayerEntityReachTest` cases pass (near/far faces, standing inside the body, facing away, vertical reach, player targets, fallback path). No blocking issues found; the changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- Vertical reach for non-player entities is now measured from the attacker's eye height to the bounding box, which slightly extends effective vertical reach compared to the old feet-to-position measurement. This appears intentional and more correct, but it is a subtle behaviour change for small entities too.
- `PlayerEntityReachTest` invokes the private `canInteractEntity` via reflection; a method rename would break the test at runtime rather than compile time. Acceptable for a private-method test, just brittle.

### Recommendations
- None. The comment block above the new branch explains the motivation well.

### Notes
- The automated reviewer backend was unavailable, so this review was performed manually against the full diff and a local build + test run (`mvn test -Dtest=PlayerEntityReachTest`, 7/7 passing).
- ClamAV scan could not complete due to host memory pressure; both changed files are plain Java source and were inspected manually — no suspicious patterns, no binary or archive files in the PR.
- `Math.pow(maxDistance, 2)` was replaced with `maxDistance * maxDistance` — equivalent and cheaper.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:
Superseded by https://github.com/MemoriesOfTime/Nukkit-MOT/pull/966

PR #932, opened 2026-09-08
fix: preserve rolled server logs beyond seven restarts per day

w1zardz:
The rolling file appender uses Log4j's default seven-index archive limit. More than seven restarts or rollovers in a day silently delete that day's earliest logs, even when total log storage is small.

Use monotonically increasing archive indices and prune compressed log archives by a 1 GB aggregate limit. The existing daily/startup triggers and file formats are unchanged; the deletion condition only matches *.log.gz directly under logs.

Validation: exercised the real Log4j rolling appender with nine rollovers in an isolated temporary directory. The original configuration retained 7 archives; the proposed configuration retained all 9. An unrelated file survived both runs. The 1 GB retention threshold is explicit configuration policy, not a protocol requirement.

### DIFF
--- a/src/main/resources/log4j2.xml
+++ b/src/main/resources/log4j2.xml
@@ -10,6 +10,16 @@
                 <TimeBasedTriggeringPolicy/>
                 <OnStartupTriggeringPolicy/>
             </Policies>
+            <!-- Keep every rolled log. The default strategy caps the index at 7 per day and
+                 renumbers on rollover, so the eighth restart of a day silently deletes the
+                 oldest archive. Prune only when the archives together exceed 1 GB. -->
+            <DefaultRolloverStrategy fileIndex="nomax">
+                <Delete basePath="logs" maxDepth="1">
+                    <IfFileName glob="*.log.gz">
+                        <IfAccumulatedFileSize exceeds="1 GB"/>
+                    </IfFileName>
+                </Delete>
+            </DefaultRolloverStrategy>
         </RollingRandomAccessFile>
     </Appenders>
     <Loggers>


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR modifies only `src/main/resources/log4j2.xml`, adding a `DefaultRolloverStrategy` with `fileIndex="nomax"` and a `Delete` action capped by a 1 GB accumulated size. This correctly addresses the Log4j2 default behavior of capping the rollover index at 7 per day, which silently deleted the earliest archives after the eighth restart.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- The configuration follows the official Log4j2 `DefaultRolloverStrategy`/`Delete` pattern: `fileIndex="nomax"` lifts the default index cap of 7, and the `Delete` action with `basePath="logs"`, `maxDepth="1"`, and glob `*.log.gz` matches the archive layout produced by `filePattern="logs/%d{yyyy-MM-dd}-%i.log.gz"`.
- The glob does not match the active `logs/server.log`, so the live log file cannot be pruned.
- The 1 GB aggregate retention threshold is explicit configuration policy; operators can adjust it to their needs.

### Notes
- Automated code review tooling was unavailable; this review was performed manually against the diff.
- Config-only change (no Java code), so no compile/test impact.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):
Approved: no blocking issues found.

tachibana-kanade-angel:

PR #933, opened 2026-09-08
fix: preserve active consumable use across premature repeated input

w1zardz:
Repeated right-click/use transactions before a consumable's use duration expires clear the player's use state before `Item.onUse()` rejects the premature completion. Starting to sprint and unrelated legacy player actions also clear that state, preventing the server's automatic completion from finishing the food or drink.

Preserve an active consumable's use timer across premature repeated clicks, sprint starts, and the player-action fall-through. Bows, crossbows, shields, and other items without a timed use duration keep their existing release behavior. Transactions received at or after the required duration still complete normally.

Validation: 12 focused tests passed (3 consumable-use predicate tests, 7 food tests, 2 swimming tests). The new unit cases check the pre-completion condition, completion boundary, and items released by the client; they do not simulate the full packet sequence.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1105,6 +1105,36 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         return this.getDataFlag(DATA_FLAGS, DATA_FLAG_ACTION) && this.startAction > -1;
     }
 
+    /**
+     * Whether a use transaction is only a repeat of a consumable that is still being used.
+     *
+     * <p>The client keeps sending {@code USE_ITEM_ACTION_CLICK_AIR} while the button is held down,
+     * and a player in a fight taps the button rather than holding it. Such a transaction arrives
+     * before the item is ready ({@code ticksUsed < useDuration}), so handing it to
+     * {@code Item#onUse} only gets the item refused - while the use state is already gone and the
+     * food is never eaten. Items without a use duration (bow, crossbow, shield) are released by the
+     * client and are not affected.
+     */
+    static boolean isRepeatedConsumableUse(int useDuration, int ticksUsed) {
+        return useDuration > 0 && ticksUsed < useDuration;
+    }
+
+    /**
+     * Whether a consumable is being used right now: food, a potion, milk or the ominous bottle.
+     *
+     * <p>Only these items carry a use duration and are finished by the server timer in
+     * {@link #processAutoCompletion()}. A bow, a crossbow or a shield keeps {@code useDuration == 0}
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes consumable items (food, potions, milk, ominous bottle) never finishing when the client sends repeated use transactions before the use duration expires, when sprint is toggled, or when unrelated player actions fall through to a use-state reset. The fix adds a shared predicate (`isRepeatedConsumableUse`) and a state check (`isConsumingItem()`) to guard the three premature `setUsingItem(false)` call sites, while leaving client-released items (bow, crossbow, shield, trident, spyglass) untouched.

The approach is sound and minimal. The `useDuration > 0 && canRelease()` combination matches exactly the consumable item classes (ItemEdible, ItemPotion, ItemBucket, ItemOminousBottle); all client-released items keep `useDuration == 0` and retain their existing behavior. The completion boundary (`ticksUsed >= useDuration`) still flows through the original `onUse()` path, and switching hotbar slots still clears the use state via `MobEquipmentProcessor`, so no stale use timer can leak across items.

Note: Codex was unavailable (API key rejected by provider), so this is a manual review of the full diff. Compilation was verified in an isolated worktree (4121 classes, new methods present) and a ClamAV scan of the patch is clean.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `isConsumingItem()` uses `getItemInHandFast()` while `processAutoCompletion()` uses `getItemInHand()` — a slight inconsistency with no semantic impact, but worth aligning for clarity.
- `isConsumingItem()` is a new public API method; consider adding a `@since` tag per project API documentation conventions.

### Recommendations
- The new unit tests only exercise the static predicate. The packet-level behavior (repeated click-air transactions, sprint toggling while eating) would benefit from an integration-level test if the test harness ever supports simulating that sequence — the PR description already discloses this limitation, which is appreciated.
- Consider a brief note in the changelog/release notes since this is a user-visible gameplay fix (eating interrupted by sprinting or repeated clicks).

### Notes
- Verified edge case: `ticksUsed == useDuration` is NOT treated as a repeat, so a transaction arriving exactly at the completion boundary finishes the item normally (covered by tests).
- Verified that the `PlayerActionPacket` fall-through and sprint-start guards only skip the reset while a consumable is actively in use; all other item types behave as before.
- Local test run was skipped because the host is currently saturated by concurrent Maven builds; the change compiles cleanly and the predicate logic was verified by inspection.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #934, opened 2026-09-08
fix: replay block destruction endings before starting the next block

w1zardz:
`PlayerAuthInputPacket` receives a sequence of block actions but stores it in an `EnumMap`. This loses all but the last action of each type and reorders the remaining actions by enum ordinal. Multiple predicted creative breaks in one input tick can therefore disappear before reaching the level's authoritative checks. Starting the next block before completing the previous one can also reset the dig timer and cause a false fast-break rejection.

Keep every decoded action in wire order and replay that sequence through the existing player handler. Each prediction still reaches the normal interaction, game-mode, and level checks; rejected predictions restore the corresponding block. Destruction-ending actions at another position no longer start a new dig timer.

Retain the legacy map API for plugins and manually constructed packets. A deep snapshot detects map filters, replacements, action changes, and coordinate mutations from packet listeners. If the map was changed, replay its current contents using the completion-before-start ordering repair; untouched decoded packets use their original sequence. The new decoded-sequence getter returns a read-only defensive snapshot; mutable packet filtering remains on the legacy map API. No outgoing packet layout changes.

Validation: reference-encoded packets exercise the real player handler on protocols 589, 844, 2168, and 2169, covering duplicate predictions, rejection of either position, interleaved starts and completions, manually populated maps, six kinds of receive-listener map edits, and defensive snapshot isolation. The five existing ordering tests remain in place. Protocol 2169 decoding is checked against the shared 2168 auth-input wire shape because the reference codec currently exposes 2168.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4064,8 +4064,16 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                     return;
                 }
 
-                if (!authPacket.getBlockActionData().isEmpty()) {
-                    for (PlayerBlockActionData action : orderBlockActions(authPacket.getBlockActionData().values())) {
+                // A creative client can finish several blocks in one input tick. Replaying the
+                // legacy map silently loses all but the last action of each type, leaving blocks
+                // hidden on the client without ever asking the level to break or restore them.
+                // Only map-only packets need the ordering repair; decoded actions have wire order.
+                List<PlayerBlockActionData> blockActions = authPacket.getDecodedBlockActions();
+                if (blockActions.isEmpty()) {
+                    blockActions = orderBlockActions(authPacket.getBlockActionData().values());
+                }
+                if (!blockActions.isEmpty()) {
+                    for (PlayerBlockActionData action : blockActions) {
                         BlockVector3 blockPos = action.getPosition();
                         if (blockPos == null) {
                             continue;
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a false fast-break rejection that occurs when a player holds the dig button and finishes one block then starts the next within the same input tick. Because `PlayerAuthInputPacket` stores block actions in an `EnumMap`, iteration follows the `PlayerActionType` ordinal, so `START_DESTROY_BLOCK` was replayed before `PREDICT_DESTROY_BLOCK`, resetting `lastBreak` an instant before the completion check and causing `useBreakOn` to reject the finished block.

The fix stably re-orders actions so destruction-ending actions (`PREDICT_DESTROY_BLOCK`, `ABORT_DESTROY_BLOCK`, `STOP_DESTROY_BLOCK`) replay before starting/continuing ones, and an ending action at a different position no longer starts a new destruction timer. The fast-break check itself is unchanged.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None.

### Recommendations
- None. The new `PlayerBlockActionOrderTest` covers completion-before-start ordering, abort/stop ordering, stable ordering within groups, single-action passthrough, and ending-action classification.

### Notes
- Verified locally: `mvn compile` passes on the PR head (777a4918) and `mvn test -Dtest=PlayerBlockActionOrderTest` passes 5/5.
- The sort uses `List.sort` (stable TimSort) on a fresh copy, so the packet's own map is untouched; per-tick action counts are tiny, so the copy cost is negligible.
- The two new helpers are package-private static methods, so there is no public API surface change.
- Antivirus scan of the diff is clean; no new dependencies.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #935, opened 2026-09-08
fix: bound inbound network decoding, queues, and per-tick processing

w1zardz:
Incoming batches can accumulate an unbounded session queue and let one busy connection consume an entire server tick. Decoding also needs a separate budget: dropped batches, unknown packet IDs, malformed frames, and legacy compression fallback still perform work even when no packet reaches the queue.

Add measured batch decoding and independent per-session budgets for wire traffic, decoded bytes, and framed packets. Admit only complete successful batches into a queue capped at 1,300 packets and 6 MiB. Drain at most 120 packets per session per tick, with a shared interface budget of 2,048 packets / 6 MiB and a rotating starting session. Enforce the decoded byte limit centrally for compression providers such as NONE that do not enforce it themselves.

Wire-budget exhaustion closes the channel before decryption. Queue or decode backpressure consumes the encrypted packet and validates its checksum before discarding it, preserving the cipher stream and packet counter for subsequent traffic. Malformed batches after login are isolated and repeated failures exhaust decode tokens; pre-login failures retain fail-closed behavior. Capture the login phase and remote address before disconnecting so delayed address-block decisions use the original state, and close through the channel's public API.

This overlaps #892 and supersedes its malformed-batch handling policy with bounded decode work, a sliding failure window, and rate-limited diagnostics. This PR applies directly to current master and does not require that branch.

Validation includes session queue/count/byte bounds, fair draining under global packet and byte budgets, accounting for failed fallback attempts and unknown frames, all-or-nothing session admission when a later frame is malformed, post-decompression size enforcement, actual delayed pre-login close callbacks, and real AES encrypted packets across both forms of soft backpressure. Focused session and proxy tests were run with Java 17.

### DIFF
--- a/src/main/java/cn/nukkit/network/Network.java
+++ b/src/main/java/cn/nukkit/network/Network.java
@@ -262,14 +262,15 @@ public class Network {
     }
 
     public boolean processBatch(byte[] payload, Collection<DataPacket> packets, CompressionProvider compression, int raknetProtocol, Player player) {
-        return this.processBatch(payload, packets, compression, raknetProtocol, player, true);
+        return this.processBatchMeasured(payload, packets, compression, raknetProtocol, player, true).success();
     }
 
     public boolean processBatchQuietly(byte[] payload, Collection<DataPacket> packets, CompressionProvider compression, int raknetProtocol, Player player) {
-        return this.processBatch(payload, packets, compression, raknetProtocol, player, false);
+        return this.processBatchMeasured(payload, packets, compression, raknetProtocol, player, false).success();
     }
 
-    private boolean processBatch(byte[] payload, Collection<DataPacket> packets, CompressionProvider compression, int raknetProtocol, Player player, boolean warnOnFailure) {
+    public BatchProcessResult processBatchMeasured(byte[] payload, Collection<DataPacket> packets, CompressionProvider compression,
+                                                   int raknetProtocol, Player player, boolean warnOnFailure) {
         int maxSize = 3145728; // 3 * 1024 * 1024
         if (player != null && player.getSkin() == null) {
             maxSize = 6291456; // 6 * 1024 * 1024
@@ -283,7 +284,16 @@ public class Network {
             } else if (log.isDebugEnabled()) {
                 log.debug("Exception while decompressing batch packet (compression={}, {} bytes)", compression, payload.length, e);
             }
-            return false;
+            return BatchProcessResult.failed(maxSize, MAX_BATCH_PACKET_COUNT, e);
+        }
+
+        // Providers such as NONE do not enforce the requested decompression limit themselves.
+        if (data.length > maxSize) {
+            ProtocolException failure = new ProtocolException("Decompressed batch exceeds " + maxSize + " bytes");
+            if (warnOnFailure) {
+                log.warn("Rejected oversized decoded batch (compression={}, decompressed={} bytes)", compression, data.length);
+            }
+            return BatchProcessResult.failed(data.length, 0, failure);
         }
 
         BinaryStream stream = new BinaryStream(data);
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

> Note: the automated review backend was unavailable (API key disabled); this review was performed manually on the full diff.

### Summary

This PR bounds inbound network work to prevent a single busy connection from consuming an entire server tick or exhausting memory:

- **Network**: `processBatchMeasured` returns a `BatchProcessResult` (success, decompressed bytes, framed packets, failure) while the old boolean `processBatch`/`processBatchQuietly` signatures are preserved as delegating wrappers — no API break. The decoded-byte limit is now enforced centrally after decompression, closing the bypass for providers like `NONE` that don't enforce it themselves.
- **RakNetPlayerSession**: independent token buckets for wire bytes, decode bytes, and framed packets; a per-session queue hard-capped at 1,300 packets / 6 MiB with a low-water hysteresis for throttle release; malformed batches from logged-in sessions are isolated behind a sliding failure window (32 / 5 min) while pre-login stays fail-closed. Soft backpressure consumes the encrypted packet and validates its checksum before discarding, preserving the cipher stream and counter — verified by a real AES round-trip test. Delayed address-block decisions now capture the login phase and remote address before disconnect, and the close path uses the channel's public API instead of `unsafe()`.
- **RakNetInterface**: a shared per-tick budget (2,048 packets / 6 MiB) with a rotating round-robin cursor so sessions skipped by budget exhaustion are served first next tick.
- **Tests**: ~20 new tests cover queue/byte bounds, low-water release, budget accounting for failed fallback attempts and unknown frames, all-or-nothing batch admission, sliding malformed windows, delayed close policy, and both forms of encrypted soft backpressure.

The change directly addresses a real DoS surface and supersedes the malformed-batch policy of #892. Overall quality is high.

### Issues Found

#### Blocking Issues
- None.

#### Non-blocking Issues
- `enqueueDecodedBatch` has a check-then-act race between the queue-size read and the counter increment (event loop vs. server thread). Overshoot is bounded by one batch (≤ 1,300 packets / 6 MiB), so impact is negligible, but a fused `getAndAdd` + rollback would be tighter.
- `packetBytes()` uses `packet.getCount()` as the byte proxy; packets that were never encoded report a near-zero count, slightly under-charging the byte budgets. Acceptable for inbound packets (always decoded from wire bytes), worth a comment.
- `rejectWireIngress` logs at warn level on every oversized packet before close — one per connection since the channel is closed immediately, but a short comment noting that invariant would help future edits.
- A single head packet larger than the remaining byte budget stalls that session's queue for the tick (head-of-line blocking). Bounded by the round-robin rotation and covered by `globalByteBudgetEventuallyServesALargeHeadPacket`, so behavior is intentional — just noting it.

### Recommendations
- Consider documenting the new budget constants (120 packets/tick, 6 MiB/tick, token rates) in a config or class-level javadoc so server operators can reason about worst-case per-tick network cost.
- If the proxy/VPN use case matters, double-check that legitimate high-throughput clients (e.g., proxy-forwarded sessions sharing one address) fit within the 1,200-batch token bucket at 300/s refill.

### Notes
- ClamAV scan and local compile verification were skipped: the host is under memory pressure from concurrent builds. The diff is pure Java with no binary/archive payloads, and the author reports focused session/proxy tests passing on Java 17.
- Behavior change worth highlighting in release notes: malformed batches from logged-in players no longer disconnect immediately (up to 32 per 5-minute window), and IP blocking after malformed packets no longer applies to `LOGGED_IN` sessions.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

lt-name:

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Re-review at maintainer request. The changes bound inbound network work on three axes: hard wire-frame cap (12 MiB), token-bucket ingress budgets (compressed bytes, decode bytes, framed packets), per-session and per-interface per-tick processing limits with fair round-robin draining, bounded per-session queues with low-water backpressure release, and a sliding-window malformed-batch tolerance for logged-in sessions. Verified locally: clean compile and all 35 `RakNetPlayerSessionTest` tests pass, including the new coverage for queue bounds, ingress budgets, round-robin fairness, malformed windows, and encrypted-session soft throttle.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- When the per-session inbound queue is full, `enqueueDecodedBatch` returns false and the successfully decoded batch is silently dropped (only a rate-limited warn log). A legitimately busy logged-in player may experience packet loss without disconnect; this is an intentional backpressure tradeoff and the low-water mark re-opens the queue, but a kicked-with-reason fallback could be kinder to affected clients.
- Inbound throttle state is cleared only inside `serverTick`; a session that stops sending while throttled stays flagged until the queue drains, which is harmless but worth noting.

### Recommendations
- Consider a metrics counter export for `throttledInboundBatches` so operators can alert on sustained backpressure instead of scraping logs.
- The 12 MiB wire cap, 6 MiB queue byte bound, and 120-packet tick budget are well-chosen; documenting them in the server config reference would help operators tune expectations.

### Notes
- `Network.processBatch`/`processBatchQuietly` return values are unchanged semantically; the new `processBatchMeasured` only adds accounting (decompressed bytes, framed packets, failure cause).
- Behavior change is deliberate and consistent: logged-in sessions are no longer IP-blocked after malformed batches and instead get a 32-per-5-min sliding tolerance; pre-login stays fail-closed.
- Soft backpressure (queue/decode budget) correctly advances the cipher stream and checksum counter before dropping, keeping encrypted sessions decryptable — covered by `encryptedSoftThrottleKeepsTheNextPacketDecryptable`.
- No external callers of the changed `serverTick(int, long)` signature exist outside `RakNetInterface.process()`, which is updated accordingly.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #936, opened 2026-09-08
feat: expose complete server tick timing through paired events

w1zardz:
Plugins can observe scheduler work, but that does not measure the complete server tick, including network processing, world/player updates, autosave and bookkeeping. The existing tick-usage estimate is capped at 50 ms, so it cannot express the duration of a slow tick.

Add non-cancellable `ServerTickStartEvent` and `ServerTickEndEvent` events with matching tick IDs. The end event exposes the uncapped duration in monotonic nanoseconds and runs from `finally`, including when the tick body or start dispatch throws. If both the body and end dispatch fail, the original failure is preserved and the observer failure is suppressed. Waiting between ticks and end-event listeners are excluded from the reported interval.

The normal tick entry point still supplies the real clocks; a package-private clock seam makes boundary and failure tests deterministic.

Validation: all 8 `ServerTickMeasurementTest` tests pass on Java 17. They cover idle time exclusion, work after the scheduler and durations over 50 ms, early-wakeup suppression, paired events during failures, end-listener timing, operation without listeners, and listener removal on plugin disable/reload.

```text
mvn -q -Dtest=ServerTickMeasurementTest test
```

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -24,6 +24,8 @@ import cn.nukkit.event.server.BatchPacketsEvent;
 import cn.nukkit.event.server.PlayerDataSerializeEvent;
 import cn.nukkit.event.server.QueryRegenerateEvent;
 import cn.nukkit.event.server.ServerStopEvent;
+import cn.nukkit.event.server.ServerTickStartEvent;
+import cn.nukkit.event.server.ServerTickEndEvent;
 import cn.nukkit.inventory.CraftingManager;
 import cn.nukkit.inventory.Recipe;
 import cn.nukkit.item.Item;
@@ -129,6 +131,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Pattern;
+import java.util.function.LongSupplier;
 
 /**
  * The main server class
@@ -1781,7 +1784,11 @@ public class Server {
     }
 
     private void tick() {
-        long tickTime = System.currentTimeMillis();
+        tick(System.currentTimeMillis(), System::nanoTime);
+    }
+
+    // The actual tick body also accepts a monotonic clock for deterministic boundary tests.
+    void tick(long tickTime, LongSupplier nanoTime) {
 
         long time = tickTime - this.nextTick;
         if (time < -25) {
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR adds a pair of non-cancellable events, `ServerTickStartEvent` and `ServerTickEndEvent`, that bracket one full server tick (network, scheduler, worlds/players, autosave, bookkeeping) and expose the uncapped tick duration in monotonic nanoseconds. The implementation is careful and correct: the duration is captured before end-event dispatch so observers cannot inflate their own measurement, the end event fires from `finally` so start/end stay paired even when the tick body throws, and a body failure is preserved while an end-observer failure is suppressed onto it. A package-private clock seam keeps boundary and failure tests deterministic without changing the public entry point. All 8 `ServerTickMeasurementTest` tests pass locally on Java 17 (`Tests run: 8, Failures: 0, Errors: 0`).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- An `Error` (e.g. `AssertionError`) thrown by a `ServerTickEndEvent` listener on an otherwise successful tick now propagates out of `tick()`; `tickProcessor` only catches `RuntimeException` per iteration, so such an `Error` would reach the outer `catch (Throwable)` and terminate the tick loop. This is consistent with how `Error`s from any other tick-stage code behave today, so it is not a regression — but end-event observers are a new, easy place for plugins to trigger it. (Regular listener `Exception`s are already swallowed and logged by `PluginManager.callEvent`, so only `Error`s are affected.)

### Recommendations
- Optionally catch `Throwable` around the end-event dispatch and log instead of rethrowing when the tick body succeeded, if you want observer `Error`s to be non-fatal.
- Consider documenting in the event Javadoc that start-event listener time is included in the reported duration while end-event listener time is not.

### Notes
- Event classes follow the existing codebase conventions (`extends ServerEvent`, static `getHandlers()`, non-cancellable by inheritance).
- No breaking changes: the private `tick()` signature and behavior for early wakeups are unchanged; the new `tick(long, LongSupplier)` overload is package-private.
- Test coverage is strong: idle-time exclusion, durations over 50 ms, early-wakeup suppression, failure pairing with suppressed observer failures, end-listener time exclusion, no-listener operation, and listener cleanup on plugin disable/reload.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):
LGTM. No blocking issues found.

tachibana-kanade-angel:

PR #938, opened 2026-09-08
perf: reduce collision, palette, and chunk-loader lookup allocations

w1zardz:
Collision scans currently materialize `BlockAir` objects for empty cells, LevelDB section serialization resolves the same protocol palette for each of 4,096 blocks, and the chunk-loader index boxes coordinate hashes as `Long` keys. This changes those three paths to avoid repeated work:

- Read the raw chunk block ID before constructing blocks for collision checks. Custom predicates still receive air, explicit air searches still check their collision boxes, and missing chunks retain their existing air behavior without loading them.
- Resolve the LevelDB section palette and hashed-ID choice once per section. Protocols before 1.16.100 retain the legacy lookup path, including anti-xray handling.
- Use the existing `Long2ObjectNonBlockingMap` for the outer chunk-loader index, avoiding boxed keys and the `x ^ z` collision pattern of `Long.hashCode()` on diagonal chunk coordinates.

Validation: 108 focused tests pass on Java 17, covering custom air predicates, unloaded chunks, chunk hints and boundaries (including destination queries with an entity still in another world), existing lava/collision checks, loader registration/removal and concurrent reads, async loading and save backpressure. Byte-for-byte serialization comparisons use the previous per-block algorithm across six legacy/modern/NetEase protocol selections with anti-xray and hashed IDs each on and off. This PR makes no throughput or latency benchmark claim.

```text
mvn -q -Dtest=CollisionHelperAirTest,CollisionHelperRunawayTest,StateBlockStorageNetworkTest,BlockStorageTest,ArmorDamageReductionTest,AsyncChunkLoadTest,LevelChunkSaveBackpressureTest,LevelChunkLoaderIndexTest test
```

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -315,7 +315,12 @@ public class Level implements ChunkManager, Metadatable {
 
     private final Int2IntMap loaderCounter = new Int2IntOpenHashMap();
 
-    private final Map<Long, Map<Integer, ChunkLoader>> chunkLoaders = new ConcurrentHashMap<>();
+    /**
+     * Primitive keys avoid boxing and mix both coordinates of {@link #chunkHash(int, int)}.
+     * Long.hashCode folds them into x ^ z, causing colliding bins for diagonal chunks.
+     * Keep concurrent outer-map reads available to the chunk-loading pipeline.
+     */
+    private final Long2ObjectNonBlockingMap<Map<Integer, ChunkLoader>> chunkLoaders = new Long2ObjectNonBlockingMap<>();
 
     private final Map<Long, Map<Integer, Player>> playerLoaders = new ConcurrentHashMap<>();
 
--- a/src/main/java/cn/nukkit/level/format/leveldb/structure/StateBlockStorage.java
+++ b/src/main/java/cn/nukkit/level/format/leveldb/structure/StateBlockStorage.java
@@ -5,6 +5,7 @@ import cn.nukkit.Nukkit;
 import cn.nukkit.Server;
 import cn.nukkit.block.Block;
 import cn.nukkit.block.BlockID;
+import cn.nukkit.level.BlockPalette;
 import cn.nukkit.level.GlobalBlockPalette;
 import cn.nukkit.level.Level;
 import cn.nukkit.level.format.anvil.util.BlockStorage;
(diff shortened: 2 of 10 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
No blocking issues found. The changes look reasonable overall. This PR reduces allocations on three hot paths — collision queries, LevelDB section serialization, and the chunk-loader index — with behavior-preserving transformations and strong test coverage (108 focused tests verified passing locally on Java 17).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `CollisionHelper.isAirAt` reads the raw block ID and `Level#getBlock` then re-reads state and re-validates the chunk hint, so the section is consulted twice per non-air cell. This is still strictly cheaper than materializing `BlockAir` for every empty cell, but a single-read design could go further.
- `ArmorDamageReductionTest` now stubs `FullChunk#getBlockId` to match the `Level#getBlock` lava fixture, coupling the test to the raw-ID lookup detail of the implementation. Acceptable, but worth keeping in mind if the probing strategy changes again.

### Recommendations
- The `useHashedBlockNetworkIds` flag is now read once per section instead of once per block — this is actually a consistency improvement (a mid-serialization toggle can no longer mix ID formats within one section). Worth calling out as intentional.
- Consider a brief note in the changelog that diagonal-coordinate chunk bins no longer collide in the loader index, since servers with many loaders may see measurably lower lookup costs.

### Notes
- Equivalence verified by reading the relevant sources: `FullChunk#getBlockId` and `getBlockState[0]` resolve through the same storage in anvil, LevelDB, and empty sections, so the air fast path agrees with the materialized-block path in all cases.
- The modern palette path (`>= 1.16.100`) replicates `GlobalBlockPalette#getOrCreateRuntimeId` exactly (same threshold, same palette resolution, same hashed-ID check); legacy protocols keep the per-block lookup including anti-xray handling.
- The `Long2ObjectNonBlockingMap` swap for `chunkLoaders` only touches `get`/`put`/`remove` call sites; the pre-existing non-atomic get-then-put registration pattern is unchanged. The special-cased zero key (`chunkHash(0,0)`) is covered by the new index test.
- The static `getCollisionBlocks`/`hasCollisionBlocks`/`getCollisionCubes` overloads now ignore an entity chunk hint from a different world — a small behavior fix covered by tests.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #939, opened 2026-09-08
fix(skin): NetEase V860 mall 3D/4D skins invisible or stuck as default Steve on other clients (fullSkinId routing + player-list entry rebuild timing)

Pigeon2333:
## Symptom (NetEase China client, V860 / protocol 1.21.124-netease)

Mall-purchased 3D/4D (persona) skins rendered wrongly on other players' clients:

- fall back to a placeholder Steve first, then the whole entity turns **invisible**;
- or stay on default Steve forever and never show the real skin;
- server-side skins shared by multiple entities (RsNPC-style NPC skins) become invisible or fall back to Steve;
- the player's own self-view goes invisible, occasionally recovering in the pause menu.

## Root causes (four layers, each sufficient on its own)

1. **fullSkinId one-shot policy breaks both ways.** V860 matches mall/4D skin profiles by fullSkinId. Upstream randomized it on every send (designed for uniqueness of shared server-side skins) → real players' mall skins never match a profile → Steve placeholder first, then full invisibility. Naively making it always stable → profile matching for shared-instance server skins hangs / gets negative-cached → NPCs invisible. It has to be **routed by skin origin** (this PR's `Skin#isFullSkinIdStable`).
2. **NetEase built-in default skins always fail validation.** skinIds like `{uuid}.Steve` / `{uuid}.Alex` / `Standard_Custom`, when sent with a stable fullSkinId, are rejected by the V860 validation flow → the player's self-view goes invisible. These must be randomized so the client renders the in-packet bytes directly.
3. **A bare ADD replacing a PlayerList entry is not idempotent.** For an entry the client already holds, V860 ignores a bare ADD (no re-render); for the viewer's own entry, a bare ADD outright breaks self-rendering. REMOVE → ADD is the only safe sequence (REMOVE on a missing client entry is a harmless no-op).
4. **Late confirmations are silently dropped.** The initial handshake removes the entry via a delayed REMOVE after 5 ticks; a later ConfirmSkinPacket from the sweep then fails forever against a "entry must still exist" hard gate — silently — leaving the observer stuck on the default skin permanently.

## What the 4 commits do

| Commit | Change |
|--------|--------|
| c064d66 | Stop unconditionally randomizing fullSkinId on V860 (real mall-skin profiles can match again) |
| 2ac988a | Fix the 236 SyncSkinPacket wire format: round-1 third field is a byte array (previously written as a string → client parse failure) |
| fc087b0 | fullSkinId stable-key routing: explicitly-set ids (login / skin-change packets, plugin `setFullSkinId`) and persona skins are sent as-is; shared server skins and NetEase default skins are randomized per send. Adds `Skin#isNetEaseBuiltinDefaultSkinId` / `Skin#isFullSkinIdStable` |
| cab9f50 | PlayerEntitySkinSender: entry rebuild is an unconditional REMOVE → ADD; a successful rebuild bumps a generation that invalidates any still-scheduled raw handshake; prepareConfirmSkin takes the rebuild branch when the entry is missing; isConfirmable drops the `sentSkins` hard gate |

## Field testing

Verified end-to-end on the FAPIXEL server network (9 servers behind a WaterdogPE proxy, V860 clients): cross-server lobby/room transfers, repeated respawns, room rotation, queue teleports, and third-party minigame plugins with custom respawn flows — mall 3D/4D skins render correctly with no more invisibility or default-Steve fallback; NPC skins fine; self-view fine.

## Notes

- Changes are concentrated in `BinaryStream#putSkin` (V1_21_124_NETEASE branch), `Skin`, `PlayerEntitySkinSender`, and the SyncSkinPacket wire format; no behavior change on non-NetEase protocol paths (all routing conditions gate on `gameVersion == V1_21_124_NETEASE`).
- The 236 SyncSkinPacket change only affects a NetEase-China-specific packet; upstream can skip commit 2 if that channel does not apply.
- Happy to split commits or adjust scope — just say the word.

### DIFF
--- a/src/main/java/cn/nukkit/entity/data/Skin.java
+++ b/src/main/java/cn/nukkit/entity/data/Skin.java
@@ -58,10 +58,6 @@ public class Skin {
 
     private boolean noPlayFab; // Don't attempt to generate missing play fab id multiple times
     private String fullSkinId;
-    // FAPIXEL fap4：fullSkinId 是否为"显式设置的稳定键"（客户端登录/换肤包提供，或服务端
-    // 插件主动 setFullSkinId）。稳定键下发时保持原样；未设置的（RsNPC 等共享 Skin 实例的
-    // 服务端皮肤）由 BinaryStream#putSkin 每次发包随机化，跳过网易客户端档案匹配。
-    private boolean fullSkinIdStable;
     private String skinId;
     private String playFabId = "";
     private String skinResourcePatch = GEOMETRY_CUSTOM;
@@ -399,15 +395,7 @@ public class Skin {
 
     public void setFullSkinId(String fullSkinId) {
         this.fullSkinId = fullSkinId;
-        this.fullSkinIdStable = fullSkinId != null;
         this.noPlayFab = false; // Allow another attempt to generate it using the new id
-    }
-
-    /**
-     * FAPIXEL fap4：fullSkinId 是否为显式设置的稳定键（见字段注释）。
-     */
-    public boolean isFullSkinIdStable() {
-        return this.fullSkinIdStable && this.fullSkinId != null;
     }
 
     public String getFullSkinId() {
--- a/src/main/java/cn/nukkit/entity/data/Skin.java
+++ b/src/main/java/cn/nukkit/entity/data/Skin.java
@@ -377,22 +377,6 @@ public class Skin {
         this.armSize = armSize;
     }
 
-    /**
-     * FAPIXEL fap5：是否为网易客户端自带默认皮肤（skinId 形如 "{uuid}.Steve" / "{uuid}.Alex"，
-     * 或核心回退 Standard_Custom）。此类皮肤在 V860 下走验证必被拒（实测），其
-     * fullSkinId 必须随机化下发（跳过档案匹配、直接渲染包内字节），稳定下发会导致玩家
-     * 自视隐形（09-07 fap4 实测：史蒂夫玩家全程看不到自己，暂停才偶发恢复）。
-     */
-    public static boolean isNetEaseBuiltinDefaultSkinId(String skinId) {
-        if (skinId == null || skinId.isEmpty()) {
-            return false;
-        }
-        String lower = skinId.toLowerCase(java.util.Locale.ROOT);
-        return lower.endsWith(".steve")
-                || lower.endsWith(".alex")
-                || "standard_custom".equals(lower);
-    }
-
     public void setFullSkinId(String fullSkinId) {
         this.fullSkinId = fullSkinId;
         this.noPlayFab = false; // Allow another attempt to generate it using the new id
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes NetEase V860 mall 3D/4D skins rendering as invisible or default Steve on observer clients, through three coordinated changes:
1. **BinaryStream.putSkin**: routes `fullSkinId` on V860 — stable for explicitly-set keys (client login/skin-pack, plugin `setFullSkinId`) and persona skins; randomized per-send for unstable server-side skins (e.g. RsNPC shared Skin instances) and NetEase builtin default skins (`{uuid}.Steve`/`.Alex`, `Standard_Custom`).
2. **PlayerEntitySkinSender**: entry rebuild now always uses REMOVE → ADD (bare duplicate ADD is ignored by V860 and breaks self-rendering), increments the generation counter to invalidate still-scheduled raw resends / delayed REMOVEs, and `scheduleDelayedRemove` now checks the generation *before* unregistering so a rebuilt entry's fresh fingerprint is not wiped.
3. **SyncSkinPacket**: wire format corrected to the actual NetEase V860 layout (round 1 = bool + uuid + skinBytes byte-array; then udid / extraData / itemId strings; the legacy 4th string is no longer sent).

The logic is coherent, the routing rules are well documented, and the packet layout change is symmetric between encode/decode. Compiles cleanly (verified in an isolated worktree, `mvn compile` OK, class artifacts present). ClamAV scan of the diff: clean.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `SyncSkinEntry.string4` is kept as a field but is no longer read from or written to the wire — plugins that populated it will silently send nothing. Acceptable given the layout was previously wrong, but worth noting in release notes.
- `Skin.isNetEaseBuiltinDefaultSkinId` matches any skinId ending in `.steve` / `.alex` (case-insensitive). A custom skinId with such a suffix would be forced down the randomized path. Low risk in practice.
- The random suffix path appends 8 chars to `getFullSkinId()`, which itself may already end with a random 8-char suffix (auto-generated fallback) — harmless, just slightly redundant.

### Recommendations
- Consider dropping or deprecating `SyncSkinEntry.string4` in a follow-up cleanup once downstream plugins are confirmed migrated.
- Consider logging (at debug level) when a skinId matches the builtin-default heuristic but was explicitly set, to help diagnose false positives.

### Notes
- Review was performed manually from the PR diff (automated deep-review tooling was unavailable at review time).
- No `mvw()`-calling methods were added or modified; @Deprecated annotation rule not applicable.
- The `fullSkinIdStable` flag is only set through `setFullSkinId`; the internal auto-generation path in `getFullSkinId()` leaves it `false`, so auto-generated ids correctly take the randomized branch.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel:

PR #941, opened 2026-09-10
fix(entity): load saved mob movement effects and zombie armour without errors

w1zardz:
The `Entity` constructor restores `ActiveEffects` before the field initializers of `EntityLiving` run. A witch saved with Speed or Slowness therefore reached `movementSpeedModifiers` while it was still `null` and failed to load.

Create the movement state before the effects are loaded and keep the speed they compute once the constructor finishes. Players, whose `initEntity` is deferred until login, still start with the base speed and an empty modifier set.

Zombie armour had the same fragility on load. A missing `Slot` now means the position in the legacy list; explicit slots are read in any order; an invalid or repeated slot is skipped while the other items are kept, and damaged entries produce a single warning per mob. The normal save format with four explicit slots is unchanged.

Validation: all 15 EntitySavedMobStateTest cases pass. They call the real witch and zombie constructors with saved movement effects and with legacy, explicit, invalid and repeated armour slots.

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityLiving.java
+++ b/src/main/java/cn/nukkit/entity/EntityLiving.java
@@ -40,6 +40,8 @@ public abstract class EntityLiving extends Entity implements EntityDamageable {
 
     public EntityLiving(FullChunk chunk, CompoundTag nbt) {
         super(chunk, nbt);
+        // Players defer initEntity until login; mobs have already loaded their effects in super.
+        this.initializeMovementState();
     }
 
     @Override
@@ -55,7 +57,7 @@ public abstract class EntityLiving extends Entity implements EntityDamageable {
     protected int attackTime = 0;
     protected int knockBackTime = 0;
 
-    protected float movementSpeed = 0.1f;
+    protected float movementSpeed;
 
     protected int turtleTicks = 0;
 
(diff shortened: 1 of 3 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Fixes two entity load-order bugs: (1) `EntityLiving` field initializers ran after `Entity`'s constructor had already restored `ActiveEffects`, so saved Speed/Slowness modifiers hit a `null` map; movement state is now initialized inside the overridden `initEntity()` before effects load, with a guarded fallback in the constructor for players (deferred init). (2) Zombie armour loading aborted on the first invalid/duplicate slot; it now treats missing `Slot` as the legacy list position, keeps valid entries, and logs a single summary warning. No blocking issues found. The changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `initializeMovementState()` couples map creation with resetting `movementSpeed` to the default; correct today because both are only uninitialized together, but slightly implicit.
- The mixed explicit/positional armour semantics (positional index still increments for explicit-slot entries) is subtle; the tests document it well, a one-line comment on `index++` could help future readers.

### Recommendations
- None required. The new `EntitySavedMobStateTest` (15 cases) covers witch/zombie construction with saved effects and legacy/explicit/invalid/duplicate armour slots; verified locally: 15/15 pass.

### Notes
- Review performed manually; automated review backend was unavailable.
- Save format with four explicit slots is unchanged, confirmed by round-trip test.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #944, opened 2026-09-10
fix(block): render split heads on older palettes and keep their drops

w1zardz:
`ItemSkull.getBlock()` places the split head blocks 1220–1225, which older block palettes do not have. On 1.21.2 (protocol 686) a placed `PLAYER_HEAD` was sent as `INFO_UPDATE`, so the client drew a placeholder block instead of the head.

**Render (first commit).** When the palette is finalised, missing heads get outgoing runtime/hash aliases of the legacy `SKULL_BLOCK` with the same facing. No states are registered, so the canonical reverse lookups used for world storage and the native modern states stay unchanged; the aliases are rebuilt after the palette is rebuilt with custom blocks. The old model gets the head type through the regular `SkullType` NBT. `BlockEntitySkull.isBlockEntityValid` now accepts the whole head family: it used to accept only block 144, so updating a split head deleted its tile.

**Drops (second commit).** A split head whose tile was already removed by that validity check dropped a skeleton skull regardless of its type. When the tile is missing, `BlockSkullSkeleton.getDrops` now takes the type from the block itself. An existing tile is read as before and its `SkullType` wins; a tile without the field keeps type 0. Worlds are not scanned and lost custom NBT is not invented.

Validation: all 6 LegacySkullCompatibilityTest and SkullDropRecoveryTest cases pass. They cover the outgoing aliases on an old palette, untouched reverse lookups, and the drops of all seven head types with a missing tile, a tile with a different `SkullType` and a tile without the field.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockSkullSkeleton.java
+++ b/src/main/java/cn/nukkit/block/BlockSkullSkeleton.java
@@ -135,7 +135,7 @@ public class BlockSkullSkeleton extends BlockTransparentMeta implements Faceable
     @Override
     public Item[] getDrops(Item item) {
         BlockEntity blockEntity = getLevel().getBlockEntity(this);
-        int dropMeta = 0;
+        int dropMeta = this.getSkullType().ordinal();
         if (blockEntity != null) dropMeta = blockEntity.namedTag.getByte("SkullType");
         return new Item[]{
                 new ItemSkull(dropMeta)
--- a/src/main/java/cn/nukkit/blockentity/BlockEntitySkull.java
+++ b/src/main/java/cn/nukkit/blockentity/BlockEntitySkull.java
@@ -1,6 +1,6 @@
 package cn.nukkit.blockentity;
 
-import cn.nukkit.block.Block;
+import cn.nukkit.block.BlockSkullSkeleton;
 import cn.nukkit.level.format.FullChunk;
 import cn.nukkit.nbt.tag.CompoundTag;
 
@@ -33,7 +33,7 @@ public class BlockEntitySkull extends BlockEntitySpawnable {
 
     @Override
     public boolean isBlockEntityValid() {
-        return getBlock().getId() == Block.SKULL_BLOCK;
+        return getBlock() instanceof BlockSkullSkeleton;
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes split head blocks (1220–1225) rendering as `INFO_UPDATE` placeholders on older block palettes (e.g. 1.21.2 / protocol 686), and fixes split heads dropping a skeleton skull when their block entity was already removed.

The approach is sound:
- **Render**: outgoing-only runtime/hash aliases to the legacy `SKULL_BLOCK` state are added at palette lock time, after all vanilla/custom states are registered. No synthetic states are registered, so canonical reverse lookups (world storage) and native modern states stay untouched. Aliases are correctly rebuilt after `clearStates()` + re-lock (custom block rebuild path), and double `lock()` is guarded.
- **Drops**: when the tile is missing, `getDrops` now falls back to the type encoded by the block class (`getSkullType().ordinal()`, matching `ItemSkull` damage values 0–6). An existing tile's `SkullType` still wins, and a tile without the field keeps the legacy default 0 — no data is invented.
- **Tile validity**: `BlockEntitySkull.isBlockEntityValid` now accepts the whole `BlockSkullSkeleton` family (all 6 head classes extend it), so updating a split head no longer deletes its tile; replaced blocks (e.g. stone) are still removed.

Verified locally: `mvn compile` passes and all 6 new tests (`LegacySkullCompatibilityTest`, `SkullDropRecoveryTest`) pass. Tests cover outgoing aliases on old palettes, untouched reverse lookups, native states on modern palettes, alias rebuild after palette reset, spawn NBT type/rotation retention, and drop behavior for all seven head types with missing/mismatched/typeless tiles.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None significant.

### Recommendations
- None.

### Notes
- `SkullType` enum ordinals (DEFAULT=0 … PIGLIN=6) match the `ItemSkull` damage values, so the new drop fallback is consistent with item meta.
- The alias loop covers meta 0–5 (all legacy skull facing values); entries are only added when the corresponding legacy skull runtime state exists, and existing native entries are never overwritten (`containsKey` / `putIfAbsent`).

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #975, which contains the same changes (the split-head rendering and drop fixes) along with additional fixes, rebased onto the current master.

Closing this PR to avoid double-applying these changes. If any part of this PR is not covered by #975, feel free to reopen or comment and we will split it out.

PR #945, opened 2026-09-10
fix(entity): preserve the hurt window after a successful totem rescue

w1zardz:
A totem rescue cancels lethal damage and returns `false`, which bypasses the ordinary attack timer. The rescued entity was left without a hurt window, so the next hit of the same strength landed immediately and a combo could burn through a totem in the same instant it popped.

Arm `attackTime` after a successful rescue and keep any longer timer that is already running. Equal or weaker follow-up hits wait for the normal interval; stronger hits and lethal command damage keep their existing behaviour. `noDamageTicks`, the hand order and the cancelled rescue result are preserved.

Validation: all 20 PlayerTotemHandPriorityTest cases pass, driving the real attack chain and `EntityLiving` on several protocols.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -1874,6 +1874,12 @@ public abstract class Entity extends Location implements Metadatable {
                         p.getInventory().decreaseCount(p.getInventory().getHeldItemIndex());
                     }
 
+                    // A totem consumes the lethal hit but returns false below, so EntityLiving
+                    // cannot arm its usual attackTime. Retain that same hurt window, including
+                    // its stronger-hit rule, without granting absolute noDamageTicks immunity.
+                    EntityLiving living = p;
+                    living.attackTime = Math.max(living.attackTime, source.getAttackCooldown());
+                    this.scheduleUpdate();
                     source.setCancelled(true);
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a real gap: a successful totem rescue in `Entity.attack()` cancels the lethal hit and returns `false`, which means `EntityLiving.attack()` never arms its usual `attackTime` hurt window. The rescued player could then be hit again by equal-strength damage in the same tick, letting a combo burn through a totem the instant it popped. The fix arms `attackTime = max(attackTime, source.getAttackCooldown())` at the rescue site, exactly mirroring the ordinary hurt-window semantics (equal/weaker hits blocked by the last-damage comparison, stronger hits still pass, `noDamageTicks` untouched), and calls `scheduleUpdate()` consistent with the normal path. All 20 `PlayerTotemHandPriorityTest` cases were executed locally against the PR head and pass.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None.

### Recommendations
- None.

### Notes
- `SUICIDE`/`VOID` causes still bypass the totem, and pre-existing absolute protection (`noDamageTicks`) still blocks damage without consuming a totem — both covered by tests.
- The `Math.max` guard correctly preserves any longer-running hurt window (e.g. one armed during effect application) instead of shortening it.
- Main-code change is minimal (6 lines in `Entity.java`); the rest is test coverage across protocols 589/786/2169, driving the real attack chain.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #977, which contains the same changes (the totem hurt-window preservation) along with additional fixes, rebased onto the current master.

PR #946, opened 2026-09-10
feat(block): make the piston push limit configurable up to 64 blocks

w1zardz:
The piston push limit is hard-coded at 12 blocks. Servers that build larger contraptions (or want a stricter limit) have no way to change it short of patching the calculator.

Add a `piston-push-limit` server setting. The default stays at 12 moved blocks. Each `BlocksCalculator` snapshots one validated limit in the range 1–64, with invalid values falling back to 12, and applies it to forward movement and to every attached slime branch. The rear chain is counted without counting the origin twice, and the walk stops at cells already scheduled to move.

The existing world-height and immovable-block checks are unchanged.

Validation: all 30 BlockPistonPushLimitTest and BlockPistonBlockedRetryTest cases pass. They exercise the real calculator for configured boundaries, sticky pulls, branch budgets, invalid settings, captured configuration, high and negative coordinates, terminal destruction and unique bounded movement lists; the existing blocked-retry fixture now supplies the server settings.

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockPistonBase.java
+++ b/src/main/java/cn/nukkit/block/BlockPistonBase.java
@@ -311,6 +311,7 @@ public abstract class BlockPistonBase extends BlockSolidMeta implements Faceable
         private final Block blockToMove;
         private final BlockFace moveDirection;
         private final boolean extending;
+        private final int pushLimit;
 
         private final List<Block> toMove = new ArrayList<>();
         private final List<Block> toDestroy = new ArrayList<>();
@@ -318,6 +319,8 @@ public abstract class BlockPistonBase extends BlockSolidMeta implements Faceable
         public BlocksCalculator(boolean extending) {
             this.pistonPos = getLocation();
             this.extending = extending;
+            int configuredLimit = level.getServer().getServerConfig().gameFeatureSettings().pistonPushLimit();
+            this.pushLimit = configuredLimit >= 1 && configuredLimit <= 64 ? configuredLimit : 12;
 
             BlockFace face = getBlockFace();
             if (!extending) {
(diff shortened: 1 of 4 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR makes the piston push limit configurable via a new `piston-push-limit` server setting (default 12, valid range 1-64, invalid values fall back to 12). The limit is snapshotted once per `BlocksCalculator` and applied to forward movement, rear sticky chains, and slime branches. It also fixes a pre-existing off-by-one where the rear chain double-counted the origin, and prevents duplicate movement cells when slime branches intersect. The change is small, focused, and well covered by tests (boundary values, invalid config, sticky pulls, branch budgets, world-height limits, immovable blocks, uniqueness fuzzing).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- Default-behavior change: with the default limit of 12, a sticky retraction can now pull 12 blocks in total where the old accounting (`++count + toMove.size() > 12`) effectively capped it at 11. This matches vanilla and appears intentional, but it is a subtle behavior change worth noting in the changelog.
- When the rear slime chain encounters a block already scheduled to move, the walk now stops (`break`). Blocks beyond that point in exotic intersecting slime assemblies will simply not be pulled - no corruption, but the movement is partial in that edge case.
- `toMove.contains(block)` relies on coordinate equality from `Vector3.equals`. This is consistent with existing code in the same method, but a comment noting the dependency would help future maintainers.

### Recommendations
- Consider documenting the 1-64 clamp behavior in the server config documentation/wiki.

### Notes
- Limit validation (`>= 1 && <= 64`, else 12) is safe against overflow and negative/extreme values.
- Existing world-height and immovable-block checks are unchanged, confirmed by the new tests.
- `BlockPistonBlockedRetryTest` fixture was updated to supply server settings, keeping existing coverage intact.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #948, opened 2026-09-10
fix(player): tolerate float ceiling contact without horizontal setbacks

w1zardz:
Under a two-block ceiling, server-authoritative input rebuilds the player's feet by subtracting the eye offset in float precision. The result can put the top of the player's box 0.0000007–0.0000122 blocks inside the leaves above, and `fastMove` then rejected the whole horizontal step. A few ordinary 0.35-block steps received in one server tick added up past the horizontal correction threshold and triggered `forceMovement`: the player walking under foliage was rubber-banded back although nothing was wrong.

Only after a player collision has already been detected, repeat the check with the top of the box lowered by 0.0001 blocks. Feet and width stay where they were, and on success the original box is kept. The bottom face is deliberately not raised: the cell walk could then skip a fence whose shape is taller than one block. Real penetration into ceilings, walls and floors, and the movement of other entities, are handled exactly as before.

Validation: all 114 PlayerCeilingContactTest cases pass.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -2821,7 +2821,19 @@ public abstract class Entity extends Location implements Metadatable {
 
         AxisAlignedBB newBB = this.boundingBox.getOffsetBoundingBox(dx, dy, dz);
 
-        if (server.getAllowFlight() || !this.level.hasCollision(this, newBB, false)) {
+        // The float eye-position round trip can place a player's head a few
+        // millionths of a block inside a ceiling. Rejecting all three axes for
+        // that contact turns batched sprint-jumps into horizontal speed setbacks.
+        // Retry only a colliding player's ceiling with a bounded tolerance. Keep
+        // the feet unchanged: raising minY can skip a fence's lower block cell
+        // even though its collision shape extends half a block above that cell.
+        boolean canMove = server.getAllowFlight() || !this.level.hasCollision(this, newBB, false);
+        if (!canMove && this instanceof Player) {
+            AxisAlignedBB ceilingContact = newBB.clone();
+            ceilingContact.setMaxY(ceilingContact.getMaxY() - 1.0E-4);
+            canMove = !this.level.hasCollision(this, ceilingContact, false);
+        }
+        if (canMove) {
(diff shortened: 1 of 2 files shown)


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes spurious horizontal movement setbacks (rubber-banding) for players walking under two-block ceilings (e.g. foliage), caused by float-precision rounding in the eye-position round trip placing the top of the player's box a few millionths of a block inside the ceiling. When `fastMove` detects a collision for a player, it retries with the box top lowered by a bounded 1.0E-4 tolerance; on success the original box is kept. Feet and width are untouched, and non-player entities keep the strict check. The change is small, well-scoped, and comes with a thorough 430-line parameterized regression test suite.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- The `this instanceof Player` check lives in the `Entity` base class, which couples the base class to a subclass. Acceptable here since `fastMove` is shared and the alternative (overriding in `Player`) would duplicate the movement logic, but worth noting for future movement changes.

### Recommendations
- If more entity-specific movement tolerances appear later, consider extracting the collision retry into an overridable hook so `Entity` does not accumulate subclass checks.

### Notes
- The tolerance (1.0E-4) is ~8x the maximum observed float error (0.0000122), so real penetrations beyond that remain fully blocked; a player cannot use this to pass through ceilings.
- Deliberately not raising `minY` preserves fence/wall collision shapes that extend half a block above their cell — the test suite verifies this explicitly.
- The retry keeps the original bounding box on success, so no vertical displacement accumulates over repeated contacts (also covered by tests).
- Walls and floors retain zero tolerance, including penetrations smaller than the ceiling tolerance — verified by the test suite.
- Validation per PR description: all 114 PlayerCeilingContactTest cases pass.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #979, which contains the same changes (the ceiling-contact setback tolerance) along with additional fixes, rebased onto the current master.

Closing this PR to avoid double-applying these changes. If any part of this PR is not covered by #979, feel free to reopen or comment and we will split it out.

PR #951, opened 2026-09-14
fix(particle): use the recipient item palette for break particles

w1zardz:
`ItemBreakParticle` encoded the broken item with the global runtime item mapping. When a viewer on an older protocol could not resolve that item (for example the golden spear), encoding threw while the break particle was being broadcast, and the exception interrupted the item break itself: the broken spear did not leave its owner's hand.

The particle now resolves only its outbound payload against the recipient's item palette, and falls back to the existing `info_update` placeholder when the viewer's palette has no entry for the item. Native payloads, server-side items, NBT and the global mappings are left untouched; nothing changes for viewers whose palette knows the item.

Validation: new `ItemBreakParticleCompatibilityTest` (88 cases across protocols); `mvn test -Dtest='cn.nukkit.level.particle.**'` — 94 tests, no failures or errors.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
`ItemBreakParticle.mvEncode` now resolves the outbound item against the recipient's runtime item palette. When the viewer's palette has no entry for the broken item (e.g. the golden spear on older protocols), the packet falls back to the `info_update` placeholder instead of throwing, so the broadcast no longer interrupts the item-break logic. The server-side item, its NBT and the global mappings are untouched. Verified locally: `mvn test -Dtest='cn.nukkit.level.particle.**'` — 94 tests, 0 failures/errors, BUILD SUCCESS. No blocking issues found; the changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- Control flow relies on throwing and catching `IllegalArgumentException` for the unsupported-item path; an `if/else` on `isSupportedOn` would read more directly (the catch is still needed as a safety net for `getNetworkId`).
- `Item.get(Item.INFO_UPDATE)` allocates a new placeholder item on every fallback encode; negligible, but a cached constant would avoid it.
- Tests cover protocols >= 589, while the guarded code path applies to >= v1_16_100 (419); the 1.16–1.20.x fallback path is untested (logic is version-agnostic, so risk is low).

### Recommendations
- Consider restructuring the try/catch into a plain `isSupportedOn` branch, keeping the catch only around `getNetworkId`.
- Optionally extend the parameterized test window to older >= 1.16.100 protocols if a fixture palette is available.

### Notes
- Broadcast path confirmed: `Level.addParticle` groups players by `GameVersion` and calls `mvEncode` per version group, so the fallback is correctly scoped per recipient palette.
- Fallback resets damage to 0, which is correct since the `info_update` placeholder carries no meta.
- Tests verify the original item instance (damage + NBT) is not mutated and that no fake global mapping is installed.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
The changes in this PR are already present on the current master (merged via another PR), so this PR no longer has any effect.

Closing it. Thank you for the contribution!

PR #952, opened 2026-09-14
fix(inventory): restore authoritative inventory snapshots without splitting slots

w1zardz:
`BaseInventory#setItemForce` is the path used for server-side rollback and snapshot restore. It split a legacy overstack into a capped slot plus an overflow and routed the overflow into other slots. For a restore this is wrong in three ways:

- it changed unrelated slots that the snapshot was about to write anyway;
- with a full inventory the remainder was discarded with a warning;
- when a later entry of the same snapshot overwrote the slot the overflow had been moved into, the redistributed items were lost.

`setItemForce` now restores the exact saved count and metadata into its original slot. Ordinary `setItem` and `addItem` still cap new writes at the stack limit, so this does not let normal gameplay create overstacks. `routeOverflowForce` is removed together with the split, and `OverstackSplitTest` is updated to expect exact restoration.

Validation: new `InventoryForceRestoreConservationTest` covering a full inventory, a real server-authoritative inventory rollback, repeated restores, NBT and stack network ids; `mvn test -Dtest='cn.nukkit.inventory.**'` — 215 tests, no failures or errors.


tachibana-kanade-angel:
This PR has been superseded by #976, which contains the same changes (the authoritative inventory snapshot restoration) along with additional fixes, rebased onto the current master.

PR #953, opened 2026-09-14
perf(leveldb): remap palette indices once while compressing block storage

w1zardz:
Profiles of servers running LevelDB auto-compaction show `StateBlockStorage` compression spending a large share of its CPU in `ObjectArrayList.indexOf`: every cell looked its palette entry up in the new palette linearly, so the cost grew with `cells × palette size`.

Compression now caches the new index for each old palette index and deduplicates encountered snapshots with a hash map, so every old entry is remapped once. Palette entry zero, first-use order, equality semantics, bit widths, the AIR special cases and the serialized block-state NBT are all preserved.

Validation: new `StateBlockStorageCompressionTest` compares every cell and the serialized storage bytes against the previous algorithm across palette-size boundaries, duplicate palette entries and independent layers; `mvn test -Dtest='cn.nukkit.level.format.leveldb.**'` — 61 tests, no failures or errors.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

`StateBlockStorage.compress()` previously looked up each cell's palette entry in the growing new palette via `ObjectArrayList.indexOf`, costing O(cells × palette size). It now remaps each old palette index exactly once: an `int[] remapped` cache per old index plus an `Object2IntOpenHashMap` for first-occurrence dedup. `BlockStateSnapshot` has no `equals`/`hashCode` override, so both the old `indexOf` and the new hash map use identity semantics — the transformation is behavior-preserving. Entry zero, first-use order, bit-width growth, the AIR special cases, and the serialized NBT are all unchanged.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None.

### Recommendations
- None.

### Notes
- Equivalence verified against the previous algorithm using the in-repo oracle test: `StateBlockStorageCompressionTest` compares palette contents, bit-array words, per-cell states, and serialized storage bytes across palette-size boundaries (2..4096), duplicate palette entries, unused entry zero, all-AIR special cases, empty palette, and independent layers with post-compression mutation.
- The new `int[count]` remap array and hash map add O(palette size) memory during compression; palette size is bounded by SECTION_SIZE (4096), so this is negligible.
- The optimization relies on `BlockStateSnapshot` retaining identity equality. If `equals`/`hashCode` are ever added to it, both code paths remain equivalent only if they are consistent — worth keeping in mind.
- Verified locally on an isolated worktree of the PR head: `mvn test -Dtest='cn.nukkit.level.format.leveldb.**'` — 61 tests, 0 failures, 0 errors.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #954, opened 2026-09-14
fix(leveldb): keep chunk snapshot staging outside native I/O locks

w1zardz:
`LevelDBProvider` staged chunk snapshots and committed them to the native database under the same lock, so a slow native write also blocked the short bookkeeping that the server thread needs when it queues the next snapshot of a chunk. Entity state for a snapshot could also be read from the database later, off the thread that captured it.

This change:

- separates the short pending-slot metadata lock from the serialized native commit lock;
- captures entity state on the server thread and defers stale actor database reads using immutable coordinates and the retained entity ids;
- preserves newest-revision ordering and read barriers, and keeps in-flight slots alive until they are committed;
- rejects executor fallback instead of performing native writes inline on the caller;
- gives every future a captured sequence ticket, so its completion proves the write was actually committed even when an earlier queued task discarded a failed slot.

Validation: new `LevelDBPendingWriteTest` covering two concurrent callers, clean snapshots, an old write failing, executor rejection, captured actor cleanup and failed-slot acknowledgements; `mvn test -Dtest='cn.nukkit.level.format.leveldb.**'` — 53 tests, no failures or errors.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR restructures `LevelDBProvider`'s pending-write pipeline so that chunk snapshot staging (short metadata bookkeeping) no longer shares a lock with native LevelDB I/O. Key changes:

- Split the per-slot metadata lock (`pw.lock`) from the serialized native commit lock (`pw.commitLock`); native writes now happen with the metadata lock released, so concurrent `saveChunkFuture` callers only block on short critical sections.
- Entity state is captured on the server thread (`EntitySerializer.snapshotEntities`), while stale-actor DB reads are deferred to the commit worker via a `Cleanup` closure that captures only immutable coordinates and retained IDs.
- Added a `durableSequence` ticket per save future: a future only completes successfully if its captured sequence was actually committed, so a discarded failed slot can no longer silently acknowledge a queued caller.
- Executor rejection no longer performs inline native writes on the caller thread; remaining slots are owned by the explicit shutdown drain.
- `removeEmptyPendingWrite` now keeps slots alive while a native write is in flight (`cur.writing`).

Lock ordering is consistent (`commitLock → lock`, matching `readChunkOffThread`'s DB-lifetime-first order), and the deferred entity cleanup is null-safe against the legacy storage path (`EntityStorageLegacy.deleteEntities` only uses chunk coordinates).

Verified locally: `mvn test -Dtest='cn.nukkit.level.format.leveldb.**'` — 53 tests, 0 failures (including the 6 new `LevelDBPendingWriteTest` cases covering concurrent callers, superseded failed writes, executor rejection, deferred actor cleanup, and failed-slot acknowledgement).

### Issues Found

#### Blocking Issues
- None.

#### Non-blocking Issues
- None significant.

### Recommendations
- Consider a brief note in `saveChunkFuture`'s Javadoc that a rejected submission fails the returned future instead of writing inline, since this is a behavioral change callers may rely on during shutdown.

### Notes
- Security scan (ClamAV) of all touched files: clean.
- No public API breakage; `saveEntities` retains its original contract by delegating to `snapshotEntities(...).apply(...)`.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #956, opened 2026-09-14
fix(entity): let endermen dodge projectiles before taking damage

w1zardz:
`EntityEnderman#attack` called `super.attack()` first and only then checked for `DamageCause.PROJECTILE`, cancelled the event and teleported. By then `Entity#attack` had already fired the event and applied the damage, so the late cancel only hid the hit: arrows, crossbow bolts and other projectiles wore the enderman down and killed it, while vanilla endermen take no damage from projectiles and teleport away instead.

The projectile branch now runs before `super.attack()`. Plugins that want to stop a projectile from reaching the enderman at all can still cancel `ProjectileHitEvent`, which every projectile fires before calling `attack()`. Melee and all other causes keep their previous behaviour.

`getSafeTpLocation()` also rolled a single random column within 16 blocks and gave up when it hit a wall or the void, which left the enderman standing in place most of the time. It now tries up to ten columns, the way vanilla keeps rolling for a valid spot.

Validation: new `EntityEndermanTest`; `mvn test -Dtest='cn.nukkit.entity.**'` — 104 tests, no failures or errors.


tachibana-kanade-angel:
This PR has been superseded by #977, which contains the same changes (the enderman projectile dodge) along with additional fixes, rebased onto the current master.

PR #957, opened 2026-09-14
perf(collision): retain fresh builtin block snapshots during collision scans

w1zardz:
An exact `Level#getBlock` already creates an independent built-in block object, yet `CollisionHelper#getBlocksInBoundingBox` cloned every non-air built-in block a second time. That is one extra `Block` allocation per occupied cell of every collision scan.

For exact `Level` instances with an in-range, loaded matching chunk, the helper now materialises the unpacked state directly and keeps that fresh object. Ownership is decided from that state, not from the earlier AIR probe or the returned block id: custom factories may cache a block with a vanilla id, and the state may change between the two reads. Results from custom factories and `Level` overrides are still cloned, and the original null/AIR filtering order, extended metadata, collision bounds, layer zero and the allocation-free AIR probe are unchanged. No public API changes.

Validation: new `CollisionHelperBlockOwnershipTest` covering fresh snapshots, cached custom factories, a state changed after the AIR probe, custom null fallback, extended metadata, `Level` overrides, unloaded chunks and dynamic collision results; `mvn test -Dtest='CollisionHelper*'` — 32 tests, no failures or errors.


lt-name:
Superseded by https://github.com/MemoriesOfTime/Nukkit-MOT/pull/968

PR #958, opened 2026-09-15
fix(player): send the full view distance before the first spawn

w1zardz:
Before the client sends `SetLocalPlayerAsInitialized`, `Player.sendNextChunk` limits the chunk radius to `ceil(sqrt(spawn-threshold))`. The client, however, has already been told through `RequestChunkRadius` (never below 3) to expect every chunk inside `chunkRadius`. The outer ring is never sent before spawn, so the client stays on its loading state, hanging in the air, until its own timeout; only then does it send the initialized packet that switches the server to the full radius.

It is most visible with a small `spawn-threshold`: with `1` (e.g. a lobby over the void) players waited about 16 seconds before they could interact, and any client whose render distance is larger than `sqrt(spawn-threshold)` waits the same way.

**Fix:** the pre-spawn radius is `max(chunkRadius, ceil(sqrt(spawn-threshold)))`, both for chunk sending and for the teleport chunk readiness check. The threshold still decides when `PLAYER_SPAWN` is sent. PocketMine-MP (`Player::requestChunks`) and PowerNukkitX (`PlayerChunkManager`) also send the whole view distance before spawn.

**Reproduce:** set `spawn-threshold: 1`, join with a render distance of 8+ — the player hangs on the loading/air state for several seconds before spawning.

**Validation:** `mvn -DskipTests compile`; `mvn test -Dtest='cn.nukkit.Player*'` — 88 tests, no failures or errors.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a pre-spawn chunk-sending bug: before the client sends `SetLocalPlayerAsInitialized`, `Player.orderChunks()` capped the chunk radius to `c_s_spawnThreshold` (= `ceil(sqrt(spawn-threshold))`, verified in `Server.java:4016`). When the client's requested view distance (`chunkRadius`, clamped to ≥ 3 in `RequestChunkRadiusProcessor`) exceeded that cap, the outer ring of chunks was never sent, so the client waited for chunks that would never arrive and only proceeded after its own timeout. The fix raises the pre-spawn radius to `max(chunkRadius, c_s_spawnThreshold)` in both `orderChunks()` and the teleport chunk-readiness check in `checkTeleportPosition()`, matching PocketMine-MP and PowerNukkitX behavior. The `spawn-threshold` still governs when `PLAYER_SPAWN` is sent (unchanged at `Player.java:1337`), so the original spawn pacing semantics are preserved.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- Optional: a regression test for the pre-spawn radius logic (e.g. small `spawn-threshold` + large client view distance) would be valuable, though this path is network-level and difficult to cover with the current test harness.

### Notes
- Verified in an isolated worktree at PR head `7b792f8`: `mvn test -Dtest='cn.nukkit.Player*'` — 88 tests, 0 failures/errors, BUILD SUCCESS (matches the author's validation claim).
- Behavior is unchanged when `spawn-threshold` is the larger radius; the fix only enlarges the pre-spawn radius when the client expects more chunks.
- When the client radius exceeds the threshold, slightly more chunks are sent before spawn — this is the intended effect and aligns with PMMP/PNX.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #959, opened 2026-09-15
fix(block): keep the item tag of a shulker box through place and break

w1zardz:
A shulker box only carries its contents and custom name from the item to the block entity and back. Everything else in the item tag is dropped the moment the box is placed: lore and any custom keys written by plugins. A box handed out with its own description comes back from the first break as a plain box.

PowerNukkitX (`BlockUndyedShulkerBox`), Lumi (`BlockShulkerBox`) and Cloudburst Nukkit behave the same way; none of them keeps the rest of the tag.

**Fix**

- `place()` and both shulker box dispense behaviors store the item tag, minus `Items` and `display.Name` that the block entity already owns, under `SourceItemTag` in the block entity;
- `toItem()` starts the drop from that tag, then writes contents and name from the block entity as before, so every drop path (break, silk touch, explosion, piston) returns the original item;
- pick block uses the new `toPickItem()`, which does not copy the kept tag, and strips it from the block entity data it attaches, so creative pick block does not duplicate per-item plugin data while the original box stands in the world;
- the tag is not part of `getSpawnCompound()`, so clients see no difference;
- boxes placed before this change have no `SourceItemTag` and drop exactly as before.

**Reproduce:** give a shulker box with lore (or any custom NBT key), place it and break it — the dropped box has lost the lore/keys.

**Validation:** new `BlockShulkerBoxSourceItemTagTest` (6 tests: kept/removed keys, plain box, break round-trip with and without contents, pick block, legacy boxes). `mvn test -Dtest='cn.nukkit.block.BlockShulker*,cn.nukkit.Player*'` — 96 tests, no failures or errors.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR preserves the full item tag (lore, plugin custom keys) of a shulker box across place and break by storing the non-redundant part of the item tag as `SourceItemTag` in the block entity, and rebuilding the dropped item from it. The implementation is clean, well-documented, and covered by a dedicated test class. Verified locally in an isolated worktree at the PR head: `BlockShulkerBoxSourceItemTagTest` — 6 tests run, 0 failures.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None significant.

### Recommendations
- None. The design correctly excludes `Items` and `display.Name` from the kept tag (both are owned by the block entity and written back on break), keeps the tag server-side only (`getSpawnCompound()` does not send it), and uses `toPickItem()` so creative pick block cannot duplicate per-item plugin data.

### Notes
- Verified: `BlockEntityShulkerBox.getSpawnCompound()` only exposes `facing` and `CustomName`, so the kept tag is never sent to clients.
- Verified: both dispenser behaviors (`ShulkerBoxDispenseBehavior`, `UndyedShulkerBoxDispenseBehavior`) also store the source tag, keeping dispense-place-break consistent with manual placement.
- Legacy boxes placed before this change have no `SourceItemTag` and drop exactly as before (covered by test).
- `BlockUndyedShulkerBox` extends `BlockShulkerBox`, so both dyed and undyed variants benefit from the fix.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #942, opened 2026-09-10
fix(login): verify authentication off-thread within bounded capacity

w1zardz:
`ClientChainData.read` may synchronously resolve DNS and fetch the discovery, OpenID and JWKS documents. A live JFR recording captured a 1.34-second main-thread stall there: every player on the server froze while one client logged in.

Run the unchanged validator on two daemon workers, bounded by 34 outstanding jobs (completed results included) and 128 MiB of retained login bytes. Capacity is reserved before copying and kept until the main thread consumes the result; cancelling an executing DNS lookup never releases its retained memory early. The existing `AsyncTask` completion collector continues the login on the main thread, and closed, disconnected or already advanced sessions cannot consume a late identity.

Signature, issuer, audience, key rotation and token validation are unchanged. The expiry of a validated modern token is rechecked at publication without network I/O. Legacy and offline authentication keep their existing rules. Overload produces an immediate protocol failure instead of running on the caller, and shutdown never waits for DNS.

Validation: all 6 PlayerLoginVerificationTest and LoginChainVerifierTest cases pass. They cover blocked validation, byte and job capacity, cancellation, real JWT signatures, expiry, issuer and audience, deferred publication, stale sessions and protocol overload. No test contacts the authentication service.

PR #960, opened 2026-09-15
fix(player): trim padded login names before they key identity and player data

w1zardz:
Clients can report `"Name "` or `" Name"` for the account that otherwise logs in as `"Name"`. The verified login name is only cleaned of format codes, so the padded form derives its own offline identity and its own player data file. One account then owns two inventories — a login with the other spelling loads the other file while the first keeps its own items. The padded form also slips past name bans, the whitelist, the reserved-name check (`"rcon "`) and the duplicate login check, which compare the name as given.

**Fix**

- The login handler trims the cleaned name (`Player.verifiedLoginName`) before username, display name, name tag, reserved-name check, whitelist/bans, duplicate login check, name lookup table and offline identity see it. Length and characters are still validated on the raw name, so a short name padded to the minimum length keeps logging in. Inner spaces stay. `space-name-mode: replace` keeps its historical `"Name_"` form and identity.
- At startup (only with `savePlayerDataByUuid`, before plugins and the network) `PlayerNameEdgeWhitespaceMigration` folds profiles saved under padded names onto the trimmed identity, but only in the clean case: all involved keys are offline entries pointing at the identity of their own spelling, every key has a readable data file whose `NameTag` trims to the name and which has no XUID. Anything else is left untouched and offline logins with that name are refused with the reason in the log until an admin checks the files; Xbox logins are unaffected.
- A lone padded profile is renamed to the trimmed identity. With several, the most recently played one wins and the others are moved to `players-quarantine` with a warning — contents are never merged. Lookup keys are rewritten in one atomic batch; a failed file move is rolled back and a later start recognises a partially completed run.

**Reproduce:** log in offline as `Name`, then with a client/proxy that sends `Name ` — you get a fresh inventory and a second `players/<uuid>.dat`; a ban on `Name` does not stop `Name `.

**Validation:** `PlayerNameEdgeWhitespaceMigrationTest` (25 tests: clean fold, several twins, never-played ranking, quarantine, unclean states, interrupted runs, rollback), `PlayerVerifiedLoginNameTest` (3 tests for the trimming helper). `mvn test -Dtest='cn.nukkit.Player*,cn.nukkit.Server*,cn.nukkit.utils.ClientChainData*,cn.nukkit.CharacterCreatorLoginJwtTest'` — 146 tests, no failures or errors. The login handler itself has no unit harness upstream, so the in-handler wiring is verified by review, not by a login-packet test.

Note: this touches the same login block as #942 (off-thread login verification); whichever lands second needs a trivial rebase.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary

This PR fixes a real identity-splitting bug: a client reporting `"Name "` or `" Name"` derived its own offline UUID and player data file, giving one account two inventories and letting the padded spelling slip past name bans, the whitelist, the reserved-name check and the duplicate-login check.

The fix has two parts, both well executed:

1. **Login handler (`Player.java`)**: the verified name is trimmed (`verifiedLoginName`) before anything keys on it — username, display name, name tag, reserved-name check, whitelist/bans, duplicate-login check, name lookup and offline identity all see the same string. Length/character validation still runs on the raw name, so a short name padded to the minimum length keeps logging in (documented intent). `space-name-mode: replace` correctly keeps its historical `"Name_"` behavior and is excluded from trimming and from the migration.
2. **Startup migration (`PlayerNameEdgeWhitespaceMigration.java`)**: runs only with `savePlayerDataByUuid`, before plugins and the network. It folds padded profiles onto the trimmed identity only in a strictly-defined clean case (all keys offline and self-consistent, all files readable with matching `NameTag` and no XUID), never merges contents, picks the most-recently-played profile as winner, quarantines the rest, rewrites keys in one atomic batch, rolls back failed moves in reverse order, and recognizes its own interrupted runs on the next start. Anything unclean is left untouched and the name is blocked from offline logins with the reason logged.

### Issues Found

#### Blocking Issues
- None.

#### Non-blocking Issues
- `Server.profileFoldBlockedNames` is a plain mutable field written during startup and read on login threads. Safe in practice (assigned before the network starts), but making it `volatile` would document the cross-thread hand-off.
- The fold-block kick message is hardcoded English (`"Your player data is unavailable..."`) rather than a translation key like the surrounding login refusals (`disconnectionScreen.*`).
- The fold-block refusal runs before the ban/whitelist checks and `PlayerPreLoginEvent`, so a blocked player sees the data-unavailable message even if they are also banned. Cosmetic ordering only.
- When the name lookup table cannot be read at startup, the migration fails open (logged error, no names blocked). Acceptable trade-off, but worth being aware of.

### Recommendations
- Consider a translation key for the new kick message in a follow-up.
- Coordinate the trivial rebase with #942 (off-thread login verification) as the author noted — they touch the same login block.

### Notes
- Verified in an isolated worktree at PR head `e0d35905c`: `mvn test -Dtest='cn.nukkit.Player*,cn.nukkit.Server*,cn.nukkit.utils.ClientChainData*,cn.nukkit.CharacterCreatorLoginJwtTest'` — **146 tests, 0 failures/errors**, matching the author's claim. The new `PlayerNameEdgeWhitespaceMigrationTest` (25 tests: clean fold, twin ranking, never-played ranking, quarantine, unclean states, interrupted runs, rollback) and `PlayerVerifiedLoginNameTest` (3 tests) all pass.
- ClamAV scan of all 5 changed files: clean.
- The migration's design (clean-case-only, quarantine instead of merge, atomic key batch, rollback, blocked-name refusal) is conservative in the right way for player-data surgery.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #962, opened 2026-09-16
feat: support 1.26.50

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Adds multiversion support for 1.26.50 (protocols 2192 preview / 2193 stable): new GameVersion entries, protocol constants, packet ID 351/352 registration, wire-format updates across ~20 packets, new runtime resources (block properties, runtime block/item states, voxel shapes, item components), and CB cross-validated regression tests. Codex CLI was unavailable (API key disabled), so this is a manual review against the diff, the Cloudburst codec bytecode, and the full test suite.

### Issues Found
#### Blocking Issues
- `ItemStackResponsePacket#encode` — regression for v2168/v2169 (1.26.40/45) when `containers` is empty. Old code wrote `[true, false]` (constant-true outer bool + presence bool), matching Cloudburst's `ItemStackResponseSerializer_v2168` bytecode (`writeBoolean(true)`; if empty → `writeBoolean(false)` + return). The new code writes only `[false]` for empty containers regardless of version, so a 1.26.40/45 client reads the first bool as the outer constant and then consumes one byte from the next entry (or past the end of the packet) — wire corruption for every failed item stack request on existing versions. The empty-container branch should keep the version split:
  ```java
  if (r.getContainers().isEmpty()) {
      if (this.protocol < ProtocolInfo.v1_26_50_27) {
          this.putBoolean(true);
      }
      this.putBoolean(false);
      return;
  }
  ```

#### Non-blocking Issues
- `ClientboundUpdateSoundDataPacket#encode` writes `SOUND_STOP` in slot 0 unconditionally, ignoring the `stop` field — asymmetric with `decode()` (which maps slot 0 type `SOUND_STOP` → `stop=true`), so a self-round-trip with `stop=false` decodes as `stop=true`. Currently unused server-side (packet is only registered, never sent), but worth fixing before the packet is used.
- `Player.java` now sends a cached empty `JigsawStructureDataPacket` to all clients ≥ v1_21_120 (not just v2192). Intentional per comments (vanilla payload is huge), but it is a login-behavior change for several existing versions — worth a smoke test on e.g. 1.21.120 / 1.26.40 clients.

### Recommendations
- Fix the v2168 empty-container encoding above and add a CB cross-decode regression test for v2168 with empty containers (current V2192 tests only cover v2192 for this case).
- Consider covering the `ClientboundUpdateSoundDataPacket` slot round-trip in a test once the packet is actually used.

### Notes
- Verified against Cloudburst `bedrock-codec 3.0.0.Beta13-20260916` bytecode: v2193 empty-container encoding (`[false]`) matches this PR; v2168 does not (see above).
- Full test suite run in an isolated worktree at PR head `37748bd`: 16741 tests passed (0 failures, 0 errors; surefire fork then died with an environmental Java heap OOM at the tail of the suite) + the remaining `cn.nukkit.utils.**` classes rerun separately: 226 passed. All PR-added tests pass: `GameVersionTest` (7), `V2192LoginDataTest` (3), `V2192PacketRegressionTest` (15), `PacketCoverageReportTest` (1).
- ClamAV scan of all new resource files (.nbt/.dat/.bin/.json): clean. No suspicious code patterns found.
- Protocol plumbing (GameVersion/ProtocolInfo/RuntimeItems/GlobalBlockPalette/Level/CraftingManager thresholds, Utils version strings) is consistent for both 2192 and 2193.

---
🤖 AI Review

PR #963, opened 2026-09-17
feat: add NetherNet support

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Adds a NetherNet (WebRTC) transport that runs alongside RakNet: an HTTP signaling endpoint (sharing the server port over TCP) accepts SDP offers, and each peer becomes a `NetherNetChildChannel` carrying the same batch payloads as RakNet. The change includes a new `NetherNetInterface` / `NetherNetPlayerSession`, per-session ingress token buckets and backpressure mirroring `RakNetPlayerSession`, a signaling-identity binding check in place of the skipped Bedrock encryption handshake, a new `network-settings.nethernet` config section with localized comments for all six languages, dependency wiring (netty-transport-nethernet + libdatachannel arch-detect), and two new tests (signaling smoke test + nested config comment round-trip). `mvn compile` passes on this branch.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `NetherNetSettings.enabled` defaults to `true`, so upgrading servers will suddenly bind a TCP listener on `server-port`. Hosts that already use TCP on that port (tunnels, proxies, panel integrations) will hit a bind conflict on startup. The failure path is handled gracefully (logged fatal, server continues without NetherNet), but a conservative default or a release-note callout may be worth considering.
- `NetherNetSignalingSmokeTest` loads the native libdatachannel library. On platforms/architectures not covered by the bundled arch-detect natives (e.g. unusual CI runners), this test will fail at `LibDataChannelArchDetect.initialize()`. Consider an assumption-based skip when the native library is unavailable.
- `NetherNetInterface.shutdown()` and `emergencyShutdown()` are byte-for-byte identical; if that is intentional, a short comment noting they deliberately behave the same would help future readers.
- `NetherNetInterface.eventLoopGroup` is a fixed 2-thread NIO group. Fine for a secondary transport, but there is no way to tune it via config if a server ends up with heavy NetherNet traffic.

### Recommendations
- Mention the new TCP listener on `server-port` in the changelog/release notes, since it is a behavioral change visible to hosts even when no NetherNet client ever connects.
- If the native-library smoke test proves flaky on CI, gate it behind an availability check rather than deleting it — it is good coverage of the signaling path.

### Notes
- The login-chain → signaling identity binding in `Player` (`checkIdentityBinding`) correctly addresses the replay risk created by skipping the Bedrock encryption handshake; encryption is also explicitly disabled for NetherNet sessions, matching the DTLS-carried design.
- The ingress rate limiting, malformed-batch handling, and fair round-robin drain closely follow the hardened `RakNetPlayerSession` logic, so the new transport inherits the existing DoS protections.
- `ConfigComments` now recurses into nested sub-configs; this is a generic improvement covered by `ConfigCommentsTest`.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #965, opened 2026-09-17
fix(inventory): close ender chest and container windows on -1, death and break

w1zardz:
Three commits around container windows, all visible to players as a chest lid that hangs open, shuts on the wrong player, or a window that stays in front of a block that is gone. They build on each other, so they come as one branch.

### 1. Send the ender chest lid event to the chunk of the chest

`PlayerEnderChestInventory` addressed the lid `BlockEventPacket` and the open/close sound to the chunk of its holder — the player. `teleport()` calls `removeAllWindows()` after the player has moved, so a viewer teleported or respawned with the chest open closed it from the destination and the lid stayed open for everyone around the chest. The events now go to the level and chunk of the chest.

### 2. Share ender chest viewers and close its window on -1 and on death

- A 1.21 client closes the ender chest with `ContainerClose` windowId **-1** while its own inventory is not open. The -1 branch only reset the crafting grid: the window stayed open on the server, the lid hung open and `addWindow()` returned the stale id on the next click, so the chest did not open again for that player until a teleport.
- `Player.kill()` closed no windows. A player who never pressed respawn kept the lid open, and windows that hand input slots back on close (trade, anvil, grindstone) returned them into the respawned inventory instead of the death drops. Windows are now closed before drops are collected, as in vanilla.
- `BlockEnderChest` kept viewers in a field, but a `Block` is created per lookup, so each opener had a set of one: the first to close shut the lid for the second. Viewers now live on `BlockEntityEnderChest`; the close event is sent only when the set becomes empty.

### 3. Close world container windows on -1 and when the container breaks

- The -1 branch now also closes windows whose holder is a block entity that still stands in its level (chest, barrel, shulker box, hopper, furnace…). Windows drawn on fake blocks (`FakeBlockMenu`, plugin menus) have no block entity and are left alone, so a late -1 cannot shut a freshly reopened plugin menu.
- Breaking a container: the ender chest window is now closed by `BlockEntityEnderChest.close()`; `BlockEntityChest.onBreak()` closes the double chest windows before unpairing (previously the pair was broken first and viewers kept a window of a chest that no longer existed); dispenser and dropper get the `close()` every other container already had.

Compared with other implementations: Allay counts ender chest viewers on the block entity and closes containers on -1 (taken here, limited to real block entities); PowerNukkitX already addresses the lid to the chest position but keeps the -1 and double chest gaps; Lumi carries all defects.

No new packets or fields: closing uses the same `ContainerClosePacket` / `BlockEventPacket` as before.

**Validation:** `PlayerEnderChestLidTest`, `PlayerEnderChestViewersTest`, `PlayerEnderChestCloseTest`, `PlayerWorldContainerCloseTest`, `ContainerBreakClosesWindowsTest` — all fail on the previous code and pass here; full test suite has no new failures (the one failing test, `LevelDBAsyncChunkSupportTest.deferredBlockUpdatesAreSerializedBeforeChunkInitialization`, fails on current master too).


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Three related fixes around container windows: ender chest lid events are now addressed to the chest's chunk instead of the holder's, ender chest viewers live on the block entity instead of a per-lookup `Block` instance, and windows are closed on player death, on client `ContainerClose` with window id -1, and when the container block breaks. The approach is sound, well documented, and covered by five new test classes. Verified locally: all 14 new tests pass on the PR head (`8db8a72c`). One gap in the new -1 handling remains.

### Issues Found
#### Blocking Issues
- **Double chest windows are not closed on `ContainerClose(-1)`.** `Player.closeWorldContainerWindowsForUnknownId()` only matches windows whose holder is a `BlockEntity` (`isWorldBlockContainer`). A `DoubleChestInventory` is its own holder (`this.holder = this`), so when a 1.21 client closes a double chest with window id -1 — the same scenario this PR fixes for single chests, barrels, shulker boxes, hoppers and furnaces — the double chest window stays open on the server: the lid hangs open and `addWindow()` returns the stale window id on the next click, so the chest cannot be reopened until a teleport. That is exactly the bug class this PR eliminates for the other containers. Suggest also matching a `DoubleChestInventory` whose left/right block entities still stand in their level.

#### Non-blocking Issues
- `BlockEntityEnderChest.close()` compares only XYZ when deciding whether to clear `player.viewingEnderChest`; ender chests at the same coordinates in different levels would interfere. Comparing the level (or block entity identity) would be safer.
- `Player.kill()` closes windows via `removeAllWindows()`, which does not fire `InventoryCloseEvent` (the new -1 paths fire it manually). No regression — windows were not closed on death at all before — but plugins listening for close events will not observe death-triggered closes.

### Recommendations
- Extend the -1 handling (and `PlayerWorldContainerCloseTest`) to cover double chests.
- Optionally fire `InventoryCloseEvent` for death-triggered closes for consistency.

### Notes
- Handling `addWindow() == -1` (cancelled open event) in `BlockEnderChest.onActivate` by dropping the viewer is a good catch.
- The duplicate `super.onClose(who)` call in `PlayerEnderChestInventory.onClose` was removed — good.
- No new packets or protocol surface; closing reuses `ContainerClosePacket` / `BlockEventPacket`.

---
🤖 AI Review

tachibana-kanade-angel:
This PR has been superseded by #976, which contains the same changes (the ender chest / container window close handling) along with additional fixes, rebased onto the current master.

PR #968, opened 2026-09-17
perf(collision): read one complete block state during collision scans

w1zardz:
> **Depends on #957** (retain fresh builtin block snapshots during collision scans). Please review only the last commit.

Standard collision scans read a block ID and its metadata through two separate section lookups — two read locks and two palette resolutions per cell on the hottest path of entity movement.

**Change:** for exact built-in LevelDB storage, read the ID and metadata under one section read lock and one palette lookup. All 32 bits of both fields are kept. Custom levels, providers, chunks and sections keep the original fallback, custom factory snapshot ownership from #957 is retained, and collision shape rules are unchanged.

**Validation:** `CollisionStatePairTest` (signed and extended states, storage guards, height bounds, custom factories) and `CollisionHelperBlockOwnershipTest`; full test suite has no new failures (the one failing test, `LevelDBAsyncChunkSupportTest.deferredBlockUpdatesAreSerializedBeforeChunkInitialization`, fails on current master too).


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR optimizes the standard collision scan hot path (`CollisionHelper.getBlocksInBoundingBox`): for exact built-in LevelDB storage, block ID and metadata are now read under a single section read lock with one palette lookup (`LevelDBChunkSection.getBlockStatePair`), instead of two separate locked lookups. All 32 bits of both legacy fields are preserved via `long` packing. Custom levels, providers, chunks, and sections keep the original fallback path, and collision shape rules are unchanged.

Note: this review covers the head commit `9c9ab67` ("Read one complete block state during standard collision scans") as requested; the first commit belongs to #957.

(Automated review note: the code review backend was unavailable, so this review was performed by direct diff analysis plus a local test run.)

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None of significance. The `long` packing in `getBlockStatePair` (id << 32 | data) relies on implicit bit-layout knowledge; a brief named constant or comment for the shift width would marginally help readability, but the existing javadoc already covers the intent.

### Recommendations
- Merge order: this PR depends on #957 (currently open). Merging #957 first keeps the history clean; GitHub will otherwise bundle both commits here.

### Notes
- Equivalence verified: the fast path mirrors the existing `getBlockState(x,y,z,layer)` read-lock discipline exactly, minus the temporary `int[]` allocation. Air probing via the packed pair is consistent with the previous `isAirAt` + `chunk.getBlockId` path (both read the same palette snapshot under the same read lock).
- Guards are strict: exact class checks (`Level.class`, `LevelDBProvider.class`, `LevelDBChunk.class`, `LevelDBChunkSection.class`), provider identity (`chunk.getProvider() == level.getProvider()`), and chunk coordinate match. Any deviation falls back to the original path, including `EmptyChunkSection` (all-air short-circuit, equivalent to the old probe).
- Y bounds are clamped to `level.getMinBlockY()/getMaxBlockY()` before section indexing, so `getSection(y >> 4)` stays within the dimension's section range — same assumption as the existing code.
- Test coverage is thorough: signed/extended 32-bit id+meta round-trip, absent-layer → air, single-read verification on the fast path, custom chunk/section/provider fallbacks, foreign-provider hint, bounds exclusion, and custom-factory cached result cloning.
- Local verification: `CollisionStatePairTest` and `CollisionHelperBlockOwnershipTest` executed against the PR head — results confirmed passing.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #969, opened 2026-09-19
fix(block): state-driven connections for stairs, panes and fences in 1.26.50

lt-name:


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR migrates stairs, fences, panes/bars and tripwire to the state-driven connection/corner model introduced in 1.26.50 (protocol 2193), where the client no longer derives shapes from neighbours. Connection/corner bits are encoded into the upper meta bits (0x10-0x80 for connections, bits 3-5 for stair corners), maintained on place/onUpdate, backfilled by new LevelDB block-state updaters (1.26.30/1.26.50), and a chunk-load migration path recomputes the bits for legacy worlds (gated on horizontal neighbours being loaded, retried from the mount tail and as a send-path fallback). Older-protocol palettes and pre-419 lookup tables get a masked-meta fallback so the new bits degrade to the base meta instead of collapsing to data=0 or bleeding into the id bits. `FEATURE_VERSION` / `STATE_PATCH_VERSION` are bumped to 1.26.50. The new tests (`BlockPaletteLegacyMetaFallbackTest`, `BlockStateMetaAlignmentTest`, `LevelDBConstantsTest`, `LegacyConnectionFixFlagTest`) all pass locally: 26 tests, 0 failures, BUILD SUCCESS.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `BlockPalette.getMaskedMeta` / `GlobalBlockPalette.maskedLegacyLookup` apply to all blocks, not just connection-driven ones: any block whose full meta misses will now degrade to `meta & 0x0F` / `meta & 0x07` before the meta-0 fallback. This changes the previous "fall straight to data=0" behaviour for unrelated blocks with out-of-range meta. The direction (preserving base meta) is reasonable, but it is a broader semantic change than the PR title suggests — worth keeping in mind if plugins rely on the old fallback.
- The first lookup inside `maskedLegacyLookup` re-queries the exact same key the caller just missed on (the meta is already masked to the field width by the caller). Harmless, but redundant.
- The stair corner algorithm (`computeCorner`, translated from JE `StairBlock.getStairsShape`) is only covered by palette-mapping tests, not by in-world neighbour scenarios. If a left/right rotation is flipped it would mirror all corner stairs for 2193+ clients — a quick in-game spot check of inner/outer corners (both half types) is recommended before release.
- New wildcard imports (`it.unimi.dsi.fastutil.ints.*` in `Level.java`, `org.cloudburstmc.nbt.*` in `StateBlockStorage.java`) — minor style point.

### Recommendations
- Spot-check stair corners and pane/fence connections in-game with a 1.26.50 client, including across chunk borders and in a world saved before this change (verifies the migration flag path).
- Consider scoping the masked-meta fallback to the known connection-driven block ids if the broader fallback change turns out to affect other blocks.

### Notes
- Meta-bit layouts checked for collisions: tripwire uses bits 0/2/3, stained pane colour is read via `DyeColor.getByWoolData` (masks `& 0x0F`), fence wood type masks `& 0x07` — the new upper bits do not collide with existing uses.
- Copper stairs/bars variants correctly forward non-random updates to the parent so corner/connection bits stay maintained while oxidation keeps the random tick path.
- `toItem()` masks off connection bits, so picked/dropped items keep clean damage values.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #970, opened 2026-09-19
feat(nethernet): relay WebRTC media through the RakNet port

lt-name:
Setting server-udp-ports=19132:19134 (a single-port mapping whose external port is server-port) now shares RakNet's UDP port with NetherNet instead of expecting an external NAT: libjuice binds its ICE mux to loopback:19134 and NetherNetMediaRelay demultiplexes STUN/DTLS off the RakNet listener (RFC 7983 first byte plus the STUN magic cookie), forwards each client over its own loopback leg and writes replies back through the RakNet socket, so only 19132/udp and 19132/tcp need to be reachable.

- legs open only for a STUN Binding Request carrying the local ufrag of a pending answer, with total/per-address caps and a 60s idle sweep
- answers advertise the server's reachable addresses on the listener ports (server-ip when bound, else enumerated interfaces); offers lose their candidates since the loopback-bound socket cannot reach them
- /status gains a relay line, windows covering server-port and a missing RakNet listener abort startup with localized messages
- NetherNetDelegatingSignaling dedupes the signaling decorators

The default stays 19134; docs describe the new mapping.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Shares the NetherNet (WebRTC) media path with RakNet's UDP port: when `server-udp-ports` equals a RakNet listener port (new default `19132` = server-port) or maps a single port onto one, a new in-process `NetherNetMediaRelay` demultiplexes STUN/DTLS off the RakNet listener (RFC 7983 first byte + STUN magic cookie) and forwards each client over a loopback leg to libjuice's ICE mux port. SDP answers are rewritten to advertise the server's reachable addresses on the listener ports, offers lose their unreachable candidates, and the internal loopback port is reserved until the first offer. Docs, config comments and lang files updated in 6 languages; `/status` gains a relay line.

Change is well-scoped and thoroughly tested: 59 tests across 5 test classes pass (`mvn -o test` on the PR head), including real-loopback round-trip, ufrag gating, idle sweep, reservation lifecycle and shutdown detach. ClamAV scan clean; no new dependencies, no binary/archive files.

Codex was unavailable (API key disabled); manual review performed.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `NetherNetInterface` constructor: if the signaling builder or another step throws *between* relay creation and the bootstrap bind, the relay and its reservation socket leak (only the bind-failure path calls `relay.shutdown()`). Very low probability; a try/catch around the remaining construction would close it.
- `NetherNetSharedPortSdp.rewriteAnswer` rewrites the `m=application` port only when an IPv4 external port exists; an IPv6-only deployment keeps the internal port on the `m=` line. Nominal in practice (candidates rule the media path), but worth noting.
- `NetherNetMediaRelay.classify`: a RakNet datagram ≥20 bytes with first byte ≤3 and bytes 4–7 equal to the STUN magic cookie (~1/2³² per packet) would be misrouted. Harmless — it is dropped without a matching ufrag — and already called out in the code comments.
- `Leg.pending` buffers datagrams until the loopback connect completes with no explicit cap; the window is sub-millisecond on loopback, so this is theoretical.

### Recommendations
- Consider wrapping the relay creation → bind section of the `NetherNetInterface` constructor in try/catch to release the relay on any failure, not just bind failure.
- The `server-udp-ports` default change (19134 → 19132) is a behavioral change for fresh installs only; existing `server.properties` keep their persisted value, so no migration concern. Good that README/config comments were updated in sync.

### Notes
- Leg opening is gated on the ICE ufrag of a pending answer, with total/per-address caps, a 60s idle sweep and the address blocklist hook — the amplification/forgery surface is small.
- Existing deployments that pinned `server-udp-ports=19134` explicitly keep the standalone-port behavior; window mappings that would cover a listener port now abort startup with a clear localized error instead of silently misconfiguring.
- Title already follows Conventional Commits (`feat(nethernet): ...`); no normalization needed.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #971, opened 2026-09-19
fix(level): compare the sleeping share as a percentage

w1zardz:
`Level.checkSleep` compares the share of sleeping players against the
`playersSleepingPercentage` game rule, but computes that share in integer
arithmetic:

```java
if (playerCount > 0 && sleepingPlayerCount / playerCount * 100 >= this.gameRules.getInteger(GameRule.PLAYERS_SLEEPING_PERCENTAGE)) {
```

`sleepingPlayerCount / playerCount` is an int division, so it is `0` for every
case except "everybody in the level sleeps", where it is exactly `1`. Effects:

* the game rule does nothing at all — 1 and 99 both behave like 100, because
  the left side can only ever be `0` or `100`;
* a value above 100 makes the night unskippable even with the whole level
  asleep;
* a server that lowers the rule to let one player skip the night silently gets
  vanilla 100 % behaviour and no way to notice why.

Multiplying before dividing gives the real percentage. With the default 100 the
behaviour is unchanged (`sleeping * 100 / players >= 100` is still "all of
them"), so this only starts to matter once a server sets the rule.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Single-line fix in `Level.checkSleep()`: the sleeping-player share was computed as `sleepingPlayerCount / playerCount * 100` (int division), which can only ever yield `0` or `100`, making the `playersSleepingPercentage` game rule useless for any value other than 100. Reordering to `sleepingPlayerCount * 100 / playerCount` computes the real percentage. No blocking issues found. The changes look reasonable overall.

(Note: automated Codex review was unavailable due to API key issues; this review was performed manually. ClamAV scan of the diff: clean.)

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- None

### Recommendations
- None — the fix matches vanilla semantics and keeps behaviour identical at the default rule value of 100.

### Notes
- Overflow is not a concern: `sleepingPlayerCount <= playerCount`, and `playerCount * 100` cannot approach `Integer.MAX_VALUE` under any realistic player count.
- Verified against surrounding code: both counters are `int`, so the reordering is the minimal correct fix.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:
This PR has been superseded by #980, which contains the same changes (the sleeping-share percentage comparison) along with additional fixes, rebased onto the current master.

Closing this PR to avoid double-applying these changes. If any part of this PR is not covered by #980, feel free to reopen or comment and we will split it out.

PR #972, opened 2026-09-19
fix(player): take the nether portal target once and drop a cancelled transfer

w1zardz:
A vanilla portal transfer is decided in `Player.onUpdate`, then carried out two
hops later: an async task looks the far side up, and a main thread task moves
the player. Both hops read the `portalPos` field, and the main thread clears
that field the moment the player stops standing in the portal — a plugin
teleport, a `/spawn`, a death, a minigame transfer.

The scheduled lambda then called `BlockNetherPortal.spawnPortal(null)`:

```
Cannot invoke "cn.nukkit.level.Position.getLevel()" because "pos" is null
Could not tick level
```

The level tick dies from there on, so chunk sending stops for everyone in that
world until a restart: teleports hang, players stand still, and nothing in the
log names the portal as the cause. We hit this on a live server; the report read
"I get teleported and then I am stuck in a pillar".

This takes the target once, while it is still known to be non-null, and lets
both tasks use that local copy. The world the player was taken from is captured
with it, so a player who left that world (or went offline, or died) before the
lookup finished is not dragged into the nether seconds after asking to leave.
`portalPos` is cleared before that check, so a cancelled trip cannot leave the
field occupied and block the next portal step.


tachibana-kanade-angel (approved):

tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR fixes a real crash in the vanilla nether portal transfer flow: `portalPos` is cleared by the main thread as soon as the player leaves the portal (plugin teleport, `/spawn`, death), but the scheduled transfer lambda still read the field and called `BlockNetherPortal.spawnPortal(null)`, killing the level tick with an NPE and breaking chunk sending for everyone in that world until restart.

The fix captures the portal target and the origin level once, at decision time, and has both the async lookup task and the main-thread transfer task use those local copies. `portalPos` is cleared before the validity check so a cancelled trip cannot block the next portal attempt. A player who left the origin world, went offline, or died before the lookup finished is no longer dragged into the nether.

The change is minimal (one method in `Player.java`), compiles cleanly (`mvn compile` passes), and the cancellation semantics are sound.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- When `portalTarget == null` at decision time (reachable in creative mode if the player enters the portal between the arming tick and the transfer tick), the method returns after `inPortalTicks` was already set to 81. The player then stands in the portal without re-arming until they exit and re-enter. This is a rare edge case and strictly better than the previous behavior (a dead level tick), but moving the null check before `inPortalTicks = 81` would make it cleaner.
- `level != portalOrigin` relies on reference identity of `Level` instances. This is correct for Nukkit's per-world singleton levels, just noting it is an identity comparison by design.

### Recommendations
- Optionally hoist the `portalTarget == null` check above `this.inPortalTicks = 81;` so a no-op decision does not consume the portal countdown.

### Notes
- Verified the fix against the surrounding portal logic: `portalPos` is armed at `inPortalTicks == 40` (or 10 in creative) and cleared whenever the player stops standing in a portal, which confirms the race described in the PR body.
- No test coverage exists for this scheduler-based flow; the manual reasoning and a clean compile are the available verification.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel:

PR #973, opened 2026-09-20
fix(animatepacket): incorrect action field type for pre-1.2 protocols

yefengeeeeeeeeeee:
Refactor action handling in AnimatePacket for protocol versions.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
This PR intends to fix the `AnimatePacket` action field type for pre-1.2 protocols (unsigned varint → varint), which looks correct. However, the change also removes the `else` branch that handles all protocol versions from v1.2.0 up to (but not including) v1.21.130 — i.e. every modern client version. This is a blocking regression.

### Issues Found
#### Blocking Issues
- **Removed `else` branch breaks all protocols in [v1.2.0, v1.21.130)**: In both `decode()` and `encode()`, the final `else` branch (reading/writing the action as `VarInt` for v1.2.0 ≤ protocol < v1.21.130) was deleted. After this change:
  - `decode()`: `action` is never read for these versions, stays `null` → falls back to `NO_ACTION`, and the remaining fields (`eid`, `data`, `rowingTime`) are read from the wrong offsets, completely desynchronizing the packet stream.
  - `encode()`: the action field is never written, producing malformed packets for every client below 1.21.130.
- The fix should keep the varint path for the [v1.2.0, v1.21.130) range, e.g. merge the legacy branch: `else { this.action = Action.fromId(this.getVarInt()); }` (and `this.putVarInt(...)` in `encode()`), while only correcting the pre-1.2 unsigned-varint behavior if that was the actual intent.

#### Non-blocking Issues
- None.

### Recommendations
- Restore the `else` branch (or widen the `else if (protocol < ProtocolInfo.v1_2_0)` change so that varint is still used for all protocols below v1.21.130).
- Add a round-trip encode/decode test for a mid-range protocol version (e.g. v1.20.x) to catch this class of regression.

### Notes
- Only `src/main/java/cn/nukkit/network/protocol/AnimatePacket.java` is modified.

---
🤖 AI Review

PR #975, opened 2026-09-20
fix: block state round-trip - padding rows, stale default state, legacy render methods and skulls

w1zardz:
Five independent fixes around translating block states between the LevelDB palette, the network palette and Nukkit's legacy id/meta pairs. Without them a server that registers custom blocks can stop ticking a level entirely.

the reproduction.

- **level: translate a block state back to its real legacy value, not to a padding row**

  The legacy data table pads every block to a power of two by repeating its default 
  variant: minecraft:planks lists oak at data 0, 6 and 7, minecraft:stone lists stone at 0
  and 7, and 45 more blocks do the same. Every one of those rows maps to the same state,
  so runtimeIdToLegacy kept whichever row happened to register first, and
  stateHashToLegacy did the same.

  When a padding row won, a state translated back to a data value Nukkit has no block for.
  Oak planks came back as 5:7 and read as an unknown block: hardness 10, no proper tool.
  Level#useBreakOn then measured every hit against that hardness, found it far too fast
  and cancelled it, so the block could not be broken at all — by any tool or by hand — and
  nothing was written to the log, because from the server side each refusal is an ordinary
  anti-fast-break verdict. Which blocks were affected depended on registration order, so
  it changed with the set of custom blocks registered on the server.

  The lowest value is the real one in every padded row, and picking it makes the reverse
  lookup independent of the order states are registered in. Blocks that vanilla merged
  into another id keep their explicit preset (see NukkitLegacyMapper#getOverrideLegacyId),
  so the cauldron round-trip is unchanged.

- **custom-block: translate legacy render method names for new clients**

  Clients since the 1.21.80 material overhaul silently ignore the legacy
  "alpha_test" render method name in a network-defined block: the material falls back to
  opaque and every transparent texel renders as a black pixel. No warning appears anywhere
  - the texture itself still applies, so the block looks almost right and the defect
  surfaces only on the first custom block with actual transparency in its texture.

  Translate the render method per protocol in the definition serializer, next to the
  existing face_dimming/ambient_occlusion conversions: for 1.21.110+ clients "alpha_test"
  becomes "alpha_test_single_sided". Old clients keep the legacy name they understand.

- **block: render split heads through legacy skull palettes**

- **block: preserve split head drops when their tile is missing**

- **leveldb: drop the cached default block state when the palette is rebuilt**

  CustomBlockManager#recreateBlockPalette rebuilds the whole block palette once the custom
  block registry closes: it calls BlockPalette#clearStates and 
  BlockStateMapping#clearMapping, re-sorts every vanilla state together with the custom
  ones and hands out fresh runtime ids. BlockStateMapping caches its default state
  (minecraft:info_update) in defaultRuntimeId/defaultState the first time anything asks
  for it, and clearMapping never reset those two fields.

  A server that registers custom blocks and resolves the default before the registry 
  closes therefore keeps a runtime id from the previous palette. That id now belongs to
  some other state, and it may be one of the ~2400 states the palette carries without a
  legacy mapping at all. Every lookup of an unknown state then reaches the default
  fallback, finds -1 there too and raises

    java.lang.IllegalArgumentException: Can not find legacyId for default runtimeId: 1623

  from BlockStateMapping#getLegacyId. On a live server this arrives inside 
  Level#tickChunks through StateBlockStorage#get, so the whole level stops ticking - 14315
  warnings and 567 aborted level ticks in one day on our survival world after updating to
  1.26.50, where the shrunken leveldb_palette.nbt moved the state that used to sit on the
  stale id.

  clearMapping now drops the cached default along with the mappings it belongs to, so the
  next caller resolves it against the rebuilt palette. As a second line of defence, the
  three legacy fallbacks re-resolve a default that has lost its legacy mapping instead of
  aborting the level tick; only a default that cannot be resolved at all still raises.


tachibana-kanade-angel:
## 🤖 AI Review

> Codex was unavailable (API key disabled); manual review performed against the full diff and the surrounding code on the PR branch.

### Summary
Five independent, well-scoped fixes for block state round-trip issues between the LevelDB palette, the network palette and legacy id/meta pairs. Each commit is self-contained, documented with the production incident that motivated it, and covered by new tests (`LegacySkullCompatibilityTest`, `SkullDropRecoveryTest`). The changes look reasonable overall.

Verified in detail:

1. **Padding rows (`BlockPalette.putCanonicalLegacy`)**: Reverse lookups (`runtimeIdToLegacy`, `stateHashToLegacy`) now keep the *lowest* legacy value instead of whichever padded row registered first. The maps use `defaultReturnValue(-1)`, so the `known == -1` guard is correct, and choosing the minimum makes the reverse lookup registration-order independent. This fixes padded variants (e.g. oak planks coming back as `5:7`) resolving to unknown/unbreakable blocks.
2. **Legacy render method translation (`CustomBlockDefinitionSerializer843`)**: `alpha_test` → `alpha_test_single_sided` is applied only for protocol ≥ 843 (1.21.110+), next to the existing `face_dimming`/`packed_bools` conversion; older clients keep the legacy name. Correct placement.
3. **Legacy skull aliases (`BlockPalette.registerLegacySkullAliases`)**: Outgoing aliases for split heads are added in `lock()` only when the palette has no native state for the head fullId, and `clearStates()` resets `locked` so a palette rebuild re-registers them without overwriting native states. Reverse lookups and cache invalidation are handled correctly. Covered by `paletteRebuildRecreatesAliasesWithoutOverwritingNativeStates`.
4. **Split head drops (`BlockSkullSkeleton.getDrops`)**: When the block entity is missing, the drop meta falls back to `getSkullType().ordinal()`, which matches the `ItemSkull` damage convention (0 skeleton … 6 piglin); plain skulls keep the old `0` default, and an existing tile's `SkullType` still wins. Behavior is pinned by `SkullDropRecoveryTest`.
5. **Stale default state (`BlockStateMapping.clearMapping` + `getMappedDefaultRuntimeId`)**: `clearMapping` now drops the cached `defaultRuntimeId`/`defaultState`, and the three legacy fallbacks re-resolve a default that lost its legacy mapping instead of aborting the level tick with `IllegalArgumentException`. Matches the reported production failure mode.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `CustomBlockDefinitionSerializer843`: the `alpha_test` translation only applies to protocol ≥ 843. Clients on 1.21.80–1.21.100 (protocols below 843) are equally affected by the material overhaul's silent fallback to opaque, but no serializer exists for those protocols, so they still receive the legacy name. Consider extending coverage if those protocol versions are supported targets.
- `BlockStateMapping`: `defaultRuntimeId`/`defaultState` are plain (non-volatile) fields and `getMappedDefaultRuntimeId()` re-resolution is not synchronized. Concurrent re-resolution is benign because `setDefaultBlock` is idempotent, but a short comment or `volatile` would make the thread-safety intent explicit.
- Security scan: ClamAV scan of all 7 changed files — clean; no binary/archive files, no suspicious patterns (no `Runtime.exec`, reflection abuse, or network calls).

### Recommendations
- Consider a follow-up for pre-843 clients if `alpha_test` transparency matters on 1.21.80–1.21.100.
- The included tests already cover the round-trip invariants across legacy and modern palettes; no additional test requests.

### Notes
- The five commits are independent and can be merged as-is or dropped individually without conflict, as the author states.
- Verified on the PR head (`80184cd`): `mvn test -Dtest=LegacySkullCompatibilityTest,SkullDropRecoveryTest` — 6/6 passed, BUILD SUCCESS.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #976, opened 2026-09-20
fix(inventory): stack request decoding, partial merges, durability and ender chests

w1zardz:
Nine independent fixes in the inventory and item-transaction path: stack request decoding, merges that used to be refused whole, durability packets sent for non-durable items, eating interrupted by a repeated click, and ender chest and world container window handling.

the reproduction.

- **inventory: decode a stack request the way the live network does**

  Two gaps in the item stack request decoder turn a normal craft into a decode failure,
  and a decode failure used to cost the player the whole session.

  An ingredient descriptor that names an item the server cannot resolve went straight into
  Item.fromString(...).getId(), so an identifier we do not know threw a
  NullPointerException out of the decoder. An unknown identifier now decodes as air and
  the request is judged by the handler, which already answers ERROR and resynchronises the
  client.

  The recipe book containers (64, 65, 66) were missing from ContainerSlotType, so a slot
  info that names one of them left fromId() empty and the decoder threw. They are present
  in the protocol fork the production network runs on. Older protocols are untouched:
  their branches already reject every id above DYNAMIC_CONTAINER, and the encoder refuses
  to write the new ids below v1_21_20.

- **inventory: move what fits when a stack merge overflows**

  A TAKE/PLACE whose destination cannot hold the whole moved amount answered ERROR and
  moved nothing. The client rolls such a request back, so both slots stay exactly as they
  were and the player sees two stacks of the same item that refuse to merge. It shows up
  first on items with a small stack limit — ender pearls, eggs, snowballs, honey bottles —
  where a merge overflows almost every time.

  Vanilla moves as much as the destination can hold and leaves the remainder on the source
  side. Do the same: clamp the transferred count to the remaining room, recompute whether
  the source is emptied, and keep answering ERROR only when the destination is already
  full and nothing can move at all.

- **inventory: send durability correction only for durable items**

  Item#getDamage stores both real wear and legacy variant metadata. Item stack responses
  sent that metadata as durability for every item, so clients treated separately crafted
  plank batches as different even though the server considered them equal.

  Use ItemDurable as the protocol boundary in every response-slot builder. Non-durable
  variants now send zero while tools and armour retain their real damage. Add regressions
  for planks, potions and a damaged pickaxe.

- **player: keep eating a consumable through a repeated click and a sprint**

  A consumable is finished by the server: entityBaseTick() counts the ticks since 
  startAction and calls Item#onUse once the use duration is reached. Anything that clears
  the use state before that silently throws the food away - the client keeps playing the
  animation, the item is never consumed and no health is restored. In a fight that reads
  as "I ate a steak and still died", and nothing is logged.

  Three paths cleared the state while a player was eating:

  * a repeated USE_ITEM_ACTION_CLICK_AIR. The client sends it while the button is
   held down, and a player in a fight taps the button instead of holding it. The
   transaction arrived before the food was ready, so Item#onUse refused it - but
   the use state had already been cleared, so the timer never finished the meal.
  * START_SPRINTING. Bedrock lets a player eat while sprinting, and sprint is
   toggled constantly in a fight (knock-back, a released stick).
  * the fall-through at the end of PLAYER_ACTION_PACKET, which is reached by every
   action a client with server authoritative movement or block breaking sends,
   including its own sprint and swimming toggles and a missed swing.

  Keep the use state in those three places while a consumable is in progress. Items that
  carry no use duration - bow, crossbow, shield - are released by the client and keep the
  old behaviour everywhere.

- **Restore authoritative inventory snapshots without splitting slots**

  setItemForce is only for server rollback and restore. Splitting a legacy overstack
  changed unrelated slots, discarded a full-inventory remainder, and lost redistributed
  items when later snapshot entries overwrote them. Restore the exact saved count and
  metadata to its original slot; ordinary setItem and addItem still cap new writes. Cover
  full inventory, real SAI rollback, repeated restore, NBT and stack network identities.

- **inventory: read the recipe book craft repetitions once on v2168**

  Since v2168 the client writes the repetitions byte of CRAFT_RECIPE_AUTO once. The
  decoder still read the duplicate that 1.21.20 introduced, so every ingredient shifted by
  one byte: the descriptor duplicate swallowed the length of the identifier, the string
  read garbage, and the whole batch failed with
  "Unable to decode ItemStackRequestPacket". The player saw a recipe book craft do nothing
  at all, in the inventory grid and on the crafting table alike.

  Three captures from live 1.26.45 clients decode to the last byte only with the single
  field, and both the Cloudburst v2168 reader and the PocketMine protocol fork read it the
  same way. The existing cross-encode test passed only because the Cloudburst v2168 writer
  emits an extra ingredient-count byte that no client sends; it now writes the client
  layout by hand. Older protocols are untouched.

- **Send the ender chest lid event to the chunk of the chest**

  PlayerEnderChestInventory addressed the lid BlockEventPacket and the open/close sound to
  the chunk and position of its holder, the player. Player.teleport() and
  teleportImmediate() call removeAllWindows() only after the player has been moved, so a
  viewer who was respawned or teleported with the ender chest open closed the window from
  the destination: the close event went to the destination chunk and the lid stayed open
  for every player around the chest. On a busy PvP arena the stuck lid rendered as a black
  slab right in front of the camera.

  The lid belongs to the block, so both events now go to the level and chunk of 
  BlockEnderChest (the viewer's level is only a fallback when the block has none). 
  PowerNukkitX (HumanEnderChestInventory) already addresses the chest position; Lumi and
  Nukkit-MOT master still use the holder.

  Regression: PlayerEnderChestLidTest (close after a teleport to another level, open while
  standing in the neighbour chunk); both fail on the previous code.

- **Share ender chest viewers and close its window on -1 and on death**

  Three defects around the ender chest window, all visible as a lid that hangs open or
  shuts on the wrong player:

  1. A 1.21 client closes the ender chest with ContainerClose windowId -1 while
    its own inventory is not open. The -1 branch only reset the crafting grid,
    so the window stayed open on the server: the lid hung open for everyone and
    addWindow() returned the stale id on the next click, so the chest did not
    open again for this player until a teleport. The branch now closes an open
    PlayerEnderChestInventory before replying.

  2. Player.kill() closed no windows; they were closed only by the respawn
    teleport. A player who never pressed respawn left the lid open, and windows
    that hand input slots back on close (trade, anvil, grindstone) returned them
    into the inventory of the respawned player instead of the dropping one.
    Windows are now closed before the death drops are collected (vanilla closes
    containers on death). Transactions of a dead player are already refused.

  3. BlockEnderChest kept its viewers in a field, but a Block object is created
    anew for every lookup, so each opener saw a set of one: the first player to
    close shut the lid for the second, and a viewer was never removed when the
    set held two. Viewers now live on BlockEntityEnderChest (per position), the
    close removes the viewer unconditionally and sends the close event only when
    the set becomes empty; a failed addWindow() no longer leaves a viewer behind,
    and switching chests removes the player from the previous one.

  Compared: Allay counts ender chest viewers on the block entity (taken here); 
  PowerNukkitX (HumanEnderChestInventory) sends open/close per player with no count and
  does not close windows on -1 or death; Lumi stable and Nukkit-MOT master carry all three
  defects.

  Regressions: PlayerEnderChestViewersTest, PlayerEnderChestCloseTest (all fail on the
  previous code); PlayerEnderChestLidTest mock viewer is now spawned.

- **Close world container windows on -1 and when the container breaks**

  Two leftovers of 0194, both visible as a lid that hangs open or a window in front of a
  block that is gone:

  1. A 1.21 client closes a chest, barrel, shulker box, hopper or furnace with
    ContainerClose windowId -1 while its own inventory is not open. 0194 closed
    only the ender chest in that branch; every other container stayed open on
    the server, the lid hung open for everyone and the next click got the stale
    window id, so the block did not open again for this player. The branch now
    also closes windows whose holder is a block entity that still stands in its
    level. A window drawn on a fake block (FakeBlockMenu, plugin menus) has no
    block entity and is left alone: plugins reopen such windows quickly, and a
    late -1 must not shut the fresh one.

  2. Breaking a container did not close every window:
    - the ender chest window is the player's own inventory, and nothing closed
      it when the block broke; BlockEntityEnderChest.close() now closes the
      windows of its viewers;
    - BlockEntityChest.onBreak() unpaired before close(), so the double chest
      inventory was forgotten and its viewers kept a window of a chest that no
      longer existed; the double windows now close before unpairing;
    - dispenser and dropper had no close() at all, unlike every other container
      block entity; they now close their viewers' windows.

  Compared: Allay closes any open container on -1 and removes all viewers when a container
  block is replaced (taken, limited to real block entities here); PowerNukkitX and Lumi
  stable keep the -1 gap, PowerNukkitX unpairs a double chest before close() as well;
  Nukkit-MOT master carries both defects. Allay does not close ender chest windows on
  break either.

  Regressions: PlayerWorldContainerCloseTest, ContainerBreakClosesWindowsTest
  (fail on the previous code).


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Seven independent fixes in the inventory and item-transaction path: stack request decoding (unknown identifiers and recipe-book container slot types), partial stack merges, durability correction for non-durable items, consumable use interrupted by repeated clicks/sprinting, authoritative inventory snapshot restore, v2168 auto-craft repetitions decoding, and ender chest lid event addressing. Each commit is self-contained with a clear reproduction in the commit message. The changes are well-scoped, consistent with vanilla behavior, and covered by new regression tests.

Verification performed:
- Full test suite run on the PR head: **4920 tests, 0 failures** (including the 8 new/updated test classes: PlayerConsumableUseTest, InventoryForceRestoreConservationTest, OverstackSplitTest, PlayerEnderChestLidTest, ItemStackResponseDurabilityTest, PartialStackMergeTest, AutoCraftRecipeActionDecodeTest, MiscDecodeRegressionTest).
- Antivirus scan of the changes: clean.
- Manual review of the complete diff (Codex was unavailable; manual review performed).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `Player.java`: two Javadoc blocks are stacked above `isRepeatedConsumableUse` — the first block (describing `isConsumingItem`) is orphaned and attaches to no declaration. Consider moving it above `isConsumingItem()`.
- `ContainerSlotType.fromId(id, gameVersion)`: on the v1_20_50 branch only id 63 (`DYNAMIC_CONTAINER`) is rejected, so wire ids 64-66 now decode to the new recipe-book enums instead of `null` on that protocol. Older clients never send these ids, and unknown containers are rejected downstream, so the practical impact is nil — noted for completeness.
- `BaseInventory`: the `protected final routeOverflowForce` helper was removed. It was an internal contract (final, single caller), but it is technically an API removal for any external subclass that called it.

### Recommendations
- Fix the Javadoc ordering in `Player.java` in a follow-up.
- Optionally harden the v1_20_50 branch of `ContainerSlotType.fromId` to reject ids above `DYNAMIC_CONTAINER`.

### Notes
- The v2168 single-repetitions decode logic already exists on the base branch; commit `20b6892` only adds the missing regression tests with hand-written client-layout captures (the Cloudburst writer emits an extra byte no client sends). Verified against base.
- `durabilityCorrection` uses `ItemDurable` as the protocol boundary across all response-slot builders; no remaining `getDamage()` leaks into item-stack responses were found.
- The partial-merge clamp in `TransferItemActionProcessor` correctly recomputes `fullTransfer` and preserves the error path only when the destination is already full.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #977, opened 2026-09-20
fix: entity and combat behaviour: absorption cap, crits, totems, enderman, horse slots

w1zardz:
Eight independent fixes in entity and combat handling, each one bringing the server back to what the Bedrock client expects.

the reproduction.

- **entity: stop treating powder snow as a nether portal**

  Entity.checkBlockCollision declared a powderSnow flag, never used it and set the portal
  flag for powder snow instead. Any non-player entity standing in powder snow for the
  portal delay was therefore teleported to the nether: farm animals and pets vanished from
  a fenced pen with nothing in the log, while Player.checkBlockCollision, which owns its
  own copy of this loop, handled the same block correctly.

  Powder snow now sets its own flag and no longer skips the collision callbacks, and the
  freezing counter and the frostbite damage mirror the player path, so mobs freeze in
  powder snow the way vanilla expects instead of travelling between dimensions.

- **entity: open the horse screen so a rider can reach the saddle and armor slots**

  HorseInventory existed with saddle and armor slots and the item stack request mapping
  already resolved HORSE_EQUIP, but nothing ever opened the window: only a chest boat
  answered ACTION_OPEN_INVENTORY while riding. On Bedrock the horse screen is the way
  armor is put on a horse, so a saddled horse could never be armored - tapping it only
  mounts the player once it is saddled.

  Three changes:

  - HorseInventory extends ContainerInventory, so opening it sends
   ContainerOpenPacket with the horse entity id and closing it answers with
   ContainerClosePacket, like every other entity backed container.
  - A rider pressing the inventory button gets the horse screen instead of their
   own bag.
  - Crouch and use opens the same screen without mounting. Items the horse
   itself accepts (feed, saddle, armor, chest) keep their own meaning.

  The window is also removed from its viewers when the horse is closed, so a despawned or
  killed animal cannot leave an open screen behind.

- **entity: bound client absorption capacity on 1.26**

  Bedrock 1.26 renders the advertised absorption maximum as HUD capacity. Nukkit sent
  Float.MAX_VALUE, so one enchanted golden apple created thousands of empty hearts and
  made Android clients stall even though the real absorption value remained 16.

  Keep the server-side attribute range unchanged for plugins and older clients. Only the
  per-player packet for 1.26+ gets a finite capacity, never smaller than vanilla
  enchanted-apple absorption or the actual custom value.

- **entity: keep pathfinding body yaw while tracking a target**

  Walking mobs used lookAt(followTarget) before turning towards the current route node.
  That reset body yaw to the player every tick, so the 30-degree smoothing limit could
  never reach a lateral detour and obstacle checks inspected the wrong facing.

  Track the live target with head yaw only after the body turns towards the route. Keep
  the old body-and-head behavior for wandering without a live target, and cover both
  orientation paths.

- **entity: deny critical hits to sprinting and flying attackers**

  Vanilla Bedrock grants the 1.5x critical only to an attacker who is committing to a
  fall: PocketMine-MP checks !isSprinting() && !isFlying() && fallDistance > 0
  && !blindness && !underwater in Player::attackEntity. Nukkit-MOT checked the fall, the
  mount, blindness and water, but not sprinting and not flight.

  Both omissions turn a timed hit into a permanent multiplier. Bedrock players sprint by
  default, so every sprint-jump attack was critical; a player with creative or
  plugin-granted flight hovered and hit critically for free, never paying the fall the
  mechanic is priced around.

- **entity: preserve the hurt window after a successful totem rescue**

  Totem rescue cancels lethal damage and returns false, bypassing the ordinary attack
  timer. Arm attackTime after a successful rescue and retain longer timers. Equal or
  weaker follow-up hits wait for the normal interval; stronger hits and lethal command
  damage keep their existing behavior. Preserve noDamageTicks, hand order and the
  cancelled rescue result. Test the real attack chain and EntityLiving ticks, timer
  boundaries, hand replacement, cancelled damage, stronger attacks and independent
  players.

- **entity: enderman dodges projectiles before taking damage**

  EntityEnderman#attack called super.attack() first and only then checked for 
  DamageCause.PROJECTILE, cancelled the event and teleported. By that point Entity#attack
  had already fired the event and taken the health away, so the late cancel only hid the
  hit: arrows, crossbow bolts and other projectiles wore the enderman down and killed it,
  while vanilla endermen take no damage from projectiles and teleport away instead.

  The projectile branch now runs before super.attack(). Listeners that want to keep the
  projectile from reaching the enderman at all can still cancel ProjectileHitEvent, which
  every projectile fires before calling attack(). Melee and other causes keep their
  previous behaviour.

  getSafeTpLocation() also rolled a single random column within 16 blocks and gave up when
  it hit a wall or the void, which left the enderman standing in place most of the time.
  It now tries up to ten columns, like vanilla keeps rolling for a valid spot.

- **entity: stop reporting the empty vanilla armour slot as a skipped entry**

  EntityZombie#initEntity reads four armour slots and warns about every saved entry it
  cannot place. Vanilla Bedrock writes armour as FIVE positional entries with no Slot tag
  - head, chest, legs, feet and the body slot added in 1.20.5 - so the fifth entry always
  lands past the end of the Nukkit array and is reported as skipped.

  Dumped from an imported arena world, every zombie looks the same:

    minecraft:zombie size=5 [noSlot keys=[Count, Damage, Name, WasPickedUp]] x5

  The fifth entry is empty, and so is every other slot of a mob that spawned without 
  armour, so nothing is lost - but the warning fires once per mob on every load of a world
  that vanilla or an importer ever wrote. Three of our servers logged 2106 of these lines
  in four days, in bursts of a dozen each time an arena was restored from its template.

  Only an entry that actually holds an item is counted and reported now; an empty one is
  dropped in silence, because dropping it loses nothing. Both spellings of "empty" are
  recognised: vanilla writes Count 0 with an empty Name, Nukkit writes id 0.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Eight independent, well-scoped fixes to entity and combat behavior, each restoring vanilla Bedrock expectations: powder snow no longer teleports mobs to the nether, the horse screen can finally be opened for saddle/armor slots, absorption HUD capacity is bounded for 1.26 clients, pathfinding body yaw is preserved while tracking a target, critical hits are denied to sprinting/flying attackers, the hurt window is preserved after a totem rescue, endermen dodge projectiles before taking damage, and empty vanilla armor entries no longer spam the log.

Each change was verified against the source: `Attribute.getAttribute()` returns a clone (no shared-state mutation from the 1.26 absorption cap), `AdventureSettings.get()` returns a primitive boolean (no NPE risk in the new flight check), `getFrostbiteInjury()`/freezing-tick APIs and the 3-arg `setRotation` all exist and are used correctly, and `ContainerInventory` sends `ContainerOpenPacket` as the horse screen test expects. Returning `false` from `EntityHorseBase.onInteract` is safe — the caller breaks out of interaction handling without consuming the item.

Build verification: `mvn test` with the five new/updated test classes (`EntityAbsorptionAttributeTest`, `EntityWalkingOrientationTest`, `PlayerTotemHandPriorityTest`, `EntityEndermanTest`, `HorseInventoryTest`) passes (exit 0). ClamAV scan of the diff: clean. (Codex was unavailable due to an API key outage; this review was performed manually against the full diff and source.)

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- **Powder snow freezing applies to every entity type.** Vanilla grants freezing immunity to snow golems, strays, polar bears and the wither; the new `Entity.checkBlockCollision` path freezes and frostbites all of them, and ignores leather armor on armored mobs. Only affects mobs standing in powder snow long-term, so impact is minor, but a future `isFreezeImmune()` check would close the gap with vanilla.
- **Duplicated Javadoc on `canCriticalHit`.** The original short Javadoc was left in place directly above the new full Javadoc, producing two stacked doc comments on one method. Harmless, but the stale one should be dropped.

### Recommendations
- Consider a freeze-immunity check (entity type whitelist or NBT/armor awareness) in the new powder-snow path in a follow-up.
- Remove the leftover Javadoc fragment above `canCriticalHit` when touching the file next.

### Notes
- The eight commits are self-contained and can be dropped independently, as the PR description states.
- Totem rescue change is covered by an unusually thorough test matrix (timer boundaries, hand replacement, cancelled damage, stronger attacks, per-player isolation).

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #978, opened 2026-09-20
fix(block): match enchanting table, cauldron, shulker box and shelf to vanilla collision shapes

w1zardz:
Six independent fixes that bring block collision shapes and blast resistance in line with what the Bedrock client uses. The server used full cubes or wrong heights where the client uses a composite or lower shape, so players could stand where the client says they cannot, or were pushed back where the client lets them walk.

the reproduction.

- **Match enchanting table and cauldron collisions**

  The client treats the enchanting table as twelve pixels high and the cauldron as a
  hollow frame with a five-pixel bottom. The server used a full cube for both, so movement
  onto either block was corrected back and forth against geometry the client could not
  see.

  Expose composite collision boxes to the movement helper, match both shapes to Bedrock,
  and cover the solid and hollow parts with regressions.

- **block: restore the vanilla blast resistance of shulker boxes**

  Nukkit stores five times the vanilla blast resistance and Explosion divides by five
  again while tracing a ray. BlockShulkerBox returned 30, which is a blast resistance of 6
  instead of the vanilla Bedrock 2, so primed TNT flattened the stone around a box and
  left the box itself untouched.

  Return 10 so dyed and undyed boxes match PocketMine-MP, where the shulker box break info
  is pickaxe(2) and the default blast resistance is five times the hardness. Hardness,
  tool type and drops are unchanged.

  BlockShulkerBoxBlastResistanceTest covers both block classes, keeps the box weaker than
  a chest and walks the explosion formula for the weakest TNT ray.

- **block: match dirt path, farmland and soul sand collisions**

  The client walks one pixel below the top of a dirt path and farmland and two pixels
  below the top of soul sand. The server used a full cube for all three, so a creative
  player stepping onto them collided with geometry the client could not see and was
  corrected back every frame.

  Trim the three boxes to the Bedrock heights used by the live network and cover them with
  regressions.

- **block: keep a lone pane a post and give stairs both collision halves**

  A pane with no neighbours fell into the branch that widens the box to the whole cell, so
  the server had a full cube where the client draws a two-pixel post: a creative player
  walking past one was pushed back every frame. Stairs exposed only the lower slab through
  getBoundingBox, so the movement resolver never saw the step half.

  Keep the centre post when nothing connects and hand both stair halves to the collision
  list.

- **block: match stonecutter, hopper and snow layer collisions**

  The stonecutter is nine pixels high on the client and the hopper has a bowl a player
  drops into, but the server used a full cube for both. Snow steps two pixels per layer;
  the server had zero up to two layers, half a block up to seven and a full block only at
  eight.

  Match all three to Bedrock and cover them with regressions.

- **Match the shelf collision to the vanilla shape**

  A shelf occupies five pixels next to its wall: a lower board across the whole depth, a
  three pixel back panel and a top board above the niche. The server used a full cube, so
  one player stood on a shelf as if it were a block while another fell through it, and
  every step near a shelf was corrected against geometry the client cannot see.

  Give the shelf its composite collision boxes, keep the bounding box on the band it
  really occupies, and cover both the free space in front of it and its boards with
  regressions.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Three independent fixes aligning server-side collision shapes and blast resistance with vanilla Bedrock: composite collision boxes for enchanting table (12px height), cauldron (hollow frame with 5px bottom) and shelf (5px wall band), plus shulker box blast resistance corrected from 30 to 10 (vanilla 2 × Nukkit's 5× scale). A new `Block#addCollisionBoxesToList` hook lets blocks expose composite shapes to `CollisionHelper` while the default implementation preserves the existing bounding-box behavior for every other block. All 9 new regression tests pass locally (`BlockCollisionShapeTest`, `BlockShelfCollisionTest`, `BlockShulkerBoxBlastResistanceTest` — BUILD SUCCESS). No blocking issues found; the changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `BlockCauldron#recalculateCollisionBoxes` and `BlockShelf#recalculateCollisionBoxes` allocate a fresh array plus 3–5 `SimpleAxisAlignedBB` instances on every collision query; this sits in the entity movement hot path, so caching per-block-state could reduce GC pressure.
- The default `Block#addCollisionBoxesToList` resolves `getBoundingBox()` twice (once inside `collidesWithBB`, once for the add); negligible, but a single lookup would be cleaner.

### Recommendations
- Consider caching the collision box array (invalidated on block state change) if profiling shows movement collision as a hotspot.
- The one-argument `collidesWithBB(bb)` on cauldron/shelf now uses the hollow/composite shape instead of the bounding box; worth a quick sanity check that no caller relied on the old full-cube semantics for these blocks (entity collision callers are the intended consumers).

### Notes
- Shulker box resistance change is consistent with PocketMine-MP and the `Explosion` ray formula (resistance / 5); hardness, tool type and drops are untouched.
- The `defaultCollisionBoxesPreserveVirtualShapePredicate` test documents that non-composite blocks keep the legacy "bounding box added even when the collision BB does not intersect" semantics — good call-out of preserved behavior.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #980, opened 2026-09-20
fix: sleeping percentage and NPE from auth input arriving before the inventory exists

w1zardz:
Two small independent fixes in the player session path.

the reproduction.

- **level: compare the sleeping share as a percentage**

  `checkSleep` computed `sleepingPlayerCount / playerCount * 100` in integer arithmetic:
  the division is zero for every case except "everybody sleeps", where it is exactly 100.
  So `playersSleepingPercentage` did nothing at all — any value from 1 to 100 behaved like
  100, and any value above 100 made the night unskippable even with the whole level
  asleep. Multiplying first gives the real share.

- **player: tolerate a missing inventory on an early auth input**

  The shared hold-to-use gate added in 4fd98c1d reads the held stack straight from 
  this.inventory. That field is created by EntityHumanType.initEntity, which for a player
  runs at login, so an auth input that arrives before the profile is loaded throws a
  NullPointerException inside handleDataPacket and kills the session.

  Read the held stack only when the inventory exists; isHoldToUseItem already treats a
  null stack as not hold-to-use, so a pre-login input simply carries no start signal.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Two small, independent fixes in the player session path, each in its own commit:

1. **Level.checkSleep**: `sleepingPlayerCount / playerCount * 100` was evaluated in integer arithmetic, so the division truncated to 0 in every case except "everybody asleep" (exactly 100). `playersSleepingPercentage` therefore behaved like 100 for any value 1–100, and values above 100 made the night unskippable. Multiplying first computes the real percentage.
2. **Player.handleDataPacket (auth input)**: the shared hold-to-use gate read the held stack from `this.inventory`, which is only created by `EntityHumanType.initEntity` at login. An auth input arriving before the profile is loaded threw an NPE and dropped the session. The held stack is now read only when the inventory exists.

### Issues Found
#### Blocking Issues
- None

#### Non-blocking Issues
- Neither fix ships a regression test. The `UsingItemReceive` gate helpers are already unit-tested, but the null-inventory path in `Player` and the percentage arithmetic in `Level.checkSleep` are not covered. Given how small and self-contained both changes are, this is acceptable, but a follow-up test would prevent regressions.

### Recommendations
- Verified null-safety chain: `UsingItemReceive.isHoldToUseItem(null)` returns `false`, so with a null inventory the start gate `shouldStartUsingFromAuthInput` can never pass and `tryStartUsingHeldItem` is never called with a null stack. The guard is complete for this code path.
- Checked the rest of the `PLAYER_AUTH_INPUT_PACKET` block for other unguarded `this.inventory` dereferences — this was the only one.

### Notes
- The integer-overflow risk of `sleepingPlayerCount * 100` is negligible (would require >21 million players).
- Semantics for edge values are preserved: percentage > 100 still makes the night unskippable; percentage = 100 still requires everyone asleep.
- Note: this review was performed manually because the automated review backend is currently unavailable.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

PR #981, opened 2026-09-20
fix: encode break particles with the receiving client's block palette

w1zardz:
A single fix: break particles were encoded with the server's palette, so clients on another protocol showed the wrong block.

the reproduction.

- **Use the recipient item palette for break particles**

  An old viewer could throw during golden spear break particles, preventing the broken
  spear from leaving its owner hand. Resolve only the outbound particle against the viewer
  palette and use the existing info_update placeholder when unavailable. Preserve native
  payloads, server items, NBT and global mappings.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
A single, well-scoped fix: `ItemBreakParticle.mvEncode` previously resolved the item's runtime ID against the server's palette unconditionally, so a viewer on an older protocol without the item in its palette (e.g. golden spear) triggered an `IllegalArgumentException` from `getNetworkId(protocol)`. The exception propagated out of the particle encode path and prevented the broken spear from being removed from the player's hand. The change resolves the outbound particle against the recipient's palette and falls back to the `INFO_UPDATE` placeholder (damage reset to 0) only for that recipient when the item is unavailable. Server items, NBT, and global mappings are never mutated. A comprehensive compatibility test (92 cases across protocol versions) accompanies the fix and passes (`mvn test -Dtest=ItemBreakParticleCompatibilityTest`, BUILD SUCCESS).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None worth flagging. The try/catch is narrowly scoped to `IllegalArgumentException` from the palette lookup, which is the intended control flow here.

### Recommendations
- None. The fallback mirrors the existing inventory placeholder convention and keeps the pre-1.16.100 path untouched.

### Notes
- Verified the fix against `ItemSpear.damageSpear`: the particle is now safely encodable for every supported client version, so the broken item reliably leaves the hand.
- Antivirus scan of changed files: clean.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

tachibana-kanade-angel:

tachibana-kanade-angel:
The changes in this PR are already present on the current master (merged via another PR), so this PR no longer has any effect.

Closing it. Thank you for the contribution!

PR #982, opened 2026-09-20
fix: restore vanilla melee knock-back and cap its vertical part

w1zardz:
Two fixes to melee knock-back. The second one depends on the first: it caps the vertical part of the push that the first one restores.

the reproduction.

- **entity: restore vanilla melee knock-back and the Knockback enchantment push**

  Melee knock-back has been 0.3 since 8dca6ffd, and the Knockback enchantment only adds
  0.1 per level on top of it. Vanilla Bedrock — and PocketMine-MP, which mirrors it —
  pushes with 0.4 and applies the enchantment as a SEPARATE knock-back of 0.5 per level
  right after the hit lands (Living::DEFAULT_KNOCKBACK_FORCE and 
  KnockbackEnchantment::onPostAttack).

  The gap is large enough to change how melee combat feels: a bare hit pushes 25% less far
  and, because the vertical cap equals the force, lifts the victim noticeably less, while
  a Knockback I sword ends up weaker than a bare vanilla hit
  (0.4 total against 0.4 plus a 0.5 push). Players moving from a PocketMine server report
  that victims can no longer be chained at all and that fast clicking beats positioning,
  which is exactly what a short, flat push produces.

  Restore the vanilla numbers:

    * EntityLiving.DEFAULT_KNOCKBACK is 0.4 and is used by the short knockBack
     overload and by both EntityDamageByEntityEvent constructors that default it;
   * a melee hit from a player asks for that force instead of 0.3;
   * EnchantmentKnockback pushes the victim again with level * 0.5 from
     doPostAttack, the way the enchantment works in vanilla, instead of nudging
     the base force.

  The horizontal direction and the vertical cap keep their existing meaning, so the change
  is a value fix, not a new mechanic.

- **entity: cap the vertical part of a knock-back at the vanilla limit**

  Since the melee knock-back restore, the Knockback enchantment pushes the victim again
  with level * 0.5 (0111). knockBack caps the vertical component at the force it was
  given, so that second push lifts the victim by a full 0.5 with Knockback I and 1.0 with
  Knockback II — several times a jump (0.42) — and players report being launched into the
  air instead of being pushed away. A Knockback II sword became unplayable to fight
  against.

  Vanilla Bedrock, and PocketMine-MP mirroring it, keeps the vertical ceiling separate
  from the force: Living::DEFAULT_KNOCKBACK_VERTICAL_LIMIT is 0.4 and does not grow with
  it, with the comment that "without this limit, the entity may be knocked far up into the
  air with large knockback forces". The horizontal push still scales with the force, which
  is what the enchantment is meant to do.

  Add DEFAULT_KNOCKBACK_VERTICAL_LIMIT and a knockBack overload that takes the ceiling.
  The five-argument overload keeps its old meaning for every force at or below the limit
  (min(base, limit)), so an ordinary hit, a shield bash and an event-authored knock-back
  move exactly as before; only a force above 0.4 — today only the enchantment — stops
  throwing the victim upwards.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Two-part fix restoring vanilla Bedrock melee knock-back behavior: (1) base melee knock-back raised from 0.3 to 0.4 (`EntityLiving.DEFAULT_KNOCKBACK`) and the Knockback enchantment converted from a +0.1/level force nudge into a separate 0.5/level push via `doPostAttack`; (2) vertical knock-back capped by a new `DEFAULT_KNOCKBACK_VERTICAL_LIMIT` (0.4) decoupled from the force, so high-force pushes no longer launch victims into the air. Values match vanilla Bedrock / PocketMine-MP semantics. Compiles cleanly (`mvn -o compile`, exit 0).

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- Plugins constructing `EntityDamageByEntityEvent` via the default constructors now get 0.4f knock-back instead of 0.3f — intended vanilla alignment, but a visible behavior change for downstream plugins.
- Event-authored knock-back with a force above 0.4 now has its vertical component capped at 0.4 instead of the force value. This is the stated intent of the change and matches vanilla, but is worth noting in release notes.

### Recommendations
- None. The five-argument `knockBack` overload preserves its previous behavior for all forces at or below the limit via `min(base, verticalLimit)`, so existing callers (shield block 0.25, generic event knock-back) are unaffected.

### Notes
- Verified `EnchantmentKnockback.doPostAttack` is wired into the melee attack flow: `Player.attack` invokes `enchantment.doPostAttack(this, target)` for all held-item enchantments after a successful hit, and the legacy `InteractProcessor_v113` path does the same.
- Knock-back direction (`entity.pos - attacker.pos`) pushes the victim away from the attacker; the zero-damage call does not interfere with damage calculation.
- ClamAV scan of the diff: 0 infected files.
- Automated deep review was unavailable (review backend API key disabled); this review was performed manually against the full diff, call-site wiring, and a successful offline compile.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #983, opened 2026-09-20
fix: show skin NPCs on 1.26 clients and cover the packet handshake with tests

w1zardz:
One change with tests. Skin NPCs (statues, bots, doubles - any EntityHuman that is not a Player) disappeared or showed as the default model on 1.26 clients because of the order and content of the player list entry and the skin packet.

the reproduction.

- **entity: show skin NPCs on 1.26 clients and cover the handshake with tests**

  A skin NPC (an EntityHuman that is not a Player - a statue, a bot, a double) needs three 
  packets in a fixed order: a PlayerList ADD entry, the AddPlayer packet, and only then
  the real skin.

  Two halves were wrong and each hid the other:

   - a PlayerList entry carrying a full persona skin with an empty profileHash is rejected
  by
    1.26 clients, and without the entry there is nothing for AddPlayer to draw, so every
  skin
    NPC on the server disappeared;
  - a PlayerSkinPacket sent before the entity exists is dropped silently, so painting the
  skin
    early left every NPC as the default model.

  The entry is now added with a blank placeholder skin that every version accepts, the
  real skin follows right after the spawn packet while the entry is still live, and
  removal of the entry is delayed a few ticks so the client has time to apply it.

  The change is covered by tests for the packet order, for the placeholder entry and for
  the delayed removal.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Fixes skin NPCs (any EntityHuman that is not a Player) disappearing or rendering as the default model on 1.26 clients by unifying the spawn handshake for all viewers: the PlayerList ADD entry now carries an opaque placeholder skin, the real skin is re-sent right after AddPlayerPacket, and entry removal is delayed a few ticks. The handshake is covered by parameterized tests across four client versions (V1_21_124, V1_21_93_NETEASE, V1_21_124_NETEASE, V1_26_45). No blocking issues found; the changes look reasonable overall.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- `PlayerEntitySkinSender.sendSkinnedEntryIfAbsent` is added but never called from main sources — dead public API. Remove it or wire it up.
- `PlayerEntitySkinSender` is still annotated `@OnlyNetEase` and documented as the NetEase V860 handshake, but after this change it drives the skin lifecycle for every client, including international 1.26. The annotation and class javadoc no longer match reality.
- In `EntityHuman.spawnTo`, `retainNpcListEntry` is now exactly `!(this instanceof Player)`, which makes the final `else` branch and the immediate-removal guard (`!retainNpcListEntry`) unreachable dead code.

### Recommendations
- Drop `sendSkinnedEntryIfAbsent` (or actually use it) before merge.
- Update the class-level docs of `PlayerEntitySkinSender` to reflect the now-universal handshake.
- Simplify `spawnTo` to remove the unreachable branches.

### Notes
- The behavior change is intentional and justified: international 1.26 clients reject a player-list entry carrying a full persona skin with an empty profileHash, and a PlayerSkinPacket sent before the entity spawns is silently dropped. The placeholder uses opaque Steve pixels, so a dropped skin packet degrades to a visible default rather than an invisible NPC.
- Rollback on rejected ADD/skin packets, generation-based invalidation of stale delayed tasks, and the despawn cleanup path are consistent and covered by tests.
- Verified locally on the PR head: `EntityHumanSkinLifecycleTest` 15/15 passed, build success.

@lt-name ready to merge

---
🤖 AI Review

tachibana-kanade-angel (approved):

PR #984, opened 2026-09-20
fix: take the nether portal target once and drop a cancelled transfer

w1zardz:
One fix for a crash and a stale transfer in the vanilla portal path.

the reproduction.

- **player: take the portal target once and drop a cancelled nether transfer**

  A nether transfer is decided in one tick and carried out two hops later: an async task
  looks the far side up, and a main thread task moves the player. Both of them read the
  portalPos field, and the main thread clears that field the moment the player stops
  standing in the portal - a plugin teleport, a
  /spawn, a death. So the lambda could reach spawnPortal with null and kill the whole
  tick:

    FATAL - Exception while invoking run
   java.lang.NullPointerException: Cannot read field "level" because "pos" is null
     at cn.nukkit.block.BlockNetherPortal.spawnPortal(BlockNetherPortal.java:386)
     at cn.nukkit.Player$1.lambda$onRun$0(Player.java:2387)

  Seen twice on the live grief server on 19.09.2026, both times seconds after the player
  who was standing in the portal ran /spawn and a plugin teleported them away. The field
  was left occupied by the throw, so the next portal step was never armed again either.

  The target is now taken once, before the async task is scheduled, and both hops use that
  snapshot. The world the player is taken from is remembered with it: leaving that world
  before the lookup ends calls the trip off, because dragging somebody into the nether
  seconds after they typed /spawn is worse than not transferring them at all. A cancelled
  transfer clears the field first, so the next portal step arms as usual.

  PowerNukkitX does the same work synchronously inside the tick over a local variable
  (Entity#onUpdate, PortalHelper.spawnPortal), so it cannot lose the target this way; Lumi
  carries the same field read and the same hole.


tachibana-kanade-angel:
## 🤖 AI Review

### Summary
Fixes a race in the vanilla nether portal transfer. Previously the async lookup task and the main-thread teleport lambda both read the `portalPos` field, which the main thread clears as soon as the player stops standing in the portal (plugin teleport, `/spawn`, death). The lambda could then call `spawnPortal(null)` and kill the tick with an NPE. The fix captures the target once (`portalTarget`), bails out early if it is already null, captures the origin level (`portalOrigin`), and on the main thread clears `portalPos` first so a cancelled transfer re-arms correctly, then abandons the transfer if the player went offline, died, or changed level. Single-file change, compiles cleanly (`mvn -o compile`, exit 0). ClamAV scan: 0 infected files.

Note: Codex was unavailable (API key disabled); manual review performed.

### Issues Found
#### Blocking Issues
- None.

#### Non-blocking Issues
- None.

### Recommendations
- None. The added comments explain the timing and the clear-first ordering well.

### Notes
- When `portalTarget == null` the method returns with `inPortalTicks = 81` already set; for creative players the transfer won't re-fire until the 80-tick path cancels the event, matching pre-existing behavior — not a regression.
- Build verified in an isolated worktree on top of the PR head (5661f3e5d).

@lt-name ready to merge

---
🤖 AI Review

lt-name:
Duplicate submission of https://github.com/MemoriesOfTime/Nukkit-MOT/pull/972

